PR #847, opened 2021-02-08
fix(readme): v6 discontinued

philsturgeon:
v6 currently lives in master, and is the default branch, so its also showing up in NPM. This makes it very clear that v6 is used at their own risk.

It still points to hidden documentation, but one thing at a time!


ToastM8 (commented):
Test

PR #1036, opened 2021-04-12
fix: wrap parameter descriptions

marcelltoth:
As confusing as it sounds, `sl-truncate` actually forces `white-space: nowrap;`. It makes some sense as `text-overflow: ellipsis` doesn't make much sense without a fixed-width, but still, we keep walking on this mine. I certainly do.

Of course this doesn't exactly make for a great user experience: 
image

Fortunately this has been fixed in JSV as well, so I just had to implement it for params. After:
image

I also added a few example APIs so we can test against them. (I picked the same ones Marc chose for the demo project, so the two are in sync).


mallachari (approved):
Good fix!

Issue #987, opened 2021-03-30
Tag caret icon missing in Web Components


philsturgeon:
## Describe the bug

Tags show a caret icon (little arrow) to show you can expand something when using storybook, but it seems to be missing in web components.

**Web Components (examples/web-component/github-api.html)**

<img width="357" alt="Screen Shot 2021-03-30 at 7 29 56 PM" src="https://user-images.githubusercontent.com/67381/113038593-9b311200-918e-11eb-892a-ccfb6f5612ad.png">

**Storybook (v7 branch)**

<img width="277" alt="Screen Shot 2021-03-30 at 7 30 35 PM" src="https://user-images.githubusercontent.com/67381/113038598-9bc9a880-918e-11eb-9776-edc7a03be544.png">

PR #1040, opened 2021-04-12
Fix(toc): missing caret icon in web components

mmiask:
Resolves: #987

### DIFF
--- a/packages/elements-web-components/src/index.ts
+++ b/packages/elements-web-components/src/index.ts
@@ -1,4 +1,10 @@
+import { dom, library } from '@fortawesome/fontawesome-svg-core';
+import { faChevronDown, faChevronRight, faCube } from '@fortawesome/free-solid-svg-icons';
+
 import { ApiElement, StoplightProjectElement } from './components';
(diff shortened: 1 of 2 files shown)


packages/elements-web-components/src/index.ts
```diff
@@ -1,4 +1,10 @@
+import { dom, library } from '@fortawesome/fontawesome-svg-core';
+import { faChevronDown, faChevronRight, faCube } from '@fortawesome/free-solid-svg-icons';
```
marcelltoth, line 2: Do we list these as `devDependencies` of `elements-web-components`?

marcelltoth (approved):
Beyond that 1 question, looks great now! Good work @mmiask !

packages/elements-web-components/src/index.ts
mmiask: No, these are direct dependencies in `elements`

mmiask: @marcelltoth Should we?

PR #1090, opened 2021-04-19
feat: set up CI workflow for lockfile maintenance

mmiask:
Resolves: #1090


.circleci/config.yml
```diff
@@ -77,7 +77,19 @@ jobs:
...
+    docker:
+      - image: circleci/node:12
+    steps:
+      - checkout
+      - run: git checkout main
+      - run: export BRANCH_NAME=feat/lockfile-maintenance-ci-job-$(date +"%m-%d-%Y") && git checkout -b $BRANCH_NAME
+      - run: yarn upgrade
+      - run: 'git add yarn.lock && git commit -m "chore: weekly lockfile maintenance"'
+      - run: export BRANCH_NAME=feat/lockfile-maintenance-ci-job-$(date +"%m-%d-%Y") && git push --set-upstream origin $BRANCH_NAME
+      - run: export BRANCH_NAME=feat/lockfile-maintenance-ci-job-$(date +"%m-%d-%Y") && export BODY='{"head":''"'${BRANCH_NAME}'"'',"base":"main","title":"Weekly lockfile maintenance"}' && curl -X POST -H "Accept:application/vnd.github.v3+json" -u $GIT_AUTHOR_NAME:$GH_TOKEN https://api.github.com/repos/stoplightio/elements/pulls -d "$BODY"
```
marcelltoth, line 90: Tips to reduce duplication (don't do all, pick what you like):
1. `export` isn't needed in the first line as you are not reusing it, it is only for that single line. `ENV_VAR=some-value some-command` does the same.
2. You can combine multiple commands into a single `run:` using the [`command: |` syntax](https://circleci.com/docs/2.0/configuration-reference/#default-shell-options). In that case it may actually make sense to do the export in the first line, then just use the variable in the rest.
3. `git checkout main` shouldn't really be required, is is the default branch. And we always want to target the default branch anyway.
4. `git push --set-upstream origin $BRANCH_NAME` can be `git push -u origin HEAD`, that's what I do locally anyway.
5. It might make more sense to retrive the current branch name rather than repeating the logic everywhere. Like `git branch --show-current` or something.

➕ Just realized that GitHub has its own CLI, that should abstract away the API call, resulting in simpler and more robust code. You could try [installing that](https://github.com/cli/cli/blob/trunk/docs/install_linux.md) then your actual command becomes much simpler: https://cli.github.com/manual/gh_pr_create

mmiask:
@lottamus I see that GH credentials in CircleCI belong to you, are you ok with us using it for this purpose or should we come up with something else? Do we have separate account for this kind of things? `Stoplight-bot` of some sort

.circleci/config.yml
mmiask: I cleaned up the `BRANCH_NAME` variable and used `command: |` syntax.

I tried to install `gh cli` on CircleCI env but it kept crashing.

marcelltoth:
> @lottamus I see that GH credentials in CircleCI belong to you, are you ok with us using it for this purpose or should we come up with something else? Do we have separate account for this kind of things? Stoplight-bot of some sort

We should change that anyway, personal creds for automation can cause all sorts of annoyances, and we have the bot account for it anyway.

I'd say don't worry about this in this PR - unless @lottamus objects - we'll handle that separately.

mmiask:
**Additional info:**

Initially we wanted to use `Renovate` but it turned out to be too complex to set up for only lockfile maintenance. It worked fine for doing typical maintenance (including `package.json` bumps) but whenever we turned off `npm` as one of the `managers` it did nothing. So we either had `package.json  + yarn.lock` bumps or nothing

And since using external package for basically running `yarn upgrade` seems like an overkill, we decided to go with a simple CircleCI workflow

mpodlasin (approved):
Looks good to me.

Hope it works. :joy:

Issue #1076, opened 2021-04-16
Some APIs load a white blank screen instead of the Overview


philsturgeon:
## To Reproduce

1. Given this OpenAPI demo: https://elements-demo.stoplight.io/?spec=https://api.apis.guru/v2/specs/bulksms.com/1.0.0/swagger.yaml
2. See nothing in the main col and a sidebar

## Expected behavior

I would expect to see the Overview, because it shows in the ToC. If there's no ToC, I would expect to see the first item in the ToC (an endpoint probably).

## Additional context

**Screenshots**

<img width="1158" alt="image" src="https://user-images.githubusercontent.com/67381/114994163-6910f700-9e94-11eb-93c5-774c8f863cbc.png">

PR #1104, opened 2021-04-21
fix: unknown pathname

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1076

When switching documents without remounting API component previously used pathname remains and is used to load document content. Redirecting to root path in such case fixes the issue and is something we should be doing anyway.


mpodlasin (approved):
Yup, this looks good.

What about StoplightProject? Shouldn't it have similar logic? :thinking:

mallachari:
> What about StoplightProject? Shouldn't it have similar logic?

We don't have such use case currently but I think it should as we can be changing its props without routing to any new uri. I will update that as well. Thanks!

PR #1106, opened 2021-04-21
chore(storybook): bump version

marcelltoth:
Bumps storybook version to 6.2

This actually fixes storybook for me, IDK why but the 6.1 one used to always get into a refresh-loop.

Does not change the builder to Webpack 5 yet, that turns out to be a bit more difficult to do, maybe later in a separate step.


package.json
```diff
@@ -15,6 +15,10 @@
   ],
   "dependencies": {},
   "devDependencies": {
+    "@babel/plugin-proposal-class-properties": "7.12.1",
+    "@babel/plugin-proposal-object-rest-spread": "7.12.1",
+    "@babel/preset-env": "7.12.11",
+    "@babel/preset-typescript": "7.12.7",
```
marcelltoth, line 21: This is unfortunately required to work around a bug, see https://github.com/storybookjs/storybook/issues/14197

mpodlasin:
web components don't work (?)

mmiask:
Storybook does not open for me :(

Screenshot 2021-04-22 at 11 42 25

mmiask:
@marcelltoth maybe let's specify the `engine` version range, so we avoid running into errors like this in the future?

```json
"engines": {
    "node": ">=10 <=12.22.1"
  },
```

mpodlasin (approved):
Works! <3

Issue #1056, opened 2021-04-14
ToC: Use operationId before using path when summary is missing


philsturgeon:
## User story.

As a user new to OpenAPI and Stoplight I might not have entered a summary for my operation, and I'd like my OpenAPI to still look as good as possible. In general I would like the ToC title to match the title displayed.

## Is your feature request related to a problem?

This is similar to #1035, in that the ToC and the Title of the page someone is looking at should agree with each other.

## Describe the solution you'd like

The ToC should use operationId before falling back to showing the path.

## Additional context

<img width="728" alt="Screen Shot 2021-04-14 at 11 27 07 AM" src="https://user-images.githubusercontent.com/67381/114696586-b9614b00-9d14-11eb-8a34-765db73c33f0.png">

<img width="795" alt="Screen Shot 2021-04-14 at 11 27 16 AM" src="https://user-images.githubusercontent.com/67381/114696588-b9f9e180-9d14-11eb-81ae-61905b453bbb.png">

You can use Stripe as an example: https://elements-demo.stoplight.io/?spec=https://api.apis.guru/v2/specs/stripe.com/2020-08-27/openapi.yaml#/operations/Post3dSecure

DavidBiesack:
The "user" of the API doc is a developer. While some developers may call APIs via naked HTTP, many rely on code generation ( `openapi-generator` or other) to create an SDK, so the `operationId` becomes the name of a function. It is therefore connective glue in the API doc that connects the API operation to the SDK. It is also the name of the operation which developers use to communicate with other developers. It's a lot easier (more concise, more precise) to slack/email a reference to an operation via the `operationId` compared to using the method + path or the verbose summary. This, I think it is important enough to always show the `operationId` even if the operation has a summary.

philsturgeon:
That sounds like another vote for "Custom SDK Samples" to me, as we'd not want end users needing to worry about operationId just so they can find the right SDK method to use. They'd just see the right method to use in the code sample panel.

DavidBiesack:
I'm referring to a full API First OpenAPI pipeline where someone generates the SDK outside of the API doc snippets and uses the generated SDK in their client or server codebase. I'm not asking for custom SDK code samples. I don't think anyone copies/pastes code snippets from the API doc to build their SDK - that code is just for learning the API, or for toy applications, not using it in production.
[...]

PR #1110, opened 2021-04-21
feat(API): new TableOfContents using Mosaic

### LINKED ISSUE #829: test: make sure we don't mess with tag name formatting
In previous versions we have done various things to tag names, for example Capitalizing them, humanizing them, or other assumptions which might not be correct. Elements leaves tags as they are, but we could do with some tests that confirm this is the case.

<p></p>

Try adding tests for various camelCase, Under_Scored, Space Speratated, etc tags and assert that they all show literally as they were input.

lottamus:
Happy to hop on a call to walk through the changes too!

This PR is branched off of https://github.com/stoplightio/elements/pull/1109

Demo: https://deploy-preview-1110--stoplight-elements-demo.netlify.app/#/

## What's in this PR?
- Add a new TableOfContents component using Mosiac
- Add a simple loading spinner while we're fetching the API document
- Fixes #829 

## What's left to do in this PR?
- [x] TableOfContents integration tests
- [x] Icons are not working in storybook

## What are you doing in a different PR?
- ~Fix icon colors (needs an update to mosaic): https://github.com/stoplightio/mosaic/pull/178~
- Add strike through styling for deprecated operations
- Update DevPortal component to use new Mosaic TableOfContents
- Remove old UI-Kit TableOfContents
- Add fallback endpoint name: https://github.com/stoplightio/elements/issues/1056

## Questions

- We're [no longer capitalizing tags](https://github.com/stoplightio/elements/issues/829), but the Zoom API's tags are all lower case. Maybe we should only capitalize a tag if we detect it's all lowercase? 🤷 
- What do we think about moving the "API" container/component into `/components/API/index.ts`? Eventually I think we want to get rid of the `containers` directory

## Screenshots

**Design**

image

**Before**

image

**After**

image


packages/elements/src/components/MosaicTableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,236 @@
...
+  isNodeGroup,
+} from './utils';
+
+const ActiveIdContext = React.createContext<string | undefined>(undefined);
+const LinkContext = React.createContext<CustomLinkComponent | undefined>(undefined);
+
+export const TableOfContents = React.memo<TableOfContentsProps>(({ tree, activeId, Link }) => {
+  React.useEffect(() => {
+    if (activeId && typeof window !== 'undefined') {
+      const elem = window.document.getElementById(getHtmlIdFromItemId(activeId));
```
mpodlasin, line 30: Perhaps it would be possible to use `ref` or something like that?

Not a must, but would be more `react`y. ;)

packages/elements/src/components/MosaicTableOfContents/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  title: string;
+  url: string;
+};
+
+export type TableOfContentsNode<T = 'http_service' | 'http_operation' | 'model' | 'article' | 'overview'> = {
+  id: string;
+  slug: string;
+  title: string;
+  type: T;
+  meta: string;
```
mpodlasin, line 40: I don't really like this `meta` title.

It seems to be done in order to make the names more generic, but in `elements` it will be always literally just a method, like get, etc.? :thinking:

packages/elements/src/styles/elements.scss
```diff
@@ -13,6 +13,10 @@ html {
   font-size: 13px;
 }
 
+a:hover {
```
mpodlasin, line 16: Is there a better way to solve this problem?

I believe at some point we hoped to "unscope" elements styles, and this would become an issue in that case.

mpodlasin (commented):
In general, I like it.

The only *must* for me would be `a` global CSS.

After you fix some of the issues pointed out, I'll accept. :)

packages/elements/src/styles/elements.scss
lottamus: This issue is coming from the default styles in UI-Kit, which we're working towards removing. This is a temporary fix until we can remove UI-Kit completely over the next few sprints.

packages/elements/src/components/MosaicTableOfContents/types.ts
lottamus: In the DevPortal component, it could also be a version for models and APIs such as "v1" or "v2.1". Open to suggestions on renaming the prop name

packages/elements/src/components/API/SidebarLayout.tsx
```diff
@@ -1,14 +1,11 @@
-import { generateApiToC } from '@stoplight/elements-utils';
```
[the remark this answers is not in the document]
lottamus: Good point, created an issue here: https://github.com/stoplightio/elements/issues/1142

packages/elements/src/components/MosaicTableOfContents/types.ts
mpodlasin: Hmm, then maybe it does make sense. Ok, let's keep it for now, we will see how it looks later with DevPortal.

mpodlasin (approved):
Ok, I say let's merge it.

lottamus:
@mpodlasin I added some tests in 9ba8c6f and fixed the storybook icons in b44b644. 🎉

mpodlasin:
Hmm, seems that CI got stuck? :thinking:

PR #1118, opened 2021-04-22
chore: remove SASS svg function from storybook

marcelltoth:
It does two things.

1. It aligns the storybook SASS config with the `build-styles` one. Those should of course show the same thing, otherwise we can fool ourselves by thinking something works by looking only at the storybook.
2. It fixes the problem where Storybook fails to launch on Node versions above 12.

It breaks one thing: `ui-kit`'s blueprint icons.

This is intentional. They wouldn't work in production anyway (see 1.) but we also don't need it: we eliminated all blueprint icons from Elements already (and soon we'll get rid of Blueprint completely).


mpodlasin (approved):
OMG, yes thank you!

I was thinking about it recently when looking at that storybook config.

Issue #1049, opened 2021-04-13
Create `elements-dev-portal` and remove SP from `elements`


marcelltoth:
*Epic: #996*

## Summary

As described in the epic, we would like to create another package called `elements-dev-portal` that holds the `StoplightProject` component and its accessories. Then let's make sure there is no leftover SP-related code in `elements`.

## Acceptance criteria

- No mention of `StoplightProject` in `elements`, it's accessories are removed, too.
- `elements-dev-portal` built and tested on CI, just like `elements` is.
- `elements-dev-portal`'s artifact structure is analogous to `elements`'s, as [described here](https://github.com/stoplightio/elements/issues/1046).
- `elements-dev-portal` is published on NPM as an alpha / beta (same channel as `elements`)
- Documentation is updated
- Examples are updated
- E2E tests are updated
- `elements-dev-portal` has it's own Storybook, showing off some demo projects. (No internal components here.)

PR #1125, opened 2021-04-23
feat: separate stoplight project component into dev-portal package

mpodlasin:
Relates to #1049 

The main goal of this PR is to keep the changes *as small as possible*, in order to avoid conflicts and have something that works ASAP.

Most notably this PR:
* moves only `StoplightProject` component to the new `elements-dev-portal` package. In future PRs I'll probably move a few more.
* does not include README
* might involve some duplication that *could* be potentially reduced (storybook config files come to mind here).
* does not move any styles (still requiring users to import styles from `elements` package)


philsturgeon (approved):

mmiask (commented):
- [x] Elements build
- [x] Examples build
- [x] Examples run correctly (no crashing)
- [x] Demo builds and runs correctly

Other observations:
- `elements-dev-portal` Storybook opens to an error page
- `elements-dev-portal` web-components story does not run Docs
- Missing icons + styling issues (although as it is stated in this description, it's not in the scope of this PR)

Bear in mind that I did not yet go through the code, this is purely based on clicking around. I'll give more feedback once I read through it

examples/angular/angular.json
```diff
@@ -50,7 +51,7 @@
                 {
                   "type": "initial",
                   "maximumWarning": "2mb",
-                  "maximumError": "6mb"
+                  "maximumError": "9mb"
```
mallachari, line 54: What's the reason for that change? Was it breaking earlier?

packages/elements-dev-portal/package.json
```diff
@@ -0,0 +1,84 @@
...
+  "main": "",
+  "files": [
+    "**/*"
+  ],
+  "engines": {
+    "node": ">=10"
+  },
+  "scripts": {
+    "build": "yarn build.react && yarn build.webcomponents",
+    "build.react": "sl-scripts build",
```
mallachari, line 27: There is a `build.styles` that runs `build-styles.js` during postbuild in elements package. Don't we need that here too?

packages/elements-dev-portal/web-components.config.js
```diff
@@ -0,0 +1,29 @@
...
+    rules: [
+      {
+        test: /\.mjs$/,
+        include: /node_modules/,
+        type: 'javascript/auto',
+      },
+      {
+        test: /\.tsx?$/,
+        loader: 'ts-loader',
+        exclude: /node_modules/
```
mallachari, line 25:
```suggestion
        exclude: /node_modules/,
```

mallachari (commented):
Looks good to me.
@mmiask, I can't see what you described. Storybook seems fine. I don't see the error page and web-components load correctly. Can you still reproduce that?

examples/angular/angular.json
mpodlasin: We are now adding 2 web-components files - one for API and one for StoplightProject.

Shouldn't be an issue I think? I don't believe people will use *both* of those at the same time? :thinking:

mallachari: Looks like it's about bundle size. I guess we exceeded limit with new package added.

packages/elements-dev-portal/package.json
mpodlasin: I haven't moved any styles yet, I plan to do it in a separate PR.

After this PR folks still need to import styles from `@stoplight/elements`.

packages/elements-dev-portal/tsconfig.build.json
```diff
@@ -0,0 +1,21 @@
...
+    "**/__*__/**"
+  ],
+  "compilerOptions": {
+    "baseUrl": "./",
+    "outDir": "dist",
+    "moduleResolution": "node",
+    "module": "commonjs",
+    "paths": {
+      "@stoplight/elements": ["../elements/dist"],
+      "@stoplight/elements-utils": ["../elements-utils/dist"],
```
mmiask, line 16: Same as the other comment - why do we need those 2 lines?

mmiask (commented):
Looks good to me as well. I left a few questions, although they shouldn't block merging.

I'll approve once I confirm that Storybook indeed works as intended (re @mallachari comment)

mmiask:
@mallachari Storybook behavior is not consistent for me. These are the 3 different ways it runs for me. No special reproduction steps, just checkout this branch and run `yarn elements-dev-portal storybook`

Screenshot 2021-04-28 at 14 10 19

---

Screenshot 2021-04-28 at 14 11 31

---

Screenshot 2021-04-28 at 14 12 00

mpodlasin:
I am having the same issues as @mmiask 

Debugging currently.

tsconfig.json
```diff
@@ -6,7 +6,9 @@
     "baseUrl": "./packages",
     "paths": {
       "@stoplight/elements-utils": ["./elements-utils/src"],
-      "@stoplight/elements": ["./elements/src"]
+      "@stoplight/elements": ["./elements/src"],
```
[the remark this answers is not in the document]
mpodlasin: Yeah, tried it, and it crashes builds.

Presumably because `@stoplight/elements` actually doesn't fall under `@stoplight/elements/*` rule, because it doesn't end with `/`.

packages/elements-dev-portal/package.json
```diff
@@ -0,0 +1,84 @@
...
+    "type": "git",
+    "url": "https://github.com/stoplightio/elements"
+  },
+  "license": "Apache-2.0",
+  "main": "",
+  "files": [
+    "**/*"
+  ],
+  "engines": {
+    "node": ">=10"
```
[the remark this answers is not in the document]
mpodlasin: @marcelltoth was right. We shouldn't put massive restrictions here, because for the most people just using elements, those restrictions are not necessary.

These are more restrictions of our internal tools.

mallachari (approved):
Works fine now

mmiask (approved):
Approving! It still runs a bit slow if you open the new endpoint (which might be more noticeable because of StoplightProject broken styling, it actually also happens in API component), but we can address that in the next PR which will include the styling

Great job with separating these two parts, this is already a big step in having the new structure 🎉

Issue #836, opened 2021-02-03
Embed new Try it in Articles


marbemac:
As a docs writer, I can use the Stoplight Markdown syntax to include a request maker within my markdown articles.

<p></p>

It should continue to work the same way as the old Request Maker embed, through a [Stoplight Flavoured Markdown block](https://meta.stoplight.io/docs/studio/docs/Documentation/03a-stoplight-flavored-markdown.md#http-request-maker).

<p></p>

```
\```yaml http
{
  "method": "get",
  "url": "http://todos.stoplight.io/todos"
}
\```
```

The entire list of arguments for the http block is available on the documentation linked above.

![](https://pb-files.s3.amazonaws.com/production/attachments/df4b8c9fc64e8ff8eb847fcd2e0f7016a9f0a0fc8e65f44e02a97493e658a1a5/image.png)

<p></p>

PR #1134, opened 2021-04-26
feat: Embed TryIt in Articles

mallachari:
Closes: https://github.com/stoplightio/elements/issues/836

Note: The actual change is in http part of `CodeComponent` that can be seen in first commit. GH doesn't detect file rename here unfortunately.

image

### DIFF
--- a/packages/elements/src/components/MarkdownViewer/CustomComponents/CodeComponent.spec.ts
+++ b/packages/elements/src/components/MarkdownViewer/CustomComponents/CodeComponent.spec.ts
@@ -2,83 +2,81 @@ import { HttpParamStyles, IHttpRequest } from '@stoplight/types';
 
 import { parseHttpRequest } from './CodeComponent';
 
-describe('CodeComponent', () => {
-  describe('parseHttpRequest', () => {
-    it('parses HttpRequest with relative url', () => {
-      const request: IHttpRequest = {
-        method: 'get',
-        url: '/todos',
-        baseUrl: 'http://test',
-        query: {},
-        headers: {},
-      };
+describe('parseHttpRequest', () => {
+  it('parses HttpRequest with relative url', () => {
+    const request: IHttpRequest = {
+      method: 'get',
+      url: '/todos',
+      baseUrl: 'http://test',
+      query: {},
+      headers: {},
+    };
 
-      const httpOperation = parseHttpRequest(request);
+    const httpOperation = parseHttpRequest(request);
 
-      expect(httpOperation).toEqual({
-        id: '?http-operation-id?',
-        method: 'get',
-        path: '/todos',
-        servers: [{ url: 'http://test' }],
-        request: {
-          headers: [],
-          query: [],
-        },
-        responses: [],
-      });
+    expect(httpOperation).toEqual({
+      id: '?http-operation-id?',
+      method: 'get',
+      path: '/todos',
+      servers: [{ url: 'http://test' }],
+      request: {
+        headers: [],
+        query: [],
+      },
+      responses: [],
     });
+  });
 
-    it('parses HttpRequest with absolute url', () => {
-      const request: IHttpRequest = {
-        method: 'get',
-        url: 'http://test/todos',
-        baseUrl: '',
-        query: {},
-        headers: {},
-      };
+  it('parses HttpRequest with absolute url', () => {
+    const request: IHttpRequest = {
+      method: 'get',
+      url: 'http://test/todos',
+      baseUrl: '',
+      query: {},
+      headers: {},
+    };
 
-      const httpOperation = parseHttpRequest(request);
+    const httpOperation = parseHttpRequest(request);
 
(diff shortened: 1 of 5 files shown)


philsturgeon (approved):

packages/elements/src/components/MarkdownViewer/CustomComponents/CodeWithSchemaViewer.tsx
```diff
@@ -23,6 +26,48 @@ export const CodeWithSchemaViewer = (props: IComponentMappingProps<ICode<ICodeAn
     return <SchemaAndDescription title={annotations?.title} schema={parsedValue} />;
   }
 
+  if (nodeType === 'http') {
+    if (!isObject(parsedValue) || (!isHttpRequest(parsedValue) && !isHttpOperation(parsedValue))) {
+      return null;
+    }
+
+    return <TryIt httpOperation={isHttpOperation(parsedValue) ? parsedValue : parseHttpRequest(parsedValue)} />;
```
marcelltoth: Shouldn't we rather just render the `DefaultCode` in the failure path? Not a strong opinion here though.

packages/elements/src/components/MarkdownViewer/CustomComponents/CodeComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+    request: {
+      ...(data.query && Object.keys(data.query).length
+        ? {
+            query: Object.entries(data.query).map(([key, value]) => ({
+              name: key,
+              style: HttpParamStyles.Form,
+              schema: { default: Array.isArray(value) && value.length > 0 ? value[0] : value },
+            })),
+          }
+        : null),
```
marcelltoth, line 57:
```suggestion
query: Object.entries(data.query).map(([key, value]) => ({
        name: key,
        style: HttpParamStyles.Form,
        schema: { default: Array.isArray(value) && value.length > 0 ? value[0] : value },
      })),
```

Same logic applies to `headers`.

Now I know this is controversial a littlebit, as I've removed two of your checks.
1. `Object.keys(data.query).length`. I don't think this is needed as TryIt should work just fine with an empty array passed in (IIRC we default the `undefined`s to an empty array anyway.)
2. `data.query` should always be truthy per TypeScript. I know this may not be the case, but then let's fix those TS types instead. I think the core problem is back in `CodeComponent`. You use the type `IHttpRequest` but really, you are supporting `Pick<IHttpRequest, 'method' | 'url'> & Partial<IHttpRequest>`, aren't you? You can either propagate this change throughout your changes, or provide defaults for `baseUrl`, `headers` and `query` right in `CodeComponent` (which may be simpler). But the current type-guard approach is way too weak IMO and creates a lie in many real world use cases.

marcelltoth (commented):
Beyond what's written above, I'm missing a test for the actual markdown rendering. IDK what's the best point to test, it could be more of an integration test against `<Article>`, or whatever you feel fit, but I'd love to see something that actually attempts to render the thing.:)

packages/elements/src/components/MarkdownViewer/CustomComponents/CodeWithSchemaViewer.tsx
mallachari: So far we are rendering null for `json_schema`, we did render null in v6 for request maker too, so I didn't change that here. That might be worth considering though, but not sure if should be done here.

packages/elements/src/components/MarkdownViewer/CustomComponents/CodeComponent.tsx
mallachari: 1. I wanted to avoid adding that empty array but there's no problem with having it there anyway so I removed it.
2. That's correct. I've improved those types and the guard itself so it wouldn't break in case of some weird input.

mallachari:
@marcelltoth thanks for the comments. I've updated it and added a rendering test showing that it actually works.

packages/elements/src/components/MarkdownViewer/CustomComponents/CodeComponent.tsx
marcelltoth: Nice! Thank you!

packages/elements/src/components/MarkdownViewer/CustomComponents/CodeComponent.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { HttpParamStyles, IHttpRequest } from '@stoplight/types';
+
+import { parseHttpRequest } from './CodeComponent';
+
+describe('CodeComponent', () => {
+  describe('parseHttpRequest', () => {
```
marcelltoth, line 6: We don't usually apply the filename scope again as a `describe` block, AFAIK jest recognizes files as an organizational structure anyway.

marcelltoth (approved):
Thanks for adding the test!

PR #1135, opened 2021-04-26
chore: no more petstore

philsturgeon:


marcelltoth:
What's the rationale?

philsturgeon:
The rationale is the pet store is not a good example of anything so let’s not use it as an example for anything. People generally hate the pet store and I keep getting messages about it.

marcelltoth (approved):
I liked it as the simplest thing to try new features with first, but I get it

philsturgeon:
Yeah, keep the URL around along with some other good URLs. I have a bookmark folder in Firefox!

PR #1143, opened 2021-04-27
feat: add preliminary support for OAS 3.1

P0lip:
Needed by https://github.com/stoplightio/platform-internal/pull/6083

This PR is mostly about upgrading the Stoplight deps to the very latest versions and adjusting the code to reflect the breaking changes made to these deps.

I briefly tested it in Studio, but I have significantly less context than you all do, therefore a more thorough review would be appreciated.


P0lip:
Seems like something is off with circleci config - we probably shouldn't be running the `release` on such PRs

packages/elements/src/constants.ts
```diff
@@ -49,6 +49,7 @@ export const HttpSecuritySchemeColors: Dictionary<string, HttpSecurityScheme['ty
   http: 'orange',
   oauth2: 'red',
   openIdConnect: 'blue',
+  mutualTLS: 'grey',
```
P0lip, line 52: grey seems fine, no? :laughing:

philsturgeon:
Do we need to do anything about exclusiveMin/Max because they’re different now, right? JSV/Parameters should be a little different and we should delete the tag that shows Nullable as that’ll never show up anymore.

P0lip:
> Do we need to do anything about exclusiveMin/Max because they’re different now, right?

Probably not. This is how it looks like after the changes
image



>  JSV/Parameters should be a little different and we should delete the tag that shows Nullable as that’ll never show up anymore.

Hmm, which tag? Wasn't it implemented in json schema viewer?
Bear in mind this is V6 PR rather than V7. I believe it's V7 that has tags. If so, in V7 the tag is going to be be removed, assuming you mean the schema viewer.

packages/elements/src/constants.ts
marcelltoth: Grey is always fine, except that [in Tailwind it is `gray`](https://tailwindcss.com/docs/text-color) (unless we have it overridden)

We use `blue` for it in `v7` though, so you may just pick that.

packages/elements/src/utils/guards.ts
```diff
@@ -1,8 +1,8 @@
 import { IHttpOperation, IHttpRequest, IHttpService } from '@stoplight/types';
-import { JSONSchema4 } from 'json-schema';
+import { JSONSchema7 } from 'json-schema';
 import { isObject } from 'lodash';
 
-export function isJSONSchema(maybeSchema: unknown): maybeSchema is JSONSchema4 {
+export function isJSONSchema(maybeSchema: unknown): maybeSchema is JSONSchema7 {
```
marcelltoth, line 5: Type safety FTW 🔒 😂

package.json
```diff
@@ -12,6 +12,9 @@
   "workspaces": [
     "packages/*"
   ],
+  "resolutions": {
+    "@stoplight/json-schema-sampler": "0.2.0"
```
marcelltoth, line 16: Why the need for the resolution? Do resolutions make sense in a lib? Are they published? If so, does NPM handle them?

marcelltoth (commented):
Looking great in general, thank you for the contribution!

marcelltoth:
What? :O Why is `release` even running here? @stoplightio/undefined

mallachari:
> > Do we need to do anything about exclusiveMin/Max because they’re different now, right?
> 
> Probably not. This is how it looks like after the changes

The change with exclusiveMin/Max is that it's not boolean anymore, rather a number. So that might (or might not) affect validations in jsv and parameters.

> > JSV/Parameters should be a little different and we should delete the tag that shows Nullable as that’ll never show up anymore.
> 
> Hmm, which tag? Wasn't it implemented in json schema viewer?
> Bear in mind this is V6 PR rather than V7. I believe it's V7 that has tags. If so, in V7 the tag is going to be be removed, assuming you mean the schema viewer.

Nullable tag indeed exists in new JSV, it doesn't apply to parameters validations, so it shouldn't matter here.

packages/elements/src/constants.ts
P0lip: `blue` is occupied by `openIdConnect`, thus perhaps `yellow` provided that it's readable enough?

packages/elements/src/utils/guards.ts
P0lip: :rofl: I have to admit I love these guards.

P0lip:
> The change with exclusiveMin/Max is that it's not boolean anymore, rather a number. So that might (or might not) affect validations in jsv and parameters.

It doesn't, since we just naively take a number and render it, so even if a string or any other type was there, we'd attempt to display it.
Or am I missing some other usage of schema that's exclusive to elements?

package.json
P0lip: Prism has an explicit version set, and that's why.
I decided to explicit version, because that's what npm suggests - theoretically all `0.x.x` can come up with breaking changes, as this particular major version means the API is not stable yet, hence I wanted to make sure Prism is safe.

>  Do resolutions make sense in a lib?

I added them only to make the build passing.
Similarly to Platform, Elements use an older version of Webpack that's not capable of recognizing optional null operator, that caused the build to fail

I fixed that in a newer version -> https://github.com/stoplightio/json-schema-sampler/commit/e096188662e08bd16e9f4982e69cd27f36b1df25 and added a few useful features altogether, i.e. https://github.com/stoplightio/json-schema-sampler/commit/bdd5ae5a6e57289ef2508f8abb3065231d2d88a1

I'll try to update this dependenct in Prism in order to remove that resolution.

> If so, does NPM handle them?

nope, ignored

mallachari:
There's something wrong with parameters though (with exclusiveMaximum).

image

while on v6:
image

Shouldn't be difficult fix, I'll check that

P0lip:
Thank you Jakub!

mallachari (approved):
Works fine, good job!

Issue #1047, opened 2021-04-13
Implement independent versioning of packages


marcelltoth:
*Epic: #996*

## User story.

As a developer, I can do publish packages independently, so that I can e.g. release `@stoplight/elements-utils` (later `@stoplight/elements-core` without having to release `@stoplight/elements` with a potentially empty changelog.

Read the epic for details on versioning policy. Implement that.

## Caveat

Lerna's independent mode may not be the best. I've looked around for alternatives but there's not much around excluding those _do it all_ tools with a high vendor-lock in factor. 

We'll probably need to pick some pieces from lerna and write some custom scripts, just like how we currently have `release.sh`. If this turns out to be too hard, we can loosen the versioning requirements for the internal packages.

⚠️ When experimenting with `lerna version` make sure to supply the `--no-push` argument, otherwise it will try to push to GitHub.

PR #1144, opened 2021-04-28
Chore: independent package versioning

mmiask:
Resolves: #1047 

After a call with @marcelltoth, we agreed that having a semi-auto release for all of the packages is enough (and actually only possible thing to do)

Because we lose commits from feature branches after squash, it's impossible to determine the type of changes introduced. Let's say that in a PR we change both `elements-utils` and `elements`, and both should have different version bumps (fix for `elements` and minor for `elements-utils`). Single commit message from `main`, after merging such PR does not reflect the whole spectrum of changes.

For now, the process will look like that:

1. We bump versions locally in a feature branch
2. We open a PR and merge that branch to `main`
3. CI job will automatically release packages for us


mpodlasin (approved):
Looks straightforward enough.

Issue #1130, opened 2021-04-26
"Powered by Stoplight"


philsturgeon:
## Summary

We were so busy working on features and stability we forgot to sign our work! 

Here's how it should look:

![endpoint](https://user-images.githubusercontent.com/67381/116061398-b4888980-a67a-11eb-81cd-40682e520d6f.png)

We already have similar in hosted docs, but we're going with purple here. 

Link should be:

```
https://stoplight.io/?utm_source=elements&utm_medium=api&utm_campaign=powered_by&utm_content=/
```

philsturgeon:
@smf434 @mnaumanali94 how does that look for the UTM link? ☝🏻

mnaumanali94:
Could the medium potentially be the API name here? It would be interesting to see where it's coming from

philsturgeon:
@mnaumanali94 by api name do we mean info.title? We can do that. It’ll have a bunch of duplicates like User API, but it can give us some idea of what APIs are doing well I suppose? Anything more specific would be invasive.

mnaumanali94:
That aligns with what I was thinking. 💯

Issue #1021, opened 2021-04-07
[StoplightComponent] ...has completely broken layout


mpodlasin:
## Describe the bug

As a consequence of big `API` component refresh/refactor, layout of `StoplightComponent` is completely broken.

*When fixing, it would be preferable to extract the layout part of `API` into separate component and using this in `StoplightComponent`, so that those two components never go out of sync again.*

## To Reproduce

1. Look at StoplightComponent :joy: 

![Screenshot from 2021-04-07 14-26-57](https://user-images.githubusercontent.com/7916523/113866224-58ce8d00-97ad-11eb-9b5c-a0869dc30604.png)

PR #1146, opened 2021-04-28
feat: add "Powered by Stoplight" branding

mmiask:
Resolves: #1130


mpodlasin (commented):
Screenshot from 2021-04-30 10-03-19
Screenshot from 2021-04-30 10-03-12

Can we bring it at least a bit closer to design? I am concerned mostly with padding.

Also - nothing in stacked layout? StoplightProject? It doesn't have to be in this PR but I would at least make issues to add it in those other places. Because we will forget again. :sweat_smile:

mmiask:
> Can we bring it at least a bit closer to design? I am concerned mostly with padding.

@mpodlasin Sure thing, I'll fix the padding 👍 Is there anything else that stands out to you? I think it looks pretty much the same atm

> Also - nothing in stacked layout? StoplightProject?

I don't think there's a need to put it in `StoplightProject`, since people that will use it already have a Stoplight account - no need to further advertise. @philsturgeon - what do you think?

As for the stacked layout - I agree that it might be a good idea, but I couldn't think of anything that wouldn't look awful. One place it fits is at the bottom of the page, but at the same time I think most users will miss it if it's there

mpodlasin:
Yeah, sorry, only padding.

philsturgeon:
I'll make a followup issue for doing this is in the stacked layout as that needs design work, but if you could get it into StoplightProject that would be handy. That said as #1021 is not done yet that might not be possible, so if its still a mess please make an issue yourself for followup work.

philsturgeon:
Oh yeah @mmiask the URL should be `https://stoplight.io/?utm_source=elements&utm_medium=${info.title}&utm_campaign=powered_by&utm_content=${elements-path-url}

mpodlasin (approved):
:+1:

mmiask:
@philsturgeon @mnaumanali94 for `StoplightProject` I modified the url a little bit. Please let me know if that works for you:

**StoplightProject**:```https://stoplight.io/?utm_source=elements-dev-portal&utm_medium=${source}&utm_campaign=powered_by&utm_content=${pathname}```

example:
```
https://stoplight.io/?utm_source=elements-dev-portal&utm_medium=elements/studio-demo&utm_campaign=powered_by&utm_content=/reference/petstore/openapi.v1.yaml/components/schemas/Pets
```

---

**API** component: ```https://stoplight.io/?utm_source=elements&utm_medium=${source}&utm_campaign=powered_by&utm_content=${pathname}```

example:
```
```

philsturgeon (approved):

packages/elements/src/components/PoweredByLink.tsx
```diff
@@ -0,0 +1,32 @@
+import { faBolt } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import * as React from 'react';
+
+export const PoweredByLink: React.FC<{
+  headless?: boolean;
+  source: string;
+  pathname: string;
+  isStoplightProject?: boolean;
```
mpodlasin, line 9: Is there a way to avoid that? I would prefer to avoid mentioning `StoplightProject` in a project that - at least in principle - should know nothing about `StoplightProject`.

philsturgeon: Could be: package: elements | elements-dev-portal

marcelltoth: Could also be: `package: string`, then you have 0 coupling.

mpodlasin: @mmiask yes, I also would prefer what @marcelltoth proposed.

PR #1148, opened 2021-04-29
fix: release

mallachari:
Fixes releasing on every commit issue


marcelltoth (approved):
Let's hope this works 😅

PR #1150, opened 2021-04-29
feat: add styles to elements-dev-portal

mpodlasin:
Relates to https://github.com/stoplightio/elements/issues/1049


philsturgeon (approved):

mallachari (approved):
Perfect! Works fine!

examples/angular/angular.json
```diff
@@ -25,7 +25,7 @@
             ],
             "styles": [
               "src/styles.css",
-              "node_modules/@stoplight/elements/styles/elements.min.css"
+              "node_modules/@stoplight/elements-dev-portal/styles/elements-dev-portal.min.css"
```
mallachari, line 28: That's the only thing that I don't get here. Why is it `elements-dev-portal` here and not `elements`? We still build both from what I can see

PR #1151, opened 2021-04-29
fix: incorrect lerna command in CirlcleCI job

mmiask:
Recent failing of a workflow in CircleCI exposed an error in the config file. This PR addresses that, simply changing `--from-package`(which doesn't exist) to `from-package` (learn more at: https://github.com/lerna/lerna/tree/main/commands/publish#bump-from-package)


mpodlasin (approved):
:rocket:

PR #1152, opened 2021-04-29
fix: icons missing in angular example

mmiask:
We somehow overlooked missing icons in the `MosaicTableOfContents`. The icons are missing because that ToC uses `fal` prefix that refers to `light` icons that are the part of the `Pro` version of Fontawesome (which we don't want to use in Elements since it's open-source)

Before:

Screenshot 2021-04-29 at 20 02 31

After:

Screenshot 2021-04-29 at 19 55 53


packages/elements/src/components/MosaicTableOfContents/TableOfContents.tsx
```diff
@@ -24,6 +24,14 @@ import {
 const ActiveIdContext = React.createContext<string | undefined>(undefined);
 const LinkContext = React.createContext<CustomLinkComponent | undefined>(undefined);
 
+const NODE_TYPE_META_ICON: { [nodeType: string]: IIconProps['icon'] } = {
```
mpodlasin, line 27: Why those constants were moved here though? :thinking:

philsturgeon (commented):
Is there a test we can add that makes sure icons are loading?

packages/elements/src/components/MosaicTableOfContents/TableOfContents.tsx
mmiask: There's no need to update Mosaic directly. What I did here was just moving the constants to Elements and using the `fas` prefix which indicates the free package from `FontAwesome`

mmiask: Mosaic constants use `fal`, which points to a Pro version. Elements doesn't handle the Pro version so it fails to display these icons

mpodlasin: I still don't understand. There is `./constants` import. This file keeps... well, constants, as the name suggests.

And yet you moved *some* constants from that `./constants` file to the main file, while you left some others.

I don't fully understand why.

mpodlasin: Screenshot from 2021-04-30 15-00-10

In other words, after this change, those constants are left unused. Why?

mpodlasin: Aaaaah, I know where you are confused.

Michael, those constants are *not* mosaic constants. Take a closer look where this file is located :joy:

mpodlasin:
Can we take a look at `react-cra` as well? Last time I've seen it, it didn't have icons.

marcelltoth:
It's currently broken in Storybook, too, which means everywhere I guess.

mmiask:
> Is there a test we can add that makes sure icons are loading?

@philsturgeon I initially thought that we could achieve that pretty easily, but unfortunately that's not the case. This is what I discovered:

- It's absolutely doable to query against FontAwesome icons using their `[data-icon]` attributes
- I wrote a simple test for catching missing icons
- At first it seemed to work, but when I looked into the `StoplightProject`, the `cloud` icons were missing (the ones that are displayed next to an API document entry in ToC) and at the same time the tests were passing (`<svg>` element with `[data-icon=cloud]` was present but it didn't contain the correct `<path>` value pointing to the actual graphic)

The alternative is to have some visual testing tools. We can definitely consider that, but I'd say for now let's make sure we test them manually for each PR.

marcelltoth (approved):
Works great for me now, thanks @mmiask !

philsturgeon (approved):
Ok! "tests are harder than hell to add so will think about doing it when we have visual tests" is an ok outcome for me. Thanks for looking into it @mmiask!

PR #1153, opened 2021-04-30
fix: make elements-dev-portal private

mmiask:
This PR makes `elements-dev-portal` private for now, which should prevent the unwanted release from `lerna`.


mpodlasin (approved):
:crossed_fingers:

PR #1154, opened 2021-04-30
chore: remove mentions of StoplightProject from elements

mpodlasin:
Relates to #1049


philsturgeon (approved):

packages/elements/src/containers/Provider.tsx
```diff
@@ -1,15 +1,9 @@
...
+import { IActiveInfo } from '../types';
 
-export interface IProvider extends IActiveInfo {
-  authToken?: string;
-  isStoplightProjectComponent?: boolean;
-}
-
-const defaultIcons: NodeIconMapping = {};
-export const IconsContext = createNamedContext<NodeIconMapping>('IconsContext', defaultIcons);
+export interface IProvider extends IActiveInfo {}
```
mmiask, line 6:
```suggestion
```

No need to duplicate - `IProvider` doesn't actually extend `IActiveInfo` here

mmiask: After removing it, you can use `IActiveInfo` in `Provider` instead of `IProvider`

packages/elements/src/containers/Provider.tsx
```diff
@@ -1,15 +1,9 @@
...
-import { IActiveInfo, NodeIconMapping } from '../types';
+import { IActiveInfo } from '../types';
 
-export interface IProvider extends IActiveInfo {
-  authToken?: string;
-  isStoplightProjectComponent?: boolean;
-}
-
-const defaultIcons: NodeIconMapping = {};
-export const IconsContext = createNamedContext<NodeIconMapping>('IconsContext', defaultIcons);
```
mmiask, line 12: Where was this used before? Elements seem to be working fine without it

mmiask (approved):
Looks good! I left one suggestion, but it shouldn't block the PR. Great job! 🎉

packages/elements/src/containers/Provider.tsx
mpodlasin: That's the thing - I haven't found a single usage of that :joy:

packages/elements/src/containers/Provider.tsx
mpodlasin: I am leaving this for now.

It might prove to be useful soon.

If not, I will be voting for removing Provider entirely, since it only wraps 1 other provider and does nothing else.

Issue #1116, opened 2021-04-22
Upgrade to http-spec@v4 for OAS3.1 Support


philsturgeon:
Upgrading to http-spec v4 will take care of https://github.com/stoplightio/elements/issues/1017 and a whole lot of other rubbish that we won't need to think about.

It's a beta right now but that's ok. 

There will be a few breaking changes: 

- exclusiveMinimum and exclusiveMaximum are integers not boolean modifiers
- It returns JSONSchema7 not JSONSchema4
- It's ditched openapi-sampler for @stoplight/json-schema-sampler and we might want to do the same
- Delete the `Nullable` tag, it should be normalized as "string or null" now. See https://github.com/stoplightio/elements/issues/1017 for more context.

P0lip:
@marcelltoth I think you won't need to do anything here - I'm taking care of it at the moment, because I need it for Studio.
There might be some changes you'll want to address separately, but hopefully minor.

marcelltoth:
@P0lip AFAICT your PR targets `v6`, so while this became much simpler because of it, we still need to port it to v7. Keeping this open still.

P0lip:
> @P0lip AFAICT your PR targets v6, so while this became much simpler because of it, we still need to port it to v7. Keeping this open still.

Yup, I know. I had started the work on v7 (https://github.com/stoplightio/elements/commit/6016ca80380322d0ae74d04e532b6808b26aff3f, just pushed the latest code I had) before Jakub J picked it up, but when I realized he's on it, I decided not to interfere.
You shouldn't need any futher changes in JSV, because I took care of it.

marcelltoth:
>  I'd like to stress we need this to be implemented ASAP, so if you need a hand, LMK.

That shouldn't really be the case, see Slack message. That being said, @mallachari 's PR should be there soon

PR #1155, opened 2021-04-30
feat: support for oas 3.1

mallachari:
Closes: https://github.com/stoplightio/elements/issues/1116
Requires: https://github.com/stoplightio/json-schema-tree/pull/9 for `Nullable` support

Updates `http-spec` and `types` dependencies
Replaces `openapi-sampler` with `@stoplight/json-schema-sampler`
Updates code to  support JSONSchema7

Note: JSONSchema7 lacks `[k: string]: any;` property type (it was present in JSONSchema4) which restricts it from using custom properties. That's the reason of removing `x-tags` and few others from fixtures (they weren't used anyway).


yarn.lock
```diff
@@ -3884,7 +3884,7 @@
   dependencies:
     eslint-config-prettier "^7.1.0"
 
-"@stoplight/http-spec@^3.1.1", "@stoplight/http-spec@^3.2.6":
+"@stoplight/http-spec@^3.1.1":
```
philsturgeon, line 3887: Can we get rid of this lockfile entry somehow?

yarn.lock
```diff
@@ -17065,7 +17094,7 @@ open@^7.0.2, open@^7.0.3:
     is-docker "^2.0.0"
     is-wsl "^2.1.1"
 
-openapi-sampler@^1.0.0-beta.16, openapi-sampler@^1.0.0-beta.18:
```
philsturgeon, line 17068: Is this still needing to be here?

philsturgeon:
Here's a test OAS I use to show off some of the new OAS3.1 things, and im looking for exclusiveMinimum support.

You might need to log into the meta workspace for this URL to work, but https://deploy-preview-1155--stoplight-elements-demo.netlify.app/?spec=https://stoplight.io/api/v1/projects/meta/different-openapi-versions/nodes/reference/OAS3-1.yaml?branch=main&deref=optimizedBundle&token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJodHRwczovL3N0b3BsaWdodC5pby9qd3QvY2xhaW1zIjp7IngtaGFzdXJhLWRlZmF1bHQtcm9sZSI6Im5vZGVfZXhwb3J0IiwieC1oYXN1cmEtYWxsb3dlZC1yb2xlcyI6WyJub2RlX2V4cG9ydCJdLCJ4LWhhc3VyYS1icmFuY2gtbm9kZS1pZCI6Ijc3ODAyMDYifSwiaWF0IjoxNjE5NzgyNzcwLCJleHAiOjE2MjAyMTQ3NzAsImlzcyI6InN0b3BsaWdodCIsInN1YiI6Ii91c2Vycy9ib3QifQ.v63zMeZphzvL5LpN836ZagR4B1Pa2C0VY-81jQNxae0#/operations/post-user

The User model is reused on the request and the response, and it defines a minimum age, but we arent showing `>18` anywhere.

```
        age:
          type: number
          description: Hold old is the user at time of signup.
          exclusiveMinimum: 18
          examples:
            - 36
```

yarn.lock
marcelltoth: It's a false positive, from `demo`.

yarn.lock
marcelltoth: Same as above. I'll soon send a PR trying to fix this but it's just noise, shouldn't bother us for this PR.

packages/elements/src/components/Docs/HttpOperation/Body.tsx
```diff
@@ -47,7 +46,7 @@ export const Body = ({ body: { contents = [], description }, onChange }: BodyPro
 
       {isJSONSchema(schema) && (
         <Box>
-          <JsonSchemaViewer resolveRef={refResolver} schema={schema as JSONSchema4} viewMode="write" hideExamples />
+          <JsonSchemaViewer resolveRef={refResolver} schema={schema} viewMode="write" hideExamples />
```
marcelltoth, line 49: Finally! 💪

packages/elements/src/components/TryIt/FormDataBody.tsx
```diff
@@ -33,38 +33,36 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({ specification, value
...
-                  value={value instanceof File ? value : undefined}
-                  onChange={newValue =>
-                    newValue
-                      ? onChangeValues({ ...values, [parameter.name]: newValue })
-                      : onChangeValues(omit(values, parameter.name))
-                  }
-                />
-              );
-            }
+        {mapSchemaPropertiesToParameters(parameters).map(parameter => {
```
marcelltoth, line 36: Nice refactoring!

marcelltoth (approved):
Looks good here!

philsturgeon (approved):

mallachari:
> The User model is reused on the request and the response, and it defines a minimum age, but we arent showing `>18` anywhere.
> 
> ```
>         age:
>           type: number
>           description: Hold old is the user at time of signup.
>           exclusiveMinimum: 18
>           examples:
>             - 36
> ```

@philsturgeon the issue you spotted occurred because http-spec transformes schema as it was oas2/3 document. For it to be transformed correctly (or rather not transformed as it is already `JSONSchema7`) we need to add `jsonSchemaDialect` property to the document. I've added that change and it looks good now.

packages/elements/src/utils/oas/index.ts
```diff
@@ -26,11 +26,20 @@ const isOas3 = (parsed: unknown): parsed is OpenAPIObject =>
...
+  Number.parseFloat(String((parsed as Partial<{ openapi: unknown }>).openapi)) === 3.1;
+
 const OAS_MODEL_REGEXP = /((definitions|components)\/?(schemas)?)\//;
 export const MODEL_REGEXP = /schemas\//;
 export const OPERATION_REGEXP = /\/operations\/.+|paths\/.+\/(get|post|put|patch|delete|head|options|trace)$/;
 
 export function transformOasToServiceNode(apiDescriptionDocument: unknown) {
+  if (isOas31(apiDescriptionDocument)) {
+    apiDescriptionDocument.jsonSchemaDialect = 'http://json-schema.org/draft-07/schema#';
+    return computeServiceNode(apiDescriptionDocument, oas3SourceMap, transformOas3Service, transformOas3Operation);
```
marcelltoth, line 41: This is a bit shady to me, as we now are introducing a side-effect to the function (and potentially to `API` as a whole - you can pass the spec object directly). Can I suggest
```suggestion
    return computeServiceNode({...apiDescriptionDocument, jsonSchemaDialect: 'http://json-schema.org/draft-07/schema#'}, oas3SourceMap, transformOas3Service, transformOas3Operation);
```
instead?

mallachari: Sure, that works as well

PR #1156, opened 2021-04-30
chore: demo -> elements dependency

marcelltoth:
Fixes the way `demo` is referencing `elements`. This change makes lerna recognize the relationship which in turn makes the lockfile a bit cleaner. Not much else to gain, but nothing to lose either.

cc @philsturgeon

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -10,7 +10,7 @@
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
-    "@stoplight/elements": "*",
+    "@stoplight/elements": "7.0.0-beta.1",
     "@stoplight/mosaic": "1.0.0-beta.47",
     "history": "^5.0.0",
     "react": "16.14.0",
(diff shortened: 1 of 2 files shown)


philsturgeon (approved):

demo/package.json
```diff
@@ -10,7 +10,7 @@
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
-    "@stoplight/elements": "*",
+    "@stoplight/elements": "7.0.0-beta.1",
```
philsturgeon, line 13: Would `7.0.0@beta` make this less rigid so we dont have to do updates when new betas are out? 
```suggestion
    "@stoplight/elements": "7.0.0@beta",
```

philsturgeon (approved):

demo/package.json
marcelltoth: I'm intentionally making it more rigid. It's part of the lerna workspace, so whenever we release a new version, lerna will auto-bump this dependency, too. No manual updates needed.:)

mallachari (approved):
Nice, I like it!

PR #1161, opened 2021-05-04
feat: use Menu instead of Select for RequestSamples

### LINKED ISSUE #758: RequestSamples: Use the Menu component to improve language select UX
Use the Mosaic [Menu component](https://mosaic.vercel.app/docs/forms/menu) for the try it now language selector, so this single level select can be a two-level menu with language > library.

**Currently**

**After**

Keep all the same libraries, just get them showing up in this menu component.

marcelltoth:
Closes #758 

**Draft because**
There are two, well..., mosaic related issues blocking this:
1. I have no idea how to make it close on selection, other than strapping a key on it which is ugly and slow. It looks like any `MenuItem` that has `checked !== undefined` will leave the `Menu` open after the user clicks on it but this behavior is not documented, so I don't even know what the intended behavior is. Waiting on answer on Slack.
2. Just like JSV, `Menu` is suffering from the same Box-related performance problems, making it painfully slow even in relatively simple applications like this dropdown. It drops to about 6-7 FPS (in debug mode, while profiling) on my machine, just by hovering my mouse over items, without any CPU slowdown.


philsturgeon:
Whilst we wait for those two issues, it seems like this sub menu stuff might need a little rethinking too. 

image

Instead of a menuitem with `{}` maybe we could omit the whole menuitem.

packages/elements/src/components/RequestSamples/RequestSamples.tsx
```diff
@@ -36,24 +36,35 @@ export const RequestSamples = React.memo<RequestSamplesProps>(({ request }) => {
...
+            label="Request Sample Language"
+            trigger={
+              <Button size="sm" iconRight="caret-down">
+                Request Sample: {selectedLanguage} / {selectedLibrary}
+              </Button>
+            }
+          >
+            {Object.entries(requestSampleConfigs).map(([language, config]) => (
+              <MenuItem key={language} indent text={language} checked={selectedLanguage === language}>
+                {Object.keys(config.libraries ?? {}).map(library => (
```
mallachari, line 53:
```suggestion
                {config.libraries && Object.keys(config.libraries).map(library => (
```

That would prevent rendering empty array

packages/elements/src/components/RequestSamples/RequestSamples.tsx
```diff
@@ -36,24 +36,35 @@ export const RequestSamples = React.memo<RequestSamplesProps>(({ request }) => {
...
+          <Menu
+            label="Request Sample Language"
+            trigger={
+              <Button size="sm" iconRight="caret-down">
+                Request Sample: {selectedLanguage} / {selectedLibrary}
+              </Button>
+            }
+          >
+            {Object.entries(requestSampleConfigs).map(([language, config]) => (
+              <MenuItem key={language} indent text={language} checked={selectedLanguage === language}>
```
mallachari, line 52: We might want to add `onClick` here as well in case it's only one level menu

mallachari (commented):
Good attempt though we need to have that closing solved first

marcelltoth:
ouch, I've meant to make this a draft (as indicated in the description). Thanks for the reviews nevertheless, I've now updated the component to fix the problems you highlighted, and also to work around Mosaic problem no 1. No 2 is still there but that's something we can live with until they update it.

mallachari (approved):
Perfect!
One thing that I might change is removing that button border. For me previous one looked better. What do you think?
Also looks like tests need some update :)

Issue #1158, opened 2021-05-03
Make schema view "flatten" allOf consistently


philsturgeon:
## Describe the bug

We might not be using allOf consistently when rendering schemas in request/response vs schema view.

## To Reproduce

When looking at the requestBody schema from the [List sub accounts](https://elements-demo.stoplight.io/?spec=https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/zoom.yaml#/operations/accounts) endpoint of the Zoom API the schema is "merged" into a single object.

![image](https://user-images.githubusercontent.com/67381/116869959-f6cd4000-ac11-11eb-92b5-76e900f72ff4.png)

However, when looking at the [same schema directly](https://elements-demo.stoplight.io/?spec=https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/zoom.yaml#/schemas/AccountList) the schemas are not merged. You can see it's composed of multiple schemas. 

![image](https://user-images.githubusercontent.com/67381/116869966-f9c83080-ac11-11eb-88cf-11cdc26eab85.png)

## Expected behavior

The two renderings of the same schema should look the same, but we might be handling allOf differently in these two different locations.

raleigh04:
Phil says this should be fixed for GA (end of May).

Is this JST or something else?

@P0lip thoughts?

P0lip:
It's not JST. It's an issue with elements.
This is the culprit
See https://github.com/stoplightio/elements/blob/main/packages/elements/src/components/SchemaAndDescription.tsx#L45L46

![image](https://user-images.githubusercontent.com/9273484/117064148-18b7e700-ad26-11eb-940c-6775c7dadcc0.png)
![image](https://user-images.githubusercontent.com/9273484/117064201-2e2d1100-ad26-11eb-84bd-9f27f5fd2c5e.png)

PR #1166, opened 2021-05-04
fix(schema-and-description): do not pass default title & description

P0lip:
Fixes https://github.com/stoplightio/elements/issues/1158


packages/elements/src/components/SchemaAndDescription/index.tsx
```diff
@@ -39,13 +38,7 @@ export const SchemaAndDescription = ({
       <JsonSchemaViewer
         resolveRef={resolveRef}
         className={cn(className, CLASSNAMES.block)}
-        schema={
-          {
-            ...schema,
-            title: undefined,
-            description: undefined,
```
mallachari, line 46: This was done on purpose as we didn't want to display title and description here. It could be just removed from original schema object, but we actually don't need to do that anymore so that change is fine

mallachari (approved):
Good and honest work!

philsturgeon (approved):

philsturgeon:
Great work! Let's get it in :D

PR #1167, opened 2021-05-05
fix: stoplight project layout

### LINKED ISSUE #1021: [StoplightComponent] ...has completely broken layout

mpodlasin:
Fixes #1021


mallachari (approved):
Finally it look good!

Issue #1136, opened 2021-04-26
Operation: Securty Badges no longer clickable


philsturgeon:
## Describe the bug

Security badges should link to their definition on the overview page. They used to work, so this seems like a regression. 

## To Reproduce

1. Given this OpenAPI document https://elements-demo.stoplight.io/?spec=https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/zoom.yaml#/operations/meeting
2. Try clicking API Key
3. Cant click it

## Expected behavior

Should link to the Overview tab and expand the appropriate panel. Test it on something with multiple security definitions and make sure it scrolls to the right panel.

## Additional context

**Screenshots**

<img width="1103" alt="Screen Shot 2021-04-26 at 2 59 14 PM" src="https://user-images.githubusercontent.com/67381/116095253-28895880-a6a0-11eb-81da-168123f86dd3.png">

PR #1168, opened 2021-05-05
fix: clickable security badges

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1136

We introduced additional operations uris `/operations/{operatioName}` some time ago that weren't reflected in checking service uri for linking security badges. This fixes it.


marcelltoth (approved):
Awesome, even has tests added, thank you!

packages/elements/src/utils/oas/security.ts
```diff
@@ -33,7 +33,7 @@ const getOauthScopeMapper = (securityScheme: IOauth2SecurityScheme) => (flow: st
 };
 
 export function getServiceUriFromOperation(uri: string) {
-  const match = uri?.match(/(.*)\/paths/);
+  const match = uri?.match(/(.*)\/(paths|operations)/);
```
philsturgeon, line 36: This feels potentially brittle as we might add more URL shortcuts and cleaning in the future, but I accept that this is a bug fix and not time for tech debt. If anyone can think of anything smart to do please make an issue with ideas, but otherwise this is a good change. Thank you for adding a test!!

philsturgeon (approved):

PR #1171, opened 2021-05-06
fix: prepend origin to urls without it

### LINKED ISSUE #1075: Another wild "unable to generate request sample" appears
## Describe the bug

Found an API failing to make a request sample, so lets see if we can figure out the cause in case it effects other folks too. 

## To Reproduce

1. Given this OpenAPI demo: https://elements-demo.stoplight.io/?spec=https://api.apis.guru/v2/specs/britbox.co.uk/3.730.267-ref-1-39-0/openapi.yaml#/operations/getBookmarks
2. See error

## Expected behavior

Request sample should show something useful even if its just a simple GET.

## Additional context

**Screenshots**
image

mpodlasin:
Fixes #1075

### DIFF
--- a/packages/elements/src/__fixtures__/services/without-origin.ts
+++ b/packages/elements/src/__fixtures__/services/without-origin.ts
@@ -1,0 +1,69 @@
+import { IHttpService } from '@stoplight/types';
+
+export const httpServiceWithoutOrigin: IHttpService = {
+  id: '?http-service?',
+  name: 'Petstore (OpenAPI v3)',
+  description:
+    'This is a sample server Petstore server.  You can find out more about Swagger at [http://swagger.io](http://swagger.io) or on [irc.freenode.net, #swagger](http://swagger.io/irc/).  For this sample, you can use the api key `special-key` to test the authorization filters. \n\n',
+  termsOfService: 'https://terms-of-service.com',
+  contact: {
+    name: 'Support',
+    url: 'https://stoplight.io',
+    email: 'support@stoplight.io',
+  },
+  license: {
+    name: 'Apache 2.0',
+    url: 'https://www.apache.org/licenses/LICENSE-2.0.txt',
+  },
+  version: '1.0.0',
+  servers: [
+    {
+      url: '/api',
+      name: 'Production API',
+    },
+    {
+      url: 'https://api.staging.stoplight.io',
+      name: 'Staging API',
+    },
+    {
+      url: 'https://api.int.stoplight.io',
+      name: 'Integration API',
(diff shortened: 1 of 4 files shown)


packages/elements/src/__fixtures__/services/without-origin.ts
```diff
@@ -0,0 +1,69 @@
+import { IHttpService } from '@stoplight/types';
+
+export const httpServiceWithoutOrigin: IHttpService = {
+  id: '?http-service?',
+  name: 'Petstore (OpenAPI v3)',
```
philsturgeon, line 5: Can we keep petstore usage to a minimum? It's a nit of a running joke in the OAS community, and... not a good joke.

mpodlasin: Honestly I just copied some other fixture and modified it for the specific test.

Do you want me to change it?

philsturgeon:
If you could quickly change it, let’s have a “no new pet stores” policy. :)

--
Phil Sturgeon
Product @ Stoplight.io

packages/elements/src/utils/http-spec/IServer.ts
```diff
@@ -3,6 +3,10 @@ import { IServer } from '@stoplight/types';
 export const getServerUrlWithDefaultValues = (server: IServer): string => {
   let url = server.url;
 
+  if (url[0] === '/' && typeof window !== 'undefined') {
```
mallachari, line 6: We may also want to do that when there is no `/`. In such case it also shows nothing in `API Base URL` part and breaks RequestSamples. Unless we want that to happen for these kind of inputs.

packages/elements/src/__fixtures__/services/without-origin.ts
```diff
@@ -0,0 +1,69 @@
...
+      url: 'https://api.int.stoplight.io',
+      name: 'Integration API',
+    },
+    {
+      url: 'https://localhost:4060',
+      name: 'Development API',
+    },
+  ],
+  security: [],
+  securitySchemes: [
```
mallachari, line 38: Those security schemes seem unnecessary here

mallachari (commented):
Works fine with the case described in issue. It doesn't when we remove `/` - see the comment. Unless that's expected behavior, if so then I'll accept :)

mpodlasin:
@mallachari @philsturgeon fixed your comments, please take a second look :)

philsturgeon (approved):

mallachari (approved):
:+1:

Issue #1048, opened 2021-04-13
Split `elements` into `elements` and `elements-core`


marcelltoth:
*Epic: #996*

## Summary

As described in the epic, extract another package called `elements-core` that holds the common material from `API` and `StoplightProject`. Also split the storybook -> move any internal component demos to `@stoplight/elements-core`'s storybook instance (which you should create).

### Out of scope

Creating `elements-dev-portal`. For this issue, keep both `API` and `StoplightProject` in `elements`.

## Additional information

This is ideally done after #1047 so you don't have to deal with setting up publishing for the new package, but can be started earlier to better parallelize work.

marcelltoth:
@mallachari I don't think this should be closed yet, correct?

philsturgeon:
We can close this now as we have elements-core published as a beta.

PR #1178, opened 2021-05-06
feat: elements-core

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1048
Not yet completed but good enough for first check.

`elements-core` repo extracted from elements containing core components for both `elements` and `elements-dev-portal`:
- `elements` keeps `API` container and components along with utils for computing oas nodes 
- some components/containers could be moved to `elements-dev-portal` (unless we want to keep them internal)

To do:
- [x] `elements-core` storybook broken styles (`elements` and `elements-dev-portal` look fine)
- [x] final cleanup of all repos
- [ ] possibly some components cleanup (though I didn't want to do much of that to not over complicate this PR)
- [x] `build.docs` runs locally but not in circleci
- [ ] ...


packages/elements-core/package.json
```diff
@@ -0,0 +1,126 @@
+{
+  "name": "@stoplight/elements-core",
+  "version": "7.0.0-beta.1",
```
mpodlasin, line 3: Since business decision was to start numbering `dev-portal` from `1.0.0` anyways, perhaps we could start from `1.0.0` here as well? :thinking:

marcelltoth: What? :o Do we start dev portal from 1.0.0? @philsturgeon

philsturgeon: Yep yep! It’s the first version of this extension package. Elements and Elements Dev Portal are different things and don’t need to have matching numbers.

philsturgeon: I don’t mind what you number the core and utils as. They’re not user facing so could be v7.0

packages/elements/.storybook/preview.jsx
```diff
@@ -3,11 +3,11 @@ import { Provider as MosaicProvider } from '@stoplight/mosaic';
 import { Title, Subtitle, Description, Primary, ArgsTable, PRIMARY_STORY } from '@storybook/addon-docs/blocks';
 import customTheme from './theme';
 
-import '../src/styles/elements.scss';
+import '../src/styles/elements-scoped.scss';
```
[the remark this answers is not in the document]
mallachari: As `elements.scss` and all the rest exceped `elements-scoped.scss` were moved to `elements-core`

packages/elements-dev-portal/src/styles/elements-dev-portal-scoped.scss
```diff
@@ -1 +1 @@
-@import '~@stoplight/elements/dist/styles/elements-scoped.scss';
\ No newline at end of file
+@import '~@stoplight/elements-core/dist/styles/elements-core-scoped.scss';
```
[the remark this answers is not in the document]
mpodlasin: That's not so obvious (?)

It's entirely possible that `dev-portal` while have it's own components (and hence styles) in the future, that's why currently it is built like that.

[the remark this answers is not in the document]
marcelltoth: True

marcelltoth (approved):

packages/elements-core/package.json
philsturgeon: Let's go with `elements-core@1.0.0` and let them both diverge over time.

philsturgeon (approved):

PR #1187, opened 2021-05-07
fix: misaligned mocking button

mpodlasin:


marcelltoth (approved):
I'll be honest, this is trust rather than a review, but go for it 🚀 

(We should really be deploying the `dev-portal` storybook to Netlify)

PR #1190, opened 2021-05-10
feat(dev-portal): provider component

lottamus:


mpodlasin (commented):
I am kind of confused though. :sweat_smile: 

Can you elaborate a bit more where and how this will be used?

Will this force users of `StoplightProject` to wrap it into `Provider`? :thinking: 

**EDIT:** I've read more PRs and it seems that the intention is to keep this Provider as an internal detail.

I would however like to avoid confustion. In `elements` package we have component called `Provider`, which is intended for external use. Therefore I would advocate for changing the name of this Provider, to not confuse the two.

`RootProvider`? `StoplightProjectProvider`? `InternalProvider`? :thinking: 

I also agree with @marcelltoth comments on slack that if we start using `react-query` we should remove `swr` usages from `elements`. In effect part of this Provider would actually land in `elements` package, possibly making this whole `Provider` component unnecessary to some degree.

marcelltoth (commented):
+1 for what @mpodlasin said.

I don't see any coherent features added in this PR, and also there is no *issue* that would describe what this is meant to serve. How am I supposed to decide if it's correct or not? 🙂

lottamus:
> Will this force users of StoplightProject to wrap it into Provider? 🤔

You can see where I'm using it within StoplightProject: https://github.com/stoplightio/elements/pull/1195/files#diff-9056b4f053ebcfec5961095382761a48313c53d5b4dc316b952be0004de1fbc9R91

> I would however like to avoid confustion. In elements package we have component called Provider, which is intended for external use. Therefore I would advocate for changing the name of this Provider, to not confuse the two.

I propose we change the name to `DevPortalProvider`. Does that work for ya'll?

> I also agree with @marcelltoth comments on slack that if we start using react-query we should remove swr usages from elements. In effect part of this Provider would actually land in elements package, possibly making this whole Provider component unnecessary to some degree.

Sure, would you like me to remove `swr` in this PR or a separate one?

> I don't see any coherent features added in this PR, and also there is no issue that would describe what this is meant to serve. How am I supposed to decide if it's correct or not? 🙂

This chunk of code was used in every other PR in the chain, since each one interacts with react-query + the platformUrl.

lottamus:
> Well it provides the platformUrl so maybe... PlatformUrlProvider? 😄

Hmm but it also provides the react-query client context, and in the future we may want to provide some other values? Is it better to have a single provider with lots of contexts? Or lots of providers?

We have the PlatformUrlContext too: https://github.com/stoplightio/elements/pull/1190/files#diff-70677ff8c89b0fe7ab32b1574715641135df6e8a83d32c399527bbbc017e4f22R16

mpodlasin:
> > Well it provides the platformUrl so maybe... PlatformUrlProvider? smile
> 
> Hmm but it also provides the react-query client context, and in the future we may want to provide some other values? Is it better to have a single provider with lots of contexts? Or lots of providers?
> 
> We have the PlatformUrlContext too: https://github.com/stoplightio/elements/pull/1190/files#diff-70677ff8c89b0fe7ab32b1574715641135df6e8a83d32c399527bbbc017e4f22R16

What I would do is remove `swr` in this PR, because this PR is IMHO small enough.

Then I would rename the `Provider` to `DevPortalProvider` **given** it wraps more than one provider.

If it only wraps `parameterUrl` thingy in the end - I would remove it.

lottamus:
@mpodlasin sounds good, I'll get it done shortly! 👍

packages/elements-core/src/hoc/withQueryClientProvider.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { QueryClient, QueryClientProvider } from 'react-query';
+
+import { getDisplayName } from './utils';
+
+const queryClient = new QueryClient({
```
philsturgeon, line 6: Does react-query respect the API's declared cacheability through Cache-Control headers or does it have its own client-side conventions for caching?

philsturgeon (approved):
This works just fine, and although I'd like to chat more about react-query sometime this PR aint the place. It's working on my machine, and its blocking stuff, so lets try and get this in today. :D

packages/elements-dev-portal/package.json
```diff
@@ -48,7 +48,8 @@
   "dependencies": {
     "@stoplight/elements-core": "^7.0.0-beta.1",
     "lodash": "^4.17.19",
-    "react-router-dom": "^5.2.0"
+    "react-router-dom": "^5.2.0",
+    "react-query": "^3.15.2"
```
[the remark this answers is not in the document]
marcelltoth: In the next PR `dev-portal` will need it, too, but you're right that it feels superfluous here, and it's certainly missing from core. It's also missing in `elements`.

packages/elements/src/containers/API.tsx
```diff
@@ -58,11 +56,17 @@ const propsAreWithDocument = (props: APIProps): props is APIPropsWithDocument =>
...
+  const { layout, apiDescriptionUrl = '' } = props;
   const apiDescriptionDocument = propsAreWithDocument(props) ? props.apiDescriptionDocument : undefined;
 
   const documentShouldBeFetched = apiDescriptionUrl && !apiDescriptionDocument;
-  const { data: fetchedDocument, error } = useSwr(documentShouldBeFetched ? apiDescriptionUrl! : null, fetcher);
+  const { data: fetchedDocument, error } = useQuery(
+    [apiDescriptionUrl],
+    () => fetch(apiDescriptionUrl).then(res => res.text()),
+    {
+      enabled: !!documentShouldBeFetched,
```
marcelltoth, line 67: Tiny suggestion: It'd be nicer if `documentShouldBeFetched` was an actual boolean instead.

packages/elements-core/src/hoc/withQueryClientProvider.tsx
marcelltoth: AFAICT this is one layer above Cache-Control and browser cache. It has it's in memory cache, which we configure. Then when it decides to make a request the browser may still respond from the cache it if has some.

marcelltoth (commented):
Now this is a great example of the principle of _single logical change -> single PR_ 👏 

Thank you @lottamus , looking forward to seeing it merged.

mpodlasin (approved):

PR #1191, opened 2021-05-10
feat(dev-portal): node content component

lottamus:
- Fixes https://github.com/stoplightio/platform-internal/issues/5645
- Adds handler to fetch node content
- Adds React hook for fetching node content

Can see example of this stuff working in our dev portal PR env: https://dev-portal-fuvmsk8zx-stoplight.vercel.app/docs/platform/ZG9jOjQ0-welcome-to-the-stoplight-docs


packages/elements-dev-portal/src/components/NodeContent/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const [resolvedUriWithoutAnchor, hash] = resolvedUri.split('#');
+    const edge = node.outbound_edges.find(edge => edge.uri === url || edge.uri === resolvedUriWithoutAnchor);
+
+    if (edge) {
+      return (
+        <Link to={edge.slug} hash={hash}>
+          {children}
+        </Link>
+      );
+    }
```
marcelltoth, line 49: What is this? 🤔

packages/elements-dev-portal/src/handlers/getNodeContent.ts
```diff
@@ -0,0 +1,34 @@
+import { Node } from '../interfaces/node';
+
+export const getNodeContent = async (
```
marcelltoth, line 3: In general we prefer colocating related code, rather than having functional folders. As you can probably see we are moving away from having folders like `utils`, `tests`, etc. 

IMO neither this folder or the function add any value, you could just move it into `useGetNodeContent`. Or actually I would consider inlining it, you don't really gain anything by declaring then destructuring parameters again.

packages/elements-dev-portal/src/handlers/getNodeContent.ts
```diff
@@ -0,0 +1,34 @@
...
+    throw new Error(data);
+  }
+
+  return data;
+};
+
+function getNodeIdFromSlug(nodeSlug: string) {
+  const [nodeId] = nodeSlug.split('-');
+  return nodeId;
+}
```
marcelltoth, line 34: Destructuring is nice, but not everything is a nail 😄 

```suggestion
function getNodeIdFromSlug(nodeSlug: string) {
  return nodeSlug.split('-')[0];
}
```

packages/elements-dev-portal/src/hooks/useGetNodeContent.ts
```diff
@@ -0,0 +1,24 @@
...
+  {
+    nodeSlug,
+    projectId,
+    branchSlug,
+  }: {
+    nodeSlug: string;
+    projectId: string;
+    branchSlug?: string;
+  },
+  requestHeaders?: Record<string, string>,
```
marcelltoth, line 17: I'm not sure I like this `requestHeaders` parameter. It's leaking the otherwise solid abstraction. Also as the rest are in a nicely structured object, this feels like something that was added last minute, feels like tech debt. What are you trying to achieve with it?

packages/elements-dev-portal/src/interfaces/node.ts
```diff
@@ -0,0 +1,28 @@
+export type Node = NodeSummary & {
```
marcelltoth, line 1: I agree it's kind of hard to colocate this with anything, that's for sure.

But it's certainly not `interfaces`, there isn't a single `interface` in this file. What about doing what we always do, and creating a top level `types.ts`, then forward-exporting the `Node` type from `index.ts`? That is probably the only thing from here that we really need.

packages/elements/src/containers/API.tsx
```diff
@@ -58,11 +56,17 @@ const propsAreWithDocument = (props: APIProps): props is APIPropsWithDocument =>
...
 
   const documentShouldBeFetched = apiDescriptionUrl && !apiDescriptionDocument;
-  const { data: fetchedDocument, error } = useSwr(documentShouldBeFetched ? apiDescriptionUrl! : null, fetcher);
+  const { data: fetchedDocument, error } = useQuery(
+    [apiDescriptionUrl],
+    () => fetch(apiDescriptionUrl).then(res => res.text()),
+    {
+      enabled: !!documentShouldBeFetched,
+    },
+  );
```
marcelltoth, line 69: I'm assuming this change will be handled another PR, right? The provider one.

packages/elements-dev-portal/src/handlers/getNodeContent.ts
lottamus: One use case for keeping them separate is if someone needs to call the handler outside a hook, such as in the NextJS dev-portal: https://github.com/stoplightio/dev-portal/blob/make-it-work/pages/docs/%5BprojectBranchSlug%5D/%5B%5B...nodeSlug%5D%5D.tsx#L61-L66

lottamus:
> I'm also a bit confused. This NodeContent component feels almost like a replacement of Docs. What is the difference? Is Docs still relevant? If not, why not just simply update it? Also this doesn't update StoplightProject itself, that is still using the old API and the old components. Shouldn't that switch to the new NodeContent + hook approach, too?

NodeContent is essentially a replacement for the Docs container, but allows you to pass data in from either the `useGetNodeContent` hook or the `getNodeContent` handler (both which interact with the new HTTP API). 

In an attempt to keep these PRs small (with minimal breaking changes), I've decided to not touch the StoplightProject nor remove any existing code until the end. https://github.com/stoplightio/elements/pull/1195

packages/elements-dev-portal/src/handlers/getNodeContent.ts
marcelltoth: Oh I see. That's why I said we really need to define the public API before we can reason about the code. Can you add the exports that you meant?

packages/elements-dev-portal/src/interfaces/node.ts
lottamus: Done in https://github.com/stoplightio/elements/pull/1191/commits/7b06bf468d589f7244993df9366dfe1c15a2f9a8

packages/elements-dev-portal/src/handlers/getNodeContent.ts
lottamus: Done in https://github.com/stoplightio/elements/pull/1191/commits/a04ea35e0dfaa054470cd9cf1a5e5906ea0774f3

packages/elements-dev-portal/src/hooks/useGetNodeContent.ts
lottamus: Removed them in https://github.com/stoplightio/elements/pull/1191/commits/811e57e8e70e05040bee14a5d55287c6af516a67

packages/elements-dev-portal/src/components/NodeContent/index.tsx
lottamus: @marcelltoth we can get the `http_service` from the `http_operation`'s inbound edges. Open to suggestions on how best to pass that down to the elements-core security badge

```ts
const httpServiceNode = httpOperationNode.inbound_edges.find(edge => edge.type === 'http_service');
```

philsturgeon:
As concerns are resolved (questions answered or code changes fix things) can we mark the discussions as resolved so people can read things easier.

packages/elements-dev-portal/src/components/NodeContent/index.tsx
mallachari: If we can easily get http service uri from operation node, and if we have access to that operation node (I assume we do through context) then we can do the same as we do now, just without the function that @philsturgeon mentioned, which could be removed making things easier.

philsturgeon:
I'll abstain from review on this one and leave it up to the engineers.

mpodlasin (approved):
LGTM

Issue #897, opened 2021-03-03
Branch selector in sidebar


marbemac:
Bring across the branch selector from hosted docs. Functionality is unchanged from what is in production.

<p></p>

- Position above ToC

- multiple published branches being displayed and selectable in a dropdown.

- If there are no published branches to switch between, hide the selector entirely.

- URL's should prepend /{projectSlug}:{branchSlug}/... to the current URL structure

- Any sort of error response from the Stoplight API should result in the component not showing

![](https://pb-files.s3.amazonaws.com/production/attachments/875ccecae8e2001a63e681983286419d0d9f36b6399a1b07282ef4f2e577650f/afbeelding.png)

<p></p>

PR #1193, opened 2021-05-10
feat(dev-portal): branch selector

lottamus:
closes https://github.com/stoplightio/elements/issues/897

image

Can see example of branch selector in our dev portal PR env: https://dev-portal-fuvmsk8zx-stoplight.vercel.app/docs/api


philsturgeon:
Did we intentionally remove the word Versions from this menu? It feels a little less clear what this menu is about now.

**Before**

Screen Shot 2021-05-10 at 3 51 34 PM


**After**

Screen Shot 2021-05-10 at 3 51 25 PM

packages/elements-dev-portal/.storybook/preview.jsx
```diff
@@ -44,7 +46,11 @@ const MosaicProviderDecorator = (Story) => (
   <MosaicProvider><Story/></MosaicProvider>
 );
 
-export const decorators = [ThemeProvider, ProviderDecorator, MosaicProviderDecorator, PersistenceBoundaryDecorator];
+const DevPortalProviderDecorator = (Story, context) => {
+  return (<Provider platformUrl={context.args.platformUrl}><Story/></Provider>);
```
mallachari, line 50: Dev-portal storybook runs better with that change :wink:
```suggestion
  return (<DevPortalProvider platformUrl={context.args.platformUrl}><Story/></DevPortalProvider>);
```

It might also require adding `QueryClientProvider` here in storybook

mallachari (changes requested):
Looks good at current form. Not yet in ToC but I assume it will come later. I agree with @phil that the `versions` description would be more informative.
One change request though as it breaks `dev-portal` storybook

packages/elements-dev-portal/.storybook/preview.jsx
lottamus: Fixed with https://github.com/stoplightio/elements/pull/1193/commits/6426e2ad7cb9c8c046dcdd86784ea4e14374c5f9

mallachari: Does it work for you without `QueryClientProvider`? I had to add it to make it not throw
```ts
const QueryClientProviderDecorator = (Story) => {
  return (<QueryClientProvider client={queryClient}><Story/></QueryClientProvider>);
}
```

lottamus: Oh good catch, forgot I refactored that out of the DevPortalProvider. I'll add that decorator, thanks!

mallachari (approved):
Works fine now!

philsturgeon (approved):
Based on that screenshot you've got my vote!

mpodlasin (approved):

Issue #1172, opened 2021-05-06
Dev Portal: Search


marbemac:
Implementing the frontend logic for the search functionality provided in the backend.

marbemac:
Link to subfeature: https://stoplightio.productboard.com/feature-board/planning/features/6464686

PR #1194, opened 2021-05-10
feat(dev-portal): search component

lottamus:
- Fixes https://github.com/stoplightio/elements/issues/1172
- Fixes https://github.com/stoplightio/platform-internal/issues/5654

Can see example of search working in our dev portal PR env: https://dev-portal-fuvmsk8zx-stoplight.vercel.app/docs/platform

image


philsturgeon:
@lottamus will the markdown be rendered when this is done?

philsturgeon:
An alternative would be strip it? If there’s no easy way then matching the result, rendering the full thing, then stripping HTML? 

We don’t need everything rendered just basic bold, italic, links, etc or at least remove the syntax? 

Either way, it currently just looks broken.

--
Phil Sturgeon
Product @ Stoplight.io

philsturgeon:
Ok, spoke to @lottamus lets not block on search results.

mallachari (approved):
Works fine and looks good

PR #1195, opened 2021-05-10
feat(dev-portal): stoplight project component

lottamus:
A roll-up of the rest of the PRs - helps put them all in context, but suggest not reviewing or worrying about this PR until we get through reviews on the other ones since merging those will significantly decrease the diff on this one.


examples/angular/src/app/stoplight-project/stoplight-project.component.ts
```diff
@@ -7,7 +7,7 @@ import { environment } from '../../environments/environment';
   templateUrl: './stoplight-project.component.html',
 })
 export class StoplightProjectComponent {
-  workspaceSlug = 'elements';
-  projectSlug = 'studio-demo';
+  projectId = 'cHJqOjY';
+  platformUrl = 'https://x-6195.stoplight-dev.com';
```
philsturgeon, line 11: Would this be better off called `workspaceUrl`? What you'll see when you go to this URL is a workspace.

mpodlasin:
Screenshot from 2021-05-12 16-26-45
Screenshot from 2021-05-12 16-26-38

Cloud icon differes from designes, but most importantly looks very heavy IMHO.

@philsturgeon opinion?

philsturgeon:
Yeah let's get as close to the designs as we can. White/empty cloud is good.

lottamus:
> Cloud icon differs from designs, but most importantly looks very heavy IMHO.

Can we do the light icons on the free plan? I'm pretty sure Fontawesome only offers the solid icons for free: https://fontawesome.com/icons/cloud?style=solid

mpodlasin:
Screenshot from 2021-05-13 09-55-00

I imagine that's why in the previous version the cloud icons and fonts were not black, but a bit brighter, to not give such "heavy" look.

mpodlasin:
Screenshot from 2021-05-14 12-27-33

Image resolution doesn't work (I believe it used to)

packages/elements-dev-portal/src/containers/StoplightProject.tsx
```diff
@@ -1,110 +1,116 @@
...
+          return <Redirect to={branchSlug ? `/branches/${branchSlug}/${firstNode.slug}` : `/${firstNode.slug}`} />;
+        }
+      }
+    }
+  } else if (!isFetched) {
+    elem = <Loading />;
+  } else if (!node) {
+    elem = <NotFound />;
+  } else {
+    elem = <NodeContent node={node} Link={Link} />;
```
marcelltoth, line 59: This logic is a bit convoluted. What about rewriting it like so:
```tsx
if(!nodeSlug && .... && ) {
  // ...
  return <Redicrect
}
let elem = <NotFound />;
if(node){
  elem = <NodeContent .../>
}
else if(...){
  elem = <Loading />
}

```

cypress/integration/stoplight-project.ts
```diff
@@ -11,7 +11,7 @@ describe('Stoplight component', () => {
   });
 
   it('opens via url', () => {
-    cy.visit('/stoplight-project/reference/todos/todo.v1.yaml/components/schemas/Todos');
+    cy.visit('/stoplight-project/c2NoOjEwNg-todo');
```
marcelltoth, line 14: Wasn't the new URL scheme supposed to be more readable? :trollface:

marcelltoth (commented):
Looks good to me, except for those little tweaks I've highlighted, but I've found some bugs. They may be unrelated.

We did lose the ability to provide a branch name to the component, however, and also there's no authToken support at all ATM, hopefully that's fine.

Images don't load in markdown, and also MD links open in new tabs. Even hash ones.

cypress/integration/stoplight-project.ts
philsturgeon: These hashes help point to the right node, which is a different "new URL scheme" to what we did recently making URLs _inside_ OpenAPI nicer. 

Everything on the left of the file: shorter and less redundant.
Everything on the right of the file: less squiggly.

packages/elements-dev-portal/src/containers/StoplightProject.tsx
philsturgeon: Could you push a commit for any code changes you recommend?

cypress/integration/stoplight-project.ts
marcelltoth: But then why did the `/components/schemas/Todos` disappear? @mpodlasin ?

mpodlasin: I literally have no clue. I was desparate to make the build pass in a reasonable way.

Now we can investigate :joy:

mallachari (commented):
It looks really good. I like the routing approach, table of contents having nodes type data, and nodes itself with inbound/outbound edges.
One thing that needs taking care of (probably within another pr) is broken security badges that don't link to overview, which is understandable as we changed uri structure. Thing is now we have two different structures (old in API and new in Stoplight Project) and the link is added on `HttpOperation` level. We would need to come up with a solution to support both ways (or change the API to support new uris with node slugs). So if we are ok with temporarily having unlinkable badges for Stoplight Project then this PR looks good.

cypress/integration/stoplight-project.ts
mallachari: I understand that it dissipated because it is represented by node slug. We don't need these information in uri as it is now all within node data.

marcelltoth (commented):
Also, there is zero network error handling and tests, are we OK with that?

philsturgeon:
> Also, there is zero network error handling or unit tests, are we OK with that?

Can we look at the network error handling we have, and the tests we have, and copy it across commented out? This PR should be deleting whatever its replacing, so letting those tests live on means we can add the error handling and testing back in a followup PR.

philsturgeon:
> It looks really good. I like the routing approach, table of contents having nodes type data, and nodes itself with inbound/outbound edges.
> One thing that needs taking care of (probably within another pr) is broken security badges that don't link to overview, which is understandable as we changed uri structure. Thing is now we have two different structures (old in API and new in Stoplight Project) and the link is added on `HttpOperation` level. We would need to come up with a solution to support both ways (or change the API to support new uris with node slugs). So if we are ok with temporarily having unlinkable badges for Stoplight Project then this PR looks good.

Yeah we can break them momentarily as it will only be broken in StoplightProject right? If you could make an issue for this follow up work, and add any ideas you have to fix it in that issue, we can sort it out soon. It's not the biggest problem.

philsturgeon:
> It looks really good. I like the routing approach, table of contents having nodes type data, and nodes itself with inbound/outbound edges.
> One thing that needs taking care of (probably within another pr) is broken security badges that don't link to overview, which is understandable as we changed uri structure. Thing is now we have two different structures (old in API and new in Stoplight Project) and the link is added on `HttpOperation` level. We would need to come up with a solution to support both ways (or change the API to support new uris with node slugs). So if we are ok with temporarily having unlinkable badges for Stoplight Project then this PR looks good.

Yeah we can break them momentarily as it will only be broken in StoplightProject right? If you could make an issue for this follow up work, and add any ideas you have to fix it in that issue, we can sort it out soon. It's not the biggest problem.

marcelltoth:
> Can we look at the network error handling we have, and the tests we have, and copy it across commented out? 

I'm surely not putting any commented out code anywhere, but if you can get anyone else to do so, I'll blink on it. 😄 But warning: you'll likely have a hard time fighting with eslint, too.

> This PR should be deleting whatever its replacing, so letting those tests live on means we can add the error handling and testing back in a followup PR.

Well... this is a bit self-contradicting as the old SLP *did* have proper network error handling. But I guess we'll add it later then 🤷

examples/react-cra/yarn.lock
```diff
@@ -2137,9 +2136,9 @@
     urijs "^1.19.6"
 
 "@stoplight/elements-dev-portal@file:../../packages/elements-dev-portal/dist":
-  version "1.0.0-alpha.2"
+  version "1.0.0-alpha.1"
```
mmiask, line 2139: Why are we rolling back to a previous version here?

mmiask (commented):
It looks very good. I like how we have all of the requests in one place (`useGetTableOfContents`, `useGetBranches`, `useGetNodeContent`). No major comments from me, besides the ones already posted by the rest of the team 👍 

Routing was a bit confusing to me at first, but now that I understand the approach, I like it!

mallachari:
Is there new API version of fetching mock url? Currently we're using graphql query in Docs container for that, but with new data structure it might not be correct anymore. Either way we should add it to StoplightProject the way we did in Docs container (which is not used anymore).

packages/elements-dev-portal/src/components/NodeContent/index.tsx
```diff
@@ -19,9 +20,11 @@ export const NodeContent = ({ node, Link }: NodeContentProps) => {
     <PersistenceContextProvider>
       <NodeLinkContext.Provider value={[node, Link]}>
         <MarkdownComponentsProvider value={{ link: LinkComponent }}>
-          <Box style={{ maxWidth: ['model'].includes(node.type) ? 1000 : undefined }}>
-            <Docs nodeType={node.type as NodeType} nodeData={node.data} />
-          </Box>
+          <MockingProvider mockUrl={node.links.mock_url}>
```
mallachari, line 28: This is great! No need for additional fetching. I like that new api.

philsturgeon:
Alright, this looks legit! What else do we need to do? Images are possibly fixed in https://github.com/stoplightio/platform-internal/pull/6368 and now we've got mock URLs back, anything else to do?

lottamus:
> What else do we need to do?

@philsturgeon I'm currently debugging an issue with the project node service having trouble responding to some requests such as https://stoplight.io/api/v1/projects/cHJqOjExOTY/nodes/b3A6Mzg5NDM1

philsturgeon:
@lottamus ok, so when the prod issues are resolved the tests should pass and we can merge this?

lottamus:
@philsturgeon yeah then we can release 🎉

packages/elements-dev-portal/src/handlers/getNodes.ts
```diff
@@ -23,7 +23,11 @@ export const getNodes = async ({
 
   const query = queryParams.length ? `?${queryParams.join('&')}` : '';
 
-  const response = await fetch(`${platformUrl}/api/v1/workspaces/${workspaceId}/nodes${query}`);
+  const response = await fetch(`${platformUrl}/api/v1/workspaces/${workspaceId}/nodes${query}`, {
+    headers: {
+      'Stoplight-Elements-Version': '1.0.0',
```
lottamus, line 28: Created an issue to take care of this later: https://github.com/stoplightio/elements/issues/1291

cypress/integration/stoplight-project.ts
```diff
@@ -7,18 +7,18 @@ describe('Stoplight component', () => {
     loadStoplightProjectPage();
     cy.findByText('pets').click();
     cy.findByText('Create a pet').click();
-    cy.findByRole('heading', { name: /Create a pet/i }).should('exist');
+    cy.findByRole('heading', { name: /Create a pet/i, timeout: 30000 }).should('exist');
```
lottamus, line 10: Increased the timeout until this PR is merged: https://github.com/stoplightio/platform-internal/pull/6409

philsturgeon: It got merged! :D

Issue #706, opened 2020-11-05
Logo option with x-logo support


marbemac:
The API Component should be able to show a simple logo, which either comes from `info.x-logo` or is an option passed to the component. 

<img width="325" alt="image" src="https://user-images.githubusercontent.com/67381/113151731-4a242b00-922d-11eb-95b6-7cd6679294db.png">

We can for now just put it in a centered square next to the title and see if users like that.

```
{
  "info": {
    "version": "1.0.0",
    "title": "Swagger Petstore",
    "x-logo": {
      "url": "https://redocly.github.io/redoc/petstore-logo.png",
      "backgroundColor": "#FFFFFF",
      "altText": "Petstore logo"
    }
  }
}
```

**Only in the API Component!**

If an `href` is provided, make the logo a link.

<https://github.com/Redocly/redoc/blob/master/docs/redoc-vendor-extensions.md#logoObject>

PR #1202, opened 2021-05-10
Feat: x-logo support

mmiask:
Resolves: #706

Final step after: https://github.com/stoplightio/types/pull/84 and https://github.com/stoplightio/http-spec/pull/146

**Screenshots:**

Screenshot 2021-05-10 at 15 18 18


packages/elements/src/components/Logo.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <Box
+      display="inline"
+      mr={3}
+      rounded="lg"
+      overflowY="hidden"
+      overflowX="hidden"
+      style={{ backgroundColor: logo.backgroundColor ?? 'transparent' }}
+    >
+      {logo.href ? (
+        <a href={logo.href} target="_blank" rel="noreferrer">
```
marcelltoth, line 24: `noopener`

packages/elements/src/components/Logo.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface LogoProps {
+  logo: {
+    altText: string;
+    url?: string;
+    backgroundColor?: string;
+    href?: string;
+  };
+}
+
+export const Logo: React.FC<LogoProps> = ({ logo }) => {
```
marcelltoth, line 13: Tip: you could just further destructure `logo` here.

packages/elements/src/containers/API.tsx
```diff
@@ -51,14 +51,20 @@ export interface CommonAPIProps extends RoutingProps {
    * @default "sidebar"
    */
   layout?: 'sidebar' | 'stacked';
+  logo?: {
+    url: string;
+    backgroundColor?: string;
+    altText?: string;
+    href?: string;
+  };
```
marcelltoth, line 59: Are we looking to add logo support as a prop, too? cc @philsturgeon

marcelltoth (changes requested):
2 missing pieces:

1. Poor stacked layout has no logo support. Is that intentional?
2. No tests. It'd be especially important to test an `x-logo` spec to see if transformation & display work as a whole.

packages/elements/src/containers/API.tsx
mmiask: > 2 missing pieces:
> 
> 1. Poor stacked layout has no logo support. Is that intentional?
> 2. No tests. It'd be especially important to test an `x-logo` spec to see if transformation & display work as a whole.
> 
> If these two are handled, it should be good to go, I see no problems with the code written at all.:)

@marcelltoth 
1. No design requirements for the StackedLayout, you're right. @philsturgeon - where do you think we should put it?
2. Spec transformation is already being tested in `http-spec` (if what you mean is tranlating users document to `httpService`). Unless you're talking about testing data flow, but I wouldn't personally test the outcome (display) directly, since we agreed on not testing the implementation (which I would need to do in order to check the logo). We can do visual testing of course, but not in this PR probably

mmiask:
> 2 missing pieces:
> 
> 1. Poor stacked layout has no logo support. Is that intentional?
> 2. No tests. It'd be especially important to test an `x-logo` spec to see if transformation & display work as a whole.
> 
> If these two are handled, it should be good to go, I see no problems with the code written at all.:)

@marcelltoth 
1. No design requirements for the StackedLayout, you're right. @philsturgeon - where do you think we should put it?
2. Spec transformation is already being tested in `http-spec` (if what you mean is tranlating users document to `httpService`). Unless you're talking about testing data flow, but I wouldn't personally test the outcome (display) directly, since we agreed on not testing the implementation (which I would need to do in order to check the logo). We can do visual testing of course, but not in this PR probably

packages/elements/src/containers/API.tsx
mmiask: Yes, we want users to be able to provide it both as a prop and from the API document, as described in https://github.com/stoplightio/elements/issues/706

marcelltoth:
@philsturgeon also raises a good point. His example does **not** show the current signature of the logo prop. But if we are looking to give this feature easily to web-component users we could really consider adopting that signature, because it's very finicky and ugly to pass an object to the WC component right now. This does mean a loss of certain features, of course, so it's a tradeoff. But maybe a well worth tradeoff as if they really need e.g. alt text support they can always use the `x-logo` extension.

Of course @philsturgeon should decide here.

marcelltoth:
> If web components doesnt like an object then we can just do a string, and add extra attributes later like logoAlt="" or whatever.

It technically does, but you need to pass a JSON stringified version, or set it up from javascript.

marcelltoth (approved):
https://youtu.be/K3GkSo3ujSY

PR #1203, opened 2021-05-10
Fix Demo

philsturgeon:
Some OpenAPI definitions are missing in the demo right now, let's quickly get them back.


marcelltoth (approved):

mallachari (approved):

philsturgeon:
@mallachari good spot, thank you! Fixed in https://github.com/stoplightio/elements/pull/1205

PR #1206, opened 2021-05-10
fix(schema): handle arrayish examples

P0lip:
[`examples`]( https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.9.5) in JSON Schema is an array

Needed by https://github.com/stoplightio/platform-internal/pull/6285

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ node_modules
 .pnp.js
 
 # testing
+/cypress
 /coverage
 packages/*/coverage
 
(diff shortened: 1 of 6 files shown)


marcelltoth (approved):
Looks 🆗 . Would you port this to `main` as well please? 

Especially now that @domagojk is incorporating v7 in studio (and soon we do in ninja) that would make sure we don't lose this feature. It will also simplify a lot, as in v7 you can actually make `getExamplesFromSchema` take `data: JSONSchema7` which will also simplify the implementation a lot.

P0lip:
Awesome, thank you!

PR #1209, opened 2021-05-11
Fix v6 E2E after Zoom OpenAPI file rename

philsturgeon:


marcelltoth (approved):
Thank you @philsturgeon

marcelltoth (commented):
Oh, you only changed the docs and stories. Can you make the E2E pass, too? Just do a search for `zoom.yaml`, it should be there

Issue #1140, opened 2021-04-27
Bump dependencies in examples every week


marcelltoth:
## Summary

Similar to how we maintain the lockfile in the root folder we also want to make sure we maintain the dependencies in the examples.

Obviously we can't support every framework and build tool out there. We agreed that the best compromise is to support the latest stable version of the frameworks our examples use: _Create React App_, _Gatsby_, _Angular_ (and maybe later _Next.js_ which is being used by Product's POC site). The aim of this issue is to give us confidence that a fresh Elements install will work with the latest version of every framework above. This of course means that we not only bump the lockfile here, but the constraints, too (if needed). Actually, as these are applications, not libraries, [we should probably pin it](https://docs.renovatebot.com/dependency-pinning/#so-whats-best) anyway. 

## Technology

The technology choice is of course up to the implementor, here's some suggestions:
- The simplest way you can try locally is to run `yarn upgrade --latest` in the example folder. One option would be to schedule this using CircleCI, just like the lockfile maintenance.
- The other possible way is to set up GitHub's dependabot. This would be different because it files one PR for one dependency, which we might consider an advantage. It also has some suppress features, etc. which may turn out useful if e.g. latest `react-scripts` would not support latest `typescript`, if there is a regression in a dependency or similar.
- Renovate is a tool still, of course.

## Additional info

We can even consider putting these PRs on Automerge, assuming we trust our E2E suite, which I personally do.

PR #1213, opened 2021-05-11
chore: weekly maintenance v2

marcelltoth:
Implements #1140 

Also migrates to the new Github dependabot. We don't need the old one at all anymore as Github has Security advisory and resolution [built in](https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies).

If this will look interesting we can investingate using it for (some of the?) weekly maintenance we do on the main yarn.lock as well.


.dependabot/config.yml
```diff
@@ -1,33 +0,0 @@
...
-    version_requirement_updates: 'increase_versions'
-
-    commit_message:
-      prefix: 'fix'
-      prefix_development: 'fix'
-      include_scope: true
-
-    allowed_updates:
-      - match:
-          update_type: 'security'
```
mmiask, line 18: Why are we getting rid of the security bumps? I though we decided on the custom CI job because we wanted to have both live security updates and weekly minor updates.

mmiask (commented):
New config looks good 👍  Please address my concerns and this should be good to go!

.dependabot/config.yml
marcelltoth: Yeah, I know. That used to be the case, but I recently learned that GitHub now handles security independently from updates. See article in the description, that means we don't need this here. I still feel like our custom job is probably better than Dependabot for the other use case, but if this works, we can maybe try what would happen if we used it for the lockfile maintenance, too. This is somewhat of an experiment in that regard, too.

mmiask (approved):
Good to go whenever checks pass! 🎉

marcelltoth:
WTF happened with tests? I haven't even touched code 😅

Issue #1201, opened 2021-05-10
HttpService doesn't work in isolation (and tests don't show it)


mpodlasin:
![Screenshot from 2021-05-10 14-59-22](https://user-images.githubusercontent.com/7916523/117662995-78881500-b1a0-11eb-8a72-87a8cdc36737.png)

We seem to have 2 separate problems.

In our storybook it complains that `Uncaught TypeError: useContext(...) is undefined`

On the other hand in studio PR (https://github.com/stoplightio/platform-internal/pull/6126) there seems to be a router issue (they are not wrapping ParsedDocs in a Router and they shouldn't).

This is highlighted again in our test suite, where we have to wrap `HttpOperation` in `Router` in order for tests to pass.

![Screenshot from 2021-05-10 15-02-51](https://user-images.githubusercontent.com/7916523/117664294-eaad2980-b1a1-11eb-9425-74b1bda4647f.png)

This is however a wrong strategy, because this makes `Docs` and `ParsedDocs` unrenderable outside of Router. Additionally when this component is used in some other app then `elements`, it might pick-up errorenous routes from "mother application".

PR #1218, opened 2021-05-11
chore: refactor docs components to have less variants

mpodlasin:
Related to #1201 

I've decided to do a little refactor before actual fix, because this will make my life easier.

Plus I believe it to be a right direction in general.

Reduces nr of "docs-like` components from 4 to 3.

Let me know what you think! :)


yarn.lock
```diff
@@ -3894,13 +3894,71 @@
   dependencies:
     "@sinonjs/commons" "^1.7.0"
 
+"@stoplight/elements@7.0.0-beta.1":
```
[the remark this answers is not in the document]
mpodlasin: Gone now :tada:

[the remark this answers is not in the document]
marcelltoth: I fixed that in #1217 

In general I'd say we should worry less about the lockfile, especially from a Product standpoint. It does not *declare* any dependencies, it's not shipped to customers, it's an implementation detail.

philsturgeon (approved):

packages/elements-core/src/components/Docs/Docs.tsx
```diff
@@ -46,18 +25,40 @@ export interface IDocsComponentProps<T = unknown> {
   uri?: string;
 }
 
-export const Docs = React.memo<IDocsProps>(({ nodeType, nodeData, className, headless, uri }) => {
-  const parsedNode = useParsedData(nodeType, nodeData);
+export interface DocsProps extends BaseDocsProps {
```
marcelltoth, line 28: Thank you

packages/elements-core/src/components/Docs/Docs.tsx
```diff
@@ -46,18 +25,40 @@ export interface IDocsComponentProps<T = unknown> {
   uri?: string;
```
marcelltoth, line 25: While we're here, how 'bout renaming this? `uri` means *everything*. It's great that we have JSDoc, but what if we called it like `documentUri` or something, and then the JSDoc could describe what it's used for in the first place.

packages/elements-core/src/hooks/useParsedData.ts
```diff
@@ -3,11 +3,20 @@ import { NodeType } from '@stoplight/types';
...
 
-import { JSONSchema, ParsedNode } from '../types';
+import { JSONSchema, ParsedNode, UnparsedNode } from '../types';
 import { isHttpOperation, isHttpService, isJSONSchema, isSMDASTRoot } from '../utils/guards';
 
-export function useParsedData(nodeType: string, data: unknown): ParsedNode | undefined {
-  return React.useMemo(() => parserMap[nodeType]?.(data), [nodeType, data]);
+export function useParsedData(
+  unparsedNode: UnparsedNode | undefined,
+  fallbackNode?: ParsedNode,
```
marcelltoth, line 11: This "fallback" that isn't really a fallback (as it takes priority) feels off here. More thoughts above.

packages/elements-core/src/components/Docs/Docs.tsx
```diff
@@ -46,18 +25,40 @@ export interface IDocsComponentProps<T = unknown> {
...
 
   if (!parsedNode) {
     // TODO: maybe report failure
     return null;
   }
+  const parsedDocs = <ParsedDocs node={parsedNode} {...commonProps} />;
+
+  if (useNodeForRefResolving) {
+    return <InlineRefResolverProvider document={parsedNode.data}>{parsedDocs}</InlineRefResolverProvider>;
+  }
```
marcelltoth, line 52: Any chance we can just take the inline ref resolver context and making the condition `if(!context)`? Assuming if someone has a context around they want that, otherwise we fall back to this? May be a stupid idea, just throwing it out here.

marcelltoth (commented):
There are things I love here, and things I'm not convinced by.

All the syntactical cleanup, I love. No more `interface IHateWhen extends IPrefixStuffForNoReason` 💌 . `BaseDocsProps` cleanup? Love it.

I'm not convinced by combining the two docs entry points however. 

At minimum this "feed either `node` **or** `unparsedNode` but not both" thing needs explanation. (JSDoc potentially) We could also consider enforcing it via TS. `({useNodeForRefResolving} & ({node} | {uparsedNode})` 

But as `Docs` is really just a wrapper for `useParsedData` then `ParsedDocs`, I don't really see why we should mush them into the same component. If most use cases will just feed parsed data then there is a case for simply removing `Docs` as is, renaming `ParsedDocs -> Docs`, and making the consumer call `useParsedData` themselves.

The typings are also a tiiiny bit worrying, as `ParsedNode` is a subset of `UnparsedNode`, which... I can't really explain but it feels weird.

So... IDK. I'm open to more discussion on this, and I absolutely won't block it if there's consensus that this is better combined, but at this point, I'm not convinced. The other pieces are great though, we could even do a shortcut PR shipping just those real quick while we discuss the rest of this.

packages/elements-core/src/hooks/useParsedData.ts
mpodlasin: This really comes from the fact that hooks suck.

In reagular code I would just make a readable condition and this would never ever be a problem.

Hooks have no proper support for conditionals so you end up with code smell like that.

packages/elements-core/src/components/Docs/Docs.tsx
mpodlasin: I was thinking about this, but it seemed to high-concept to pass a review :joy: 

Maybe do it in a separate PR? We still have an issue to fix problems with providers.

mpodlasin:
@marcelltoth 

I guess it's just a matter of style/perspective and there will be no clear answer here.

I just hate, hate, hate exporting 2 things (component + hook) and then asking user to combine them themselves.

This is the whole point of creating libraries - if we see an useage over and over again (which we do - useParsedData + ParsedDocs appears multiple times both in our codebase and in platform), we should abstract it for the user.

I disagree that having two options for props (parsed and unparsed) is confusing *that much*.

It's like saying that `apiDescriptionUrl` and `apiDescriptionDocument` in API is confusing and we should have 2 separate components.

mallachari:
From my point of view @marcelltoth approach seem more natural even though it may include more exports. Though I can understand need of combining all in one and it makes sense too and the example with API is more or less correct. So as it's indeed matter of who prefers what I'm slightly leaning towards options with some separation. The idea with 3 components that @mpodlasin suggested is quite good solution I think

philsturgeon:
> I just hate, hate, hate exporting 2 things (component + hook) and then asking user to combine them themselves.

I believe this a requirement for frameworks like Next.js, they have to fetch all the data then update the component seperately. Or is that something else?

mpodlasin:
yeah, as agreed on slack I'll write an "inbetween version", which I believe will satisfy everyoone.

packages/elements-core/src/components/Docs/Docs.tsx
mpodlasin: You will have to give me some name though.

Because `documentUri` doesn't tell me much either... :sweat_smile:

mallachari (approved):
That in between version looks good :+1:

Issue #1216, opened 2021-05-11
Bring back proper names for styles in elements and elements-dev-portal packages


mpodlasin:
Currently both `elements` and `elements-dev-portal` packages export styles under the name `elements-core.min.css`. 

We don't want that. `core` is an implementation detail and users should not be aware of that name.

- Change our CSS URLs to:

    - https://unpkg.com/@stoplight/elements@beta/styles.min.css
    - https://unpkg.com/@stoplight/elements-dev-portal@beta/styles.min.css

- Change react to match.

![Screenshot from 2021-05-11 11-08-29.png](https://images.zenhubusercontent.com/5f857aaa311cec7e20b4f258/d2f30d50-4183-4908-999d-8e0dc3bacc64)

PR #1253, opened 2021-05-12
fix: building styles

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1216

As we decided to make output filenames `styles.min.css` and `styles.scss` the final path would be `node_modules/@stoplight/elements/styles/styles.min.css`. 
We could avoid `styles` directory and have it `node_modules/@stoplight/elements/styles.min.css` but then we would need to extract those files from other styles to the root folder (having two locations for styles). Do we want to do that?


docs/guides/react.md
```diff
@@ -24,14 +24,14 @@ In `App.js` import the API component and CSS file from the Elements library.
 <!-- title: App.js -->
 ```jsx
 import { API } from '@stoplight/elements';
-import '@stoplight/elements/styles/elements.min.css';
+import '@stoplight/elements/styles/styles.min.css';
```
mpodlasin, line 27: good job on changing docs!

examples/angular/angular.json
```diff
@@ -25,7 +25,7 @@
             ],
             "styles": [
               "src/styles.css",
-              "node_modules/@stoplight/elements-dev-portal/styles/elements-core.min.css"
+              "node_modules/@stoplight/elements/styles/styles.min.css"
```
mpodlasin, line 28: Not sure that's the proper way. If people use dev-portal, they should install dev-portal styles, since they are more general.

mpodlasin (commented):
Re: directory of styles.

If moving styles to directory up is possible, I would to it.

If however it causes problems, or some unforeseen issues, then we can drop it of course.

examples/angular/angular.json
mallachari: What do you mean by more general? Should they differ? Currently they are build from `elements-core` and copied to `elements` and `elements-dev-portal`

marcelltoth: I'm with @mpodlasin on this.

> Currently they are build from elements-core and copied to elements and elements-dev-portal

That's an implementation detail. It may not be true soon if we decide to add specific styles to dev-portal. 

Also people that only install dev portal shouldn't need to be aware that they have a `node_modules/@stoplight/elements` directory. In some weird cases they really may not have one, it's not guaranteed unless you depend on `elements` directly. (For example it's perfectly legal for a package manager to install it as `node_modules/@stoplight/elements-dev-portal/node_modules/@stoplight/elements`)

marcelltoth:
I'd also vote for moving it up, just looks cleaner.

examples/angular/angular.json
mallachari: But then shouldn't we include both styles, for elements and dev-portal? What is making dev-portal more generic than elements and not the other way around?

marcelltoth: In case the angular example is using both (which I think it is) then yes, we should include both CSS files

mpodlasin: Yeah, you are actually correct.

It should be both. 

I just don't want to people impression that when they use `elements-dev-portal`, they should use `elements` styles, that's all.

mallachari:
So here's the update. I've ditched copying everything and made styles compile to root folder. It works fine with all examples. Demo doesn't use build package so isn't a problem.

marcelltoth (approved):

philsturgeon (approved):

mmiask (approved):

mpodlasin (approved):

Issue #708, opened 2020-11-05
Show/hide "TryIt"


marbemac:
As an Elements user, I can pass a prop into the API component, and StoplightProject component, so I can hide the "TryIt" component from all HTTP operations.

<p></p>

- Add a property `hideTryIt` to the API and StoplightProject components

- When enabled, we should hide the "TryIt" component

mallachari:
@philsturgeon does hiding TryIt mean hiding whole right panel with request samples and response examples or  should we keep them there?

philsturgeon:
Keep them. Just try it is going away. Some people won’t have a working environment to point to  (CORS drama) or won’t want to use it for other various reasons, but sample code is still good. 

--
Phil Sturgeon
Product @ Stoplight.io

mallachari:
Alright, thanks! Though now I am wondering if it fits to our strategy for using Docs and TryIt component in platform (the issue that is blocked by this one). Our current Docs show operation and TryIt with samples combined. If we hide just TryIt and leave the rest they will be shipped this way to platform which would make adding TryIt painful. Perhaps we could have it this way and add another option of hiding whole right panel so we can easily separate them when needed?

PR #1265, opened 2021-05-14
feat: hide tryit

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/708

We need to be able to hide just TryIt and whole right column. 

Initially I wanted to add support for hiding each of the right column components, but Examples depend on more complex inner logic when it comes to displaying it so I went for whole column approach. 
Also Request Samples component is currently dependent on TryIt which feeds its props.

Naming is still something I am not sure about. `docsOnly` seemed better than `hideRightColumn` but not ideal. 
I also considered one `visibility` prop with value of `docsOnly | hideTryIt` or `full | docsOnly | hideTryIt`. 
Let me know what you think.

Adding provider isn't necessary but avoids passing props through lot of components. We're starting to get plenty of those providers so we probably need to clean and combine them at some point.


philsturgeon:
I am not a fan of this option. It feels like we structured out components too rigidly, meaning that we now have to turn bits off instead of using the bits we want in the way we want them.

The end goal of elements is to let people pick and chose bits, maybe mixing their own components in between our own, and not always to be just `API` and `StoplightProject`. If we've already hit a roadblock with the structure of our components, to a point where we cant simply hide one piece of it, then we might need to revisit our approach.

If we need this quickly, we can implement docsOnly as a hidden property, undocumented, unsupported, and delete it once we have resolved our composition issue, but I wouldn't want it around for long.

mallachari:
What if we extracted right column content from HttpOperation? So we could define its content on API level and pass down to HttpOperation? This way we could either include it or not and hiding/customizing elements would be simplified. Not sure how difficult it would be though but seems feasible. What do you think @mmiask @mpodlasin @marcelltoth?

marcelltoth:
> 
> 
> What if we extracted right column content from HttpOperation? So we could define its content on API level and pass down to HttpOperation? This way we could either include it or not and hiding/customizing elements would be simplified. Not sure how difficult it would be though but seems feasible. What do you think @mmiask @mpodlasin @marcelltoth?

That's exactly what I was joking about yesterday. IIRC that's how it used to be, but then we put it into Docs.:) But see, that's how things are in SE, I'm supporting your idea if it works well in practice, show me a PR:)

mallachari:
Alright, I'll go that path then. Let's try to make it right at once and not fix later. As I think now it shouldn't be too difficult. Quite exceeding that 1 pointer though, but well ;)

mpodlasin:
BTW I don't believe hiding with boolean has to affect users.

It's Docs component that can have a boolean flag, not API component as a whole.

API can have only `hideTryIt` and we are done. 

Just to keep this in mind as an alternative solution.

Putting TryIt to Docs made thinks much cleaner, but if you are able to come up with a clean way to have them outside - I will :100: approve.

philsturgeon:
Ok, if it’s just for docs component then let’s do it for now but we need to make an issue describing the other approach suggested here. 

We’ve already got some customers trying to figure out how to do really advanced stuff and we’ll need to do it fairly soon. 

--
Phil Sturgeon
Product @ Stoplight.io

mallachari:
I've updated it following the suggestion from @mpodlasin . Also removed `showMocking` from the Provider container and moved it to `TryItProvider` along with `hideTryIt`. 
Generally when we merge new StoplightProject PR we might be able to remove that old Provider entirely or replace it with something more up to date as the current form is far from perfect.

packages/elements-core/src/components/Docs/HttpService/__tests__/index.spec.tsx
```diff
@@ -59,9 +59,9 @@ describe('HttpService', () => {
 
   it('displays mock server url when embedded in Stoplight Project', async () => {
     render(
-      <Provider host="https://stoplight.io" showMocking project="studio-demo" workspace="elements-examples">
+      <TryItProvider showMocking>
```
mpodlasin, line 62: Yeah, I highly, highly dislike that. :sweat_smile: 

We are using a `TryItProvider` on something that doesn't use `TryIt`, has nothing to do with `TryIt`, etc. :cry:

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.spec.tsx
```diff
@@ -531,6 +532,26 @@ describe('HttpOperation', () => {
       expect(screen.queryByText('some_property')).not.toBeInTheDocument();
     });
   });
+
+  describe('Visibility', () => {
+    it('should hide TryIt', async () => {
+      render(
+        <TryItProvider hideTryIt>
```
mpodlasin, line 539: I don't see any reason why we shouldn't simply pass `hideTryIt` directly to HttpOperation, just like `docsOnly`.

packages/elements-core/src/components/Docs/Docs.tsx
```diff
@@ -24,6 +24,11 @@ interface BaseDocsProps {
    * URI of the document
    */
   uri?: string;
+
+  /**
+   * Shows only operation document without right column
+   */
+  docsOnly?: boolean;
```
mpodlasin, line 31: After careful consideration I would go for having 2 options here: `hideTryIt` and `hideTryItPanel`.

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.spec.tsx
mallachari: We could do that, though we need that property on API level, which would mean that we need to add it to `APIWithSidebarLayout` too and pass it to Docs there. My argument was that layout component shouldn't need to know about this property. Do you think it's fine to do that? Unless you mean to still use context but read it on Docs level, not HttpOperation?

packages/elements-core/src/components/Docs/HttpService/__tests__/index.spec.tsx
mallachari: True, initially I assumed it's HttpOperation there. Do you think we should name it differently or not use provider at all?
I assume we would need to get rid of that old Provider anyway and replace it with data we do need. Though it can't be done now so I added that TryItProvider assuming that `showMocking` refers only to TryIt - but it does not

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.spec.tsx
mpodlasin: Quite honestly, I don't see a problem.

API has `hideTryIt` property, so for me it's 100% okay for `ApiWithSidebarLayout` to have that property as well. :shrug:

mpodlasin: I know that `context` makes passing that data easier, but I believe we overrely on Providers a bit. Rendering a component in isolation requires us to use 4-5 providers, which is a bit crazy.

Thus I believe we should avoid Providers where it's possible, and here it is absolutely possible in my opinion.

mallachari: Well, if you're fine with that then I guess we can have it this way. My idea was that we will use it later in new Provider, which we might not be able to avoid, but we can think of it when we decide to refactor it. For now I'll keep it simple and just pass the property then.

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.tsx
```diff
@@ -16,88 +16,93 @@ import { Responses } from './Responses';
...
-  const isDeprecated = !!data.deprecated;
-
-  const [responseMediaType, setResponseMediaType] = React.useState('');
-  const [responseStatusCode, setResponseStatusCode] = React.useState('');
-  const [requestBodyIndex, setTextRequestBodyIndex] = React.useState(0);
-
-  const httpServiceUri = uri && getServiceUriFromOperation(uri);
-
-  const securitySchemes = flatten(data.security);
+const HttpOperationComponent = React.memo<HttpOperationProps>(
```
mallachari, line 19: This file is mostly the same. The only change is adding two props and using them in TryIt section

packages/elements-core/src/components/Docs/Docs.tsx
```diff
@@ -24,6 +24,16 @@ interface BaseDocsProps {
...
+
+  /**
+   * Allows to hide TryIt component
+   */
+  hideTryIt?: boolean;
+
+  /**
+   * Shows only operation document without right column
+   */
+  hideTryItPanel?: boolean;
```
mmiask, line 36: Is this just for a direct usage by users? As of right now they won't be able to pass `hideTryItPanel` to the API component, just when they use `Docs`. What is the purpose for that?

mmiask (approved):
Looks good! I checked the Storybook both for the `API` and `Docs` and it works great 👍 

Thanks for adding tests! 🙏

packages/elements-core/src/components/Docs/Docs.tsx
mallachari: Yes, that's on purpose. We don't want to expose that property, rather keep it for internal usage

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.spec.tsx
mallachari: I've updated it so it doesn't use Provider. 
I've also added hiding TryIt for stacked layout, but then I had to use local context. Adding that prop to Group and Item was to painful for me :wink:. In this layout it didn't make much sense hiding just TryIt component. We dont have there examples (no reason for that) and samples are bound to TryIt data so I just hid whole tab for that case.

mpodlasin (approved):
Love it, thank you :pray:

PR #1272, opened 2021-05-14
fix: allow to render docs without react-router

### LINKED ISSUE #1201: HttpService doesn't work in isolation (and tests don't show it)

mpodlasin:
Closes #1201


packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
```diff
@@ -64,7 +63,7 @@ const HttpServiceComponent = React.memo<HttpServiceProps>(({ className, data, he
       ) : (
         <Box mb={10}>
           {description}
-          <PoweredByLink source={data.name ?? 'no-title'} pathname={pathname} packageType="elements" headless />
+          {pathname && <PoweredByLink source={data.name ?? 'no-title'} pathname={pathname} packageType="elements" headless />}
```
philsturgeon, line 66: Does this mean the powered by link is sometimes hidden?

mpodlasin: Yes, in `platform-internal` usages.

mpodlasin: ...and when we forget to provide proper props, so I will actually add e2e test that it is actually visible in API.

I don't remember if it was supposed to be visible in StoplightProject? @philsturgeon

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
```diff
@@ -64,7 +63,9 @@ const HttpServiceComponent = React.memo<HttpServiceProps>(({ className, data, he
       ) : (
         <Box mb={10}>
           {description}
-          <PoweredByLink source={data.name ?? 'no-title'} pathname={pathname} packageType="elements" headless />
+          {pathname && (
+            <PoweredByLink source={data.name ?? 'no-title'} pathname={pathname} packageType="elements" headless />
+          )}
```
marcelltoth, line 68: I also find it weird that this depends on the pathname being present :\

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
```diff
@@ -20,8 +19,8 @@ const enhanceVersionString = (version: string): string => {
 
 export type HttpServiceProps = DocsComponentProps<Partial<IHttpService>>;
 
-const HttpServiceComponent = React.memo<HttpServiceProps>(({ className, data, headless }) => {
-  const { search, pathname } = useLocation();
+const HttpServiceComponent = React.memo<HttpServiceProps>(({ className, data, headless, location = {} }) => {
+  const { search, pathname } = location;
```
marcelltoth, line 23: Can we maaybe pass in what we actually need, and not the entire `Location` object? Would be more versatile that way. But of course you can likely prove that this is what we need here.

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
mpodlasin: If `pathname` is only a "nice to have" then I can make it optional in PoweredByLink. :shrug: 

But in practice it will work the same either way, so I chose simpler implementation.

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
mpodlasin: I somewhat disagree.

If we decide that Docs components can have logic depending on location, I don't see a reason to not provide the whole location object in one go. 

If it turns out that other parameters are not necessary, no harm will be done.

If in the future more stuff will be necessary, it will be already there.

But I don't feel strongly about that either way :shrug:

mpodlasin:
@philsturgeon @marcelltoth 

I would just add that in case we forget to pass location data, it will be easier to see that we screwed up if the component is missing altogether, rather that when it has a wrong url.

In fact even writing e2e test will be *a bit* easier.

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
philsturgeon: Yeah we want powered by to show up on stoplightproject too.

To make it easier to read, there could be a hidePoweredBy variable or something like that, if its purely about platform internal. We might have an option to hide powered by if theyre on a certain plan too, so it would generally be a handy feature to have so long as its not a simple boolean in the control of the user.

mpodlasin:
I am sorry, I wasn't thinking straight on friday afternoon.

This is PoweredByStoplight displayed in HttpService component in *stack layout*.

There techinically is already a prop to control if it should be displayed, and it is `headless` prop.

Check for `pathname` is just a typeguard and it will always be `true` in our usecases It's just me playing with TS nice.

And even if I add `showPoweredBy` or something, it will be still `showPoweredBy && pathname` in the condition itself, so it will not help to make things more readable in any ways. :shrug: 

I honestly don't know what I should do here.

marcelltoth (approved):

philsturgeon:
@mpodlasin if it’s just type safe stuff not hidden functionality then it’s fine as is.

philsturgeon (approved):

PR #1284, opened 2021-05-17
feat(elements-core): index export

marcelltoth:
Resolves #6256

I tried to change as little as possible, and just mechanically replace deep imports with root exports. You may see that there's probably way too many things exported, well, that's how many we use.

One notable exception is that I tried to do away with the `JSONSchema` type in favor of `JSONSchema7` now that we don't need to handle the others. That one requires a JSV PR, too, which is soon to follow.


packages/elements-core/src/index.ts
```diff
@@ -0,0 +1,32 @@
+export { Docs, ParsedDocs } from './components/Docs';
```
mpodlasin, line 1: perhaps it's just me overcomplicating things as always, but I would either:

- separate with comments which stuff is used in "3rd party" apps (platform, studio etc.)
- or even place them in separate file (`@stoplight/elements/protected` or something)

Because currently we don't know which changes affect only "our" repos and which affect 3rd party users.

Might make versioning easier? 🤔

mallachari: Yeah, some kind of annotation might be helpful

marcelltoth: We agreed to tackle this separately.

philsturgeon (approved):

mpodlasin (approved):
Let's roll!

Issue #1285, opened 2021-05-17
StoplightProject workspace must be on Starter plan


philsturgeon:
As the StoplightProject component can be used for pretty high level and customization, the sort your normally would have to pay for on most documentation tools and on hosted Stoplight, a business decision was made to make StoplightProject require a Starter plan.

The elements service will respond with an error:

```
{
  "code": 402,
  "type": "PAYMENT_REQUIRED",
  "message": "This feature is not enabled for your workspace billing plan, or your workspace billing is out of standing."
}
```

The API will look at the origin of the request to see if its localhost, and if its not it'll give this error. 

This will require a nice error screen explaining what is happening.

> Please upgrade your Stoplight Workspace to the Starter to use Elements Dev Portal in production.

PR #1293, opened 2021-05-19
Feat/display upgrade workspace screen

mmiask:
Resolves: #1285 

The screen is clickable and points users to [Stoplight Pricing Page](https://stoplight.io/pricing/)

**Screenshots**
Screenshot 2021-05-19 at 13 46 23

**If you want to check it yourself:**

1. Change `getNodeContent.tsx`, `getBranches.tsx` and `getTableOfContents.tsx`
```ts
...const response = await fetch(`${platformUrl}/api/v1/projects/...

...const response = await fetch(`${platformUrl}/v1/projects/...
```
2. Run `prism mock services/elements/openapi.v1.json` from `platform-internal`
3. Run `yarn elements-dev-portal storybook` from Elements
4. Use url from Prism as `platformUrl`
5. Open an operation/article/model

### DIFF
--- a/packages/elements-dev-portal/src/components/UpgradeToStarter/index.tsx
+++ b/packages/elements-dev-portal/src/components/UpgradeToStarter/index.tsx
@@ -1,0 +1,21 @@
+import { faExclamationTriangle } from '@fortawesome/free-solid-svg-icons';
+import { Box, Flex, Icon } from '@stoplight/mosaic';
+import React from 'react';
+
+export const UpgradeToStarter = () => (
+  <Flex
+    as="a"
+    href="https://stoplight.io/pricing/"
+    justify="center"
+    alignItems="center"
+    w="full"
+    minH="screen"
+    color="muted"
+    flexDirection="col"
+  >
+    <Icon icon={faExclamationTriangle} size="4x" />
+    <Box pt={3}>
+      Please upgrade your Stoplight Workspace to the Starter Plan to use Elements Dev Portal in production.
+    </Box>
+  </Flex>
+);
--- a/packages/elements-dev-portal/src/components/UpgradeToStarter/index.tsx
+++ b/packages/elements-dev-portal/src/components/UpgradeToStarter/index.tsx
@@ -6,6 +6,8 @@ export const UpgradeToStarter = () => (
   <Flex
     as="a"
     href="https://stoplight.io/pricing/"
+    target="_blank"
+    rel="noreferrer noopener"
     justify="center"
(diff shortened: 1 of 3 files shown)


packages/elements-dev-portal/src/components/UpgradeToStarter/index.tsx
```diff
@@ -0,0 +1,21 @@
+import { faExclamationTriangle } from '@fortawesome/free-solid-svg-icons';
+import { Box, Flex, Icon } from '@stoplight/mosaic';
+import React from 'react';
+
+export const UpgradeToStarter = () => (
+  <Flex
+    as="a"
+    href="https://stoplight.io/pricing/"
```
marcelltoth, line 8: Would you maybe consider making this a `target="_blank" rel="noreferrer noopener"`? @philsturgeon ?

packages/elements-dev-portal/src/containers/StoplightProject.tsx
```diff
@@ -56,6 +62,10 @@ const StoplightProjectImpl: React.FC<StoplightProjectProps> = ({ projectId, hide
   let elem: JSX.Element;
   if (isLoadingNode || !isTocFetched) {
     elem = <Loading />;
+  } else if (isError) {
+    if ((nodeError as any).status === 402) {
+      elem = <UpgradeToStarter />;
+    }
```
marcelltoth: Why do we need to cast it as `any`? Is it an unknown? Can we maybe cast it to the runtime type? 

Also I know this is more of a question of API design, but is this really a good use case for 402? MDN lists it as "Experimental technology" and "nonstandard", and also it's a bit misleading: It's not the client (that's fetching the data) that needs to make a payment but the host itself. @philsturgeon I'd love to hear your thoughts on this and on 402 in general.

packages/elements-dev-portal/src/handlers/getNodeContent.ts
```diff
@@ -21,7 +21,10 @@ export const getNodeContent = async ({
   const data = await response.json();
 
   if (!response.ok) {
-    throw new Error(data);
+    throw {
+      status: response.status,
+      data: data,
+    };
```
marcelltoth, line 27: Oh I see where Status above comes from. IMO we should never throw anonymous types, and more generally anything that is not `instanceof Error`. 

Could you make a [custom error class](https://javascript.info/custom-errors) instead?

marcelltoth (approved):
Looks great now, thank you!

PR #1295, opened 2021-05-19
feat(demo): add CORS proxy functionality

### LINKED ISSUE #1039: Set up a CORS proxy with Enablers
## User story.

As a followup to #1037 we would like to add our own CORS proxy to SaaS (or to a separate domain) that provides CORS proxying for 1. our Elements demo 2. our SaaS Elements `TryIt`.

Update the demo app to use it to load the descriptions.

## Is your feature request related to a problem?

See #1037

We would like to have something like https://github.com/Redocly/cors-anywhere locked down to only allow our Origin URLs.

marcelltoth:
Closes #1039 

Unfortunately the enablers decided to enforce the origin on the proxy so it doesn't work from localhost, you have to trust your code-reading abilities with this.

image

### DIFF
--- a/demo/src/components/ElementsAPI.tsx
+++ b/demo/src/components/ElementsAPI.tsx
@@ -7,11 +7,16 @@ import React, { useContext } from 'react';
 import { GlobalContext } from '../context';
 
 export const ElementsAPI: React.FC = () => {
-  const state = useContext(GlobalContext);
+  const { apiDescriptionUrl } = useContext(GlobalContext);
+
+  const specUrlWithProxy =
+    apiDescriptionUrl && window.location.origin === 'https://elements-demo.stoplight.io'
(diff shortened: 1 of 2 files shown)


mpodlasin:
There exist browser plugins that - if I am not mistaken - allow to test this on localhost, like https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf

philsturgeon:
This is the URL I'm trying and I still get CORS errors: 

https://deploy-preview-1295--stoplight-elements-demo.netlify.app/?spec=https://developer.ohpen.com/api/1/1.1.2.json

philsturgeon (approved):
Ah yeah ok its not working in demo because enablers have not enabled the subdomains. Let's just go in blind and hope for the best.

marcelltoth:
It works!

PR #1309, opened 2021-05-21
feat: add scss to core build

mallachari:
Add scss to elements-core build. Elements and elements-dev-portal are left with just css

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.0.0-beta.5",
+  "version": "7.0.0-beta.6",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**"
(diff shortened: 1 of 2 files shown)


mpodlasin (approved):
Tested it and it works well.

philsturgeon:
This looks like it’s just adding something, but don’t we need to also remove something from somewhere, or is this the first time we’ve used SCSS?

mpodlasin:
> This looks like it’s just adding something, but don’t we need to also remove something from somewhere, or is this the first time we’ve used SCSS?

This is - most likely - temporary solution. We can't use our css in platform, because this would double mosaic and uikit styles. Once uikit is gone, this sass will be come css (and perhaps it will change into a single stylesheet.

In other words `elements.min.css = elements styles + mosaic`. But we need "elements only" styles as well.

mallachari:
We've removed scss before from built packages assuming that styles.min.css is enough, but they will be useful in platform so I added them back just for elements-core.

Issue #1325, opened 2021-05-24
Add internal badge to elements


mnaumanali94:
If an endpoint/model contains x-internal:true, add an internal badge on it. 

![Screenshot 2021-05-24 at 10.12.50 PM.png](https://images.zenhubusercontent.com/5ea9a7f489044245dbd23d46/56178281-a887-4779-9338-ba0ca6e2e10c)

paulatulis:
@mnaumanali94 should the internal badge include a tooltip?

mnaumanali94:
This endpoint/model is marked as internal and won't be visible in public docs.

PR #1326, opened 2021-05-24
feat: internal badge

paulatulis:
Addresses #1325

Adds internal badge for endpoints marked as `x-internal`
Screen Shot 2021-05-25 at 3 53 56 PM

Adds internal badge for models marked as `x-internal`
Screen Shot 2021-05-25 at 3 51 50 PM

I also updated the layout so that the icon and text are aligned...I don't think this has a negative ripple effect anywhere, but I am not super familiar with Elements, so please let me know if I should undo this.
Screen Shot 2021-05-24 at 2 18 27 PM
Screen Shot 2021-05-24 at 2 33 19 PM

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpOperation/Badges.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/Badges.tsx
@@ -1,4 +1,4 @@
-import { faExclamationCircle, faLock, IconDefinition } from '@fortawesome/free-solid-svg-icons';
+import { faExclamationCircle, faEye, faLock, IconDefinition } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { HttpSecurityScheme } from '@stoplight/types';
 import { Position, Tag, Tooltip } from '@stoplight/ui-kit';
@@ -37,8 +37,11 @@ export const DeprecatedBadge: React.FC = () => (
 );
 
 export const InternalBadge: React.FC = () => (
-  <Tooltip position={Position.BOTTOM_LEFT} content="This operation has been marked as internal.">
-    <Badge icon={faLock} className="sl-bg-danger sl-ml-0 ">
+  <Tooltip
+    position={Position.BOTTOM_LEFT}
+    content="This endpoint/model is marked as internal and won't be visible in public docs."
+  >
+    <Badge icon={faEye} className="sl-bg-danger sl-ml-0 ">
       Internal
     </Badge>
   </Tooltip>
--- a/packages/elements-core/src/components/Docs/HttpOperation/Badges.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/Badges.tsx
@@ -36,10 +36,10 @@ export const DeprecatedBadge: React.FC = () => (
   </Tooltip>
 );
 
-export const InternalBadge: React.FC = () => (
+export const InternalBadge: React.FC<{ isHttpService?: boolean }> = ({ isHttpService }) => (
   <Tooltip
     position={Position.BOTTOM_LEFT}
-    content="This endpoint/model is marked as internal and won't be visible in public docs."
+    content={`This ${isHttpService ? 'operation' : 'model'} is marked as internal and won't be visible in public docs.`}
   >
     <Badge icon={faEye} className="sl-bg-danger sl-ml-0 ">
       Internal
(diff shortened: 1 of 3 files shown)


wmhilton:
> I also updated the layout so that the icon and text are aligned...I don't think this has a negative ripple effect anywhere, but I am not super familiar with Elements, so please let me know if I should undo this.

I'm pretty sure this is fine.

> The `JSONSchema7` type does not include the `internal` prop. I am not sure if this is as simple as getting the type up-to-date or if there's something I am overlooking.

I think for Models we don't do any normalization, so instead of `internal` it's `x-internal`. And you'll probably just need to add a `// @ts-expect-error` comment to deal with the typings.

wmhilton (commented):
Looks good so far!

paulatulis:
> I think for Models we don't do any normalization, so instead of `internal` it's `x-internal`. And you'll probably just need to add a `// @ts-expect-error` comment to deal with the typings.

Ah thank you! This was much simpler than I thought it would be.

packages/elements-core/src/components/Docs/HttpOperation/Badges.tsx
```diff
@@ -34,6 +36,17 @@ export const DeprecatedBadge: React.FC = () => (
   </Tooltip>
 );
 
+export const InternalBadge: React.FC = () => (
+  <Tooltip
+    position={Position.BOTTOM_LEFT}
+    content="This endpoint/model is marked as internal and won't be visible in public docs."
```
philsturgeon, line 42: Maybe we should say operation instead of endpoint. The endpoint is /foo which might have a GET /foo operation and a deprecated POST /foo operation. We mix up these two things a lot in our ecosystem so it might be consistent with another mistaken use, but we deprecate operations not endpoints. If you spot that mistake anywhere else please fix it! 🍻

packages/elements-core/src/components/Docs/Model/Model.tsx
```diff
@@ -6,14 +7,22 @@ import * as React from 'react';
...
 const ModelComponent: React.FC<ModelProps> = ({ data, className, headless }) => {
+  const isInternal = !!data['x-internal'];
+
   return (
     <div className={cn('Model MarkdownViewer', className)}>
       {!headless && data.title !== void 0 && <h1 className={Classes.HEADING}>{data.title}</h1>}
 
+      {isInternal && (
+        <HStack spacing={2} mt={3} mb={12}>
+          {isInternal && <InternalBadge />}
```
mpodlasin, line 23: doubled `isInternal` condition here

packages/elements-core/src/components/Docs/Model/Model.tsx
```diff
@@ -6,14 +7,22 @@ import * as React from 'react';
 
 import { SchemaAndDescription } from '../../SchemaAndDescription';
 import { DocsComponentProps } from '..';
+import { InternalBadge } from '../HttpOperation/Badges';
 
 export type ModelProps = DocsComponentProps<JSONSchema7>;
 
 const ModelComponent: React.FC<ModelProps> = ({ data, className, headless }) => {
+  const isInternal = !!data['x-internal'];
```
mpodlasin, line 15: So models have `x-internal` property and operations have `internal`? 🤔 

Wouldn't it be better to unify that in `http-spec`?

mpodlasin (commented):
Code-wise this looks very good.

packages/elements-core/src/components/Docs/Model/Model.tsx
mpodlasin: Ah, I've seen other comments. @philsturgeon is this okay? 🤔

philsturgeon: I’m ok with it if you folks are. Doesn’t effect product so long as it works.

mpodlasin: I would of course prefer to see it unified.

But if for some reason that is difficult/impossible, I'll accept the PR as is.

packages/elements-core/src/components/Docs/Model/Model.tsx
paulatulis: Thank you for flagging! Addressed with 6e25ce5

packages/elements-core/src/components/Docs/Model/Model.tsx
paulatulis: If it's not a blocker and is something that can happen separately from this PR, I can make an issue for it and address it here as a follow up item. But if it's non-trivial, I'll take care of that first.

PR #1327, opened 2021-05-25
chore(angular): angular 12

marcelltoth:
I just closed a dependabot PR in favor of this.

Does the upgrade to Angular 12 which is the current latest. Had to upgrade TS to 4.2 as well as it was a requirement.

Has a nice coincidence of making the pipeline faster by about 30 seconds as Angular 12 compiles faster and that was on the critical path. (I could remove the manual ngcc call as everything is Ivy by default.) Now it's equally fast/slow as CRA.


philsturgeon (approved):

mmiask (approved):
Awesome stuff! 🎉

PR #1328, opened 2021-05-25
feat(docs): model titles

### LINKED ISSUE #1035: Schema View: Title and Description
## User story

As a docs viewer looking at schemas, I would like more information, what is it called, what is the description.

## Describe the solution you would like 

1. Show the title of the schema, from "title" if it has one, or use whatever logic the ToC uses to guess a title. 
2. If it has a description, make sure we’re showing the description. 
3. Uses the same styling for title + markdown description as we have in http operation, etc

This should all be outside of JSV, it can be part of the page that JSV is embedded in. Might be some crossover with #1030.

Screenshot from 2021-04-09 14-34-25

_Originally posted by @mpodlasin in https://github.com/stoplightio/elements/issues/1033#issuecomment-816650812_

marcelltoth:
Resolves #1035


mmiask (changes requested):
Model title does not fall back to a value from ToC

Screenshot 2021-05-25 at 11 47 33

marcelltoth:
Thank you for spotting that @mmiask, I forgot about the different handling of the API component. It should work now.

packages/elements-core/src/components/Docs/Model/Model.tsx
```diff
@@ -1,20 +1,31 @@
...
-    <div className={cn('Model MarkdownViewer', className)}>
-      {!headless && data.title !== void 0 && <h1 className={Classes.HEADING}>{data.title}</h1>}
+    <div className={cn('Model', className)}>
+      {!headless && title !== undefined && (
+        <Heading size={1} mb={4} fontWeight="semibold">
+          {title}
+        </Heading>
+      )}
 
-      <SchemaAndDescription schema={data} description={data.description} />
```
mallachari, line 17: Perhaps we can remove whole `SchemaAndDescription` as we're not using it here anymore? There's one usage in `CodeComponent` but that could be easily replaced I guess. 
And the `SchemaTitle` from there looks quite archaic.

marcelltoth: I was considering that actually. Didn't know why I didn't do it, I'll inline it into `CodeComponent` (and remove redundant parts.)

mmiask (approved):
Works great! 🎉 I tested both `dev-portal` and `elements-api` both for standalone/common models as well as the ones described in the API documents

At the same time I'm not the biggest fan of passing `nodeTitle` from top of Docs. Having it as a shared prop for all `ParsedDocs` subcomponents doesn't seem necessary to me, I would much rather see it only for `Model`. But since I couldn't figure an alternative to your solution, there's no reason in me blocking this PR, especially that it's working flawlessly 💪 

Let's ship it! 🚀

packages/elements-core/src/components/MarkdownViewer/CustomComponents/CodeComponent.tsx
```diff
@@ -22,6 +27,30 @@ function isPartialHttpRequest(maybeHttpRequest: unknown): maybeHttpRequest is Pa
...
+  title?: string;
+}
+
+const SchemaAndDescription = ({ title: titleProp, schema }: ISchemaAndDescriptionProps) => {
+  const resolveRef = useInlineRefResolver();
+  const title = titleProp ?? schema.title;
+  return (
+    <Box py={2}>
+      {title && (
+        <Flex className="sl-flex sl-items-center flex-1" p={2}>
```
mallachari, line 41: `sl-flex-1` might be better after ui-kit is out and `sl-flex` is unnecessary.
Or even better `<Flex flex={1} alignItems="center" p={2}>`

marcelltoth: Actually there should be need for neither, as I got rid of the outer flexbox. Nice catch, I updated it.

marcelltoth:
> Works great! 🎉 I tested both `dev-portal` and `elements-api` both for standalone/common models as well as the ones described in the API documents

🚀 

> 
> At the same time I'm not the biggest fan of passing `nodeTitle` from top of Docs. Having it as a shared prop for all `ParsedDocs` subcomponents doesn't seem necessary to me, I would much rather see it only for `Model`. But since I couldn't figure an alternative to your solution, there's no reason in me blocking this PR, especially that it's working flawlessly 💪

I know that. Unfortunately I couldn't find any trivial better solution, so I stuck with this.

Issue #1092, opened 2021-04-20
JSV/Params: Markdown in parameter descriptions


philsturgeon:
Can we add simple markdown support in descriptions:

- Links
- Italic
- Bold
- Underline

Twilio use links in their param descriptions:

![image.png](https://images.zenhubusercontent.com/5d1f7c5b8a1fa426d066830f/5d5debb4-77d5-4bac-a491-8cebf695cd91)

https://elements-demo.stoplight.io/?spec=https://raw.githubusercontent.com/twilio/twilio-oai/main/spec/yaml/twilio_api_v2010.yaml#/operations/ListAvailablePhoneNumberTollFree

PR #1329, opened 2021-05-25
feat(Docs): Markdown support in Parameters descriptions

mmiask:
Resolves: #1092 

Screenshot 2021-05-25 at 12 42 34

UPDATE: ~~⚠️ This does not apply to `Response` section~~ It will after merging https://github.com/stoplightio/json-schema-viewer/pull/146

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -50,7 +50,7 @@
     "@stoplight/json": "^3.10.0",
     "@stoplight/json-schema-ref-parser": "^9.0.5",
     "@stoplight/json-schema-sampler": "0.2.0",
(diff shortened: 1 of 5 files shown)


philsturgeon:
Does this cover JSV too?

mmiask:
> Does this cover JSV too?

@philsturgeon no, as JSV is not used for displaying operation parameters. I have a solutionn ready though, would you like me to push it as well?

philsturgeon:
Yeah please! The title was hinting at this: "JSV/Parameters: ..." 

Bascially those two things look really similar and markdown should work in both of them, even though they're different code bases. The end users might be a bit confused that markdown works in one but not the other. :)

Issue #1267, opened 2021-05-14
Make `examples` more independent.


mpodlasin:
Currently examples - `angular`, `react-gatsby` and `react-cra` - require user to make a build of `elements` and `elements-dev-portal` before running them.

On top of that they have `package.json` differences from a regular project, that make them impossible to copy&paste somewhere and run indepedently without refactoring `package.json` first.

We want to change that.

**AC:**
- examples use actually released packages by default (for example `elements@7.0.0.-beta.X` etc.)
- you can clone repo, immediately go to an example folder and run it using only standard commands of that repo (`yarn` & `yarn start`).
- you can copy&paste example folder to another directory and run it in the same way as above
- *and yet* the e2e tests still work and they use *builds from a repo, not released packages*

So far we have 2 potential ways to achieve this:
- use `yarn link` before running e2e tests to direct examples to `dist` folders generated in the repository
- copy & paste example folders to some temporary location, rewrite (using a script) package.json to use the file paths we currently have in them, then run the tests

If however you have any other ideas how to implement this feature, they will be appreciated.

Bonus points:
- being able to run examples using repo builds *outside of e2e tests* in a fairly easy way
- not having to use `yarn reinstall` after every new build of `elements` or `elements-dev-portal` to see changes

philsturgeon:
When working on this please update the following docs with any changes to how things work.

- docs/getting-started/2-integrations/angular.md
- docs/getting-started/2-integrations/gatsby.md
- docs/getting-started/2-integrations/next.md
- docs/getting-started/2-integrations/react.md

PR #1334, opened 2021-05-28
Feat: make examples independent

mmiask:
Resolves: #1267 

This PR does the following things:
- Makes framework integrations in `/examples` folder use public packages of Elements. This way users only need to run `yarn` and `yarn start` if they want to see examples
- Moves integrations that use the current code into `/examples-dev` folder (using script in `package.json`). All of e2e tests will run using examples from that directory
- Updates docs to be in line with the changes mentioned above


CONTRIBUTING.md
```diff
@@ -97,9 +97,10 @@ mainly because we use *Cypress* to run them, a traditionally end-to-end test run
 
 The logical steps in which these tests are run:
 1. You build *Elements* from the current TypeScript source.
-2. You install this *Elements* build into the chosen example project.
-3. You serve the example project on localhost over HTTP port 4200
-4. You run the *Cypress* test suite against this example application.
+2. You copy the contents of predefined example project from `examples` to `examples-dev`.
+3. You install this *Elements* build into the chosen example project.
```
philsturgeon, line 101: What does this line mean?

CONTRIBUTING.md
```diff
@@ -131,6 +135,8 @@ The first 3 steps are the same, but this time, instead of running the tests in h
 yarn 
 # Build Elements itself
 yarn build
+# Copy the contents of predefined example application and make it use the local build
```
philsturgeon, line 138: Is this maybe something that build should do?

CONTRIBUTING.md
mmiask: The solution I decided to take, is that users copy the contents of `examples/angular` (or any other integration), and modify `package.json` to utilize a local build (not the ones from npm). All of this is managed by `yarn copy:angular`

CONTRIBUTING.md
mmiask: It would definitely decrease the number of commands but increase the build time. I don't think it's necessary to copy the whole integration folder each time we run a build

mallachari: Did you consider manipulating original package instead of using that integration folder? Something like modifying package.json temporarily and after example build process reverting it? Or having already prepared `package.json_integration` and replace original one with it before build and reverting after?

mmiask: Yes, I did. The downside is that beacause Gatsby doesn't allow custom output folder for builds, you still would need to copy some stuff around, if you for instance would like to run both of these integrations ('public' and 'development') at the same time.

I like the idea of providing the pre-made `package.json` though 👍

mallachari: But can't the old build be replaced with new one? Would it really require using custom output folders for that?

mmiask: It can, but are there any advantages to that solution? `examples` folder might become a bit confusing for users with that approach IMHO. Each time they would want to compare their work to an already working example (which is currently in `examples`), they would need to build both of them.

Plus, if we create a separate folder for developed integrations, we get much more than just ability to run e2e, without worrying that it can affect existing ('public') examples.

package.json
```diff
@@ -58,12 +58,15 @@
...
-    "serve:react-gatsby": "(cd ./examples/react-gatsby && yarn serve)",
-    "build:react-cra": "(cd ./examples/react-cra && yarn reinstall && yarn build)",
-    "serve:react-cra": "(cd ./examples/react-cra && yarn serve)",
+    "copy:angular": "mkdir examples-dev ; cp -a -v ./examples/angular ./examples-dev/angular ; . ./use-local-elements.sh angular",
+    "build:angular": "(cd ./examples-dev/angular && yarn reinstall && yarn build)",
+    "serve:angular": "(cd ./examples-dev/angular && yarn serve)",
+    "copy:react-gatsby": "mkdir examples-dev ; cp -a ./examples/react-gatsby ./examples-dev/react-gatsby ; . ./use-local-elements.sh react-gatsby",
+    "build:react-gatsby": "(cd ./examples-dev/react-gatsby && yarn reinstall && yarn build)",
+    "serve:react-gatsby": "(cd ./examples-dev/react-gatsby && yarn serve)",
+    "copy:react-cra": "mkdir examples-dev ; cp -a ./examples/react-cra ./examples-dev/react-cra ; . ./use-local-elements.sh react-cra",
```
mallachari, line 67: Might make sense to first remove already existing `examples-dev` directory to have it always up to date

mmiask: Fortunately `cp` overwrites files by default. Or do you mean something else?

mallachari: hmm, it doesn't for me:
```
mkdir examples-dev ; cp -a ./examples/react-cra ./examples-dev/react-cra ; . ./use-local-elements.sh react-cra
mkdir: cannot create directory ‘examples-dev’: File exists
```

mallachari: Right it does now.
One last thing I'm wondering is if it's worth to include copying with build:example command. It would run with every build but don't we need it to behave this way to keep everything up to date?

philsturgeon:
cp overrides files but mkdir will fail if a directory exists. 

--
Phil Sturgeon
Product @ Stoplight.io

mmiask:
> cp overrides files but mkdir will fail if a directory exists.

@philsturgeon that shouldn't be an issue though, because it proceeds to the next command either way

mallachari (approved):
I'm convinced :+1:

CONTRIBUTING.md
philsturgeon: These seem like a list of instructions, steps that somebody has to take, but it is not clear what the contributor is meant to do for this step.

README.md
```diff
@@ -36,11 +36,12 @@ Stoplight Elements comes with a few example integration projects, showing you ho
 - **[bootstrap](./examples/bootstrap)** - A single HTML page utilizing the Web Components distribution via a global script tag.
 
 To run these examples yourself:
-1. Clone this repo and open a command line in the repo's directory.
-2. Run `yarn` to install all dependencies.
-3. Run `yarn build` to build Elements itself.
-4. Run `yarn build:react-cra` to build the *react-cra* example site. An analogous command is available for *react-gatsby* and *angular*, but is not required for **bootstrap**.
-5. Run `yarn serve:react-cra` (or equivalent) to serve the example project on `http://localhost:4200`.
+1. Clone this repo.
+2. Go to `/examples` folder within a cloned repo and open a command line in a given example directory.
```
philsturgeon, line 40:
```suggestion
2. Go to `examples` folder and open an example, e.g.: `examples/angular`.
```

Issue #1333, opened 2021-05-27
Issue with docs component while using badges


mnaumanali94:
If security is enabled (wanted to test badges) when using elements in studio, component breaks with

```
Uncaught Error: Invariant failed: You should not use <Link> outside a <Router>
    at invariant (tiny-invariant.esm.js?b4d2:10)
    at eval (react-router-dom.js?757b:183)
    at updateContextConsumer (react-dom.development.js?3b0b:18304)
    at beginWork (react-dom.development.js?3b0b:18661)
    at HTMLUnknownElement.callCallback (react-dom.development.js?3b0b:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?3b0b:237)
    at invokeGuardedCallback (react-dom.development.js?3b0b:292)
    at beginWork$1 (react-dom.development.js?3b0b:23203)
    at performUnitOfWork (react-dom.development.js?3b0b:22157)
    at workLoopSync (react-dom.development.js?3b0b:22130)
```

PR #1335, opened 2021-05-28
fix: security badge link

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1333

Adds optional `allowRouting` prop to Docs that we can use with API component and which defaults to `false` for external usage.

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.spec.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.spec.tsx
@@ -133,9 +133,7 @@ describe('HttpOperation', () => {
 
     it('should not contain link to Overview for operation with uri by default', () => {
       const { unmount } = render(
-        <Router>
-          <HttpOperation data={{ ...httpOperation }} uri="/reference/todos/openapi.v1.json/paths/~1todos/post" />
-        </Router>,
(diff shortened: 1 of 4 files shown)


philsturgeon (approved):

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.spec.tsx
```diff
@@ -115,17 +115,33 @@ describe('HttpOperation', () => {
...
       );
       const apikeyBadge = getSecurityBadge(/API Key/i);
       expect(apikeyBadge?.closest('a')).toHaveAttribute('href', '/reference/todos/openapi.v1.json?security=api_key');
 
       unmount();
     });
+
+    it('should not contain link to Overview for operation with uri by default', () => {
+      const { unmount } = render(
+        <Router>
```
mpodlasin, line 136: So shouldn't we be able to remove `Router` wrapper from that test, to really make sure it works without react router?

mallachari: Good point!

mallachari: Updated

mpodlasin (approved):
🚀

PR #1336, opened 2021-05-28
feat: export dev portal provider

mpodlasin:
Relates to https://github.com/stoplightio/platform-internal/issues/6255

We need the provider to alter URL of API to - for example - facilitate the local testing.


lottamus (approved):
👍

mallachari (approved):

Issue #1141, opened 2021-04-27
[Spike] Come up with main areas of Elements that need additional testing


marcelltoth:
## Summary

As we discussed last time, there is a need for an additional layer of testing beyond our unit tests, mostly towards the UI piece, to give us confidence that the major functions are intact. This will likely take form of 2-3 real E2E flow tests, and/or optionally some Jest based integration testing. 

## Tasks

The output of this spike is a list. Go through elements, and come up with a list of major areas you would like the E2E flow to touch. Consider what can and what cannot be efficiently tested using unit tests, and try to focus on the ones that cannot.

mpodlasin:
Some ideas after recent bugs we encountered:

- each page *type* - article, service, operation, model, etc. - should have a test if this page *type* is correctly displayed.
- check (custom) icons (???)
- send a request with TryIt
- check generated RequestSample content

marcelltoth:
Closing this as @mpodlasin already completed the spike with the list above.

PR #1373, opened 2021-06-04
fix: ref resolving in dev portal

mpodlasin:
Relates to Qualtrics issues @philsturgeon was mentioning on slack and during standup.

Simple mistake on our part.

I am not providing tests here, because I believe this should be a part of e2e suite - I left a comment in https://github.com/stoplightio/elements/issues/1141.


philsturgeon (approved):

philsturgeon:
Let's tag a beta2 so we can see if this helps Qualtrics and Calendly who also reported this issue.

packages/elements-dev-portal/src/components/NodeContent/index.tsx
```diff
@@ -39,6 +39,7 @@ export const NodeContent = ({ node, Link, hideTryIt, hideMocking }: NodeContentP
                 nodeData={node.data}
                 nodeTitle={node.title}
                 hideTryIt={hideTryIt}
+                useNodeForRefResolving
```
marcelltoth, line 42: Shouldn't this be true by default? 🤔

marcelltoth (approved):

packages/elements-dev-portal/src/components/NodeContent/index.tsx
mpodlasin: I am ok either way. 🤷

lottamus: I'm for making it true by default. Is there a use case for not resolving refs?

marcelltoth: There must be as @mpodlasin  added it as a prop and not by default. @mpodlasin do you know the reason maybe?

PR #1374, opened 2021-06-04
feat: add static html example

### LINKED ISSUE #969: Include Static HTML example
## Summary

We have many examples but we lack the most basic one that even we are going to use: a static HTML page. We should make sure to add it so it facilitates both Phil's documentation needs at #844 and QA.

## Tasks

- [ ] Add a static HTML example
- [ ] Make sure to run the E2E suite against it, just like how we run it against the gatsby and the angular examples.

## Additional context

A real site would likely want to either 
1. Import Elements from unpkg, or
2. embed Elements directly into the HTML page

so it is a self-contained HTML page. no 1. is impossible, 2. is impractical to use for testing purposes. I suggest writing the example using unpkg then using [`cy.intercept`](https://docs.cypress.io/api/commands/intercept) to redirect the request to our locally built Elements version for testing. That way we have both a legitimate example and a way to test it, without having to set up two separate files.

marcelltoth:
Resolves #969 

I had to write a custom script to run this weird dual-SPA site, but that doesn't change the fact that the HTMLs should just load as is.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,11 @@ workflows:
           example-name: angular
           requires:
             - build
+      - run-e2e-tests:
+          name: e2e-static-html
+          example-name: static-html
+          requires:
+            - build
       - release:
           requires:
             - lint-and-check
@@ -138,6 +143,7 @@ workflows:
             - build-docs
             - e2e-react-cra
             - e2e-react-gatsby
+            - e2e-static-html
             - e2e-angular
           filters:
             branches:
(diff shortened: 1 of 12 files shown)


philsturgeon (approved):

marcelltoth:
Approved by @mmiask on Slack

mmiask (approved):
All of my questions were answered during live review on a call. Looks good! Let's merge it whenever tests pass 💪  (there's a CORS issue in Cypress tests)

PR #1377, opened 2021-06-07
fix: resolving relative urls in dev portal

mpodlasin:
Fixes another issue reported by @philsturgeon for Qualtrics.

To test, open project with id `cHJqOjk3NDQ` in storybook. Before fix many urls in markdown documentation are not working (for example visit Overview > API General Instructions > Quickstart).

After fix links work.

The solution is necessary because backend provides encoded links in markdown, but decoded links in `outbound_edges` part of the response.

### DIFF
--- a/packages/elements-dev-portal/src/components/NodeContent/index.tsx
+++ b/packages/elements-dev-portal/src/components/NodeContent/index.tsx
@@ -69,8 +69,8 @@ const LinkComponent: React.FC<{ node: { url: string } }> = ({ children, node: { 
     // Resolve relative file URI with
     const resolvedUri = resolve(dirname(node.uri), url);
     const [resolvedUriWithoutAnchor, hash] = resolvedUri.split('#');
-    const decodedUrl = window.decodeURI(url);
-    const decodedResolvedUriWithoutAnchor = window.decodeURI(resolvedUriWithoutAnchor);
+    const decodedUrl = decodeURIComponent(url);
+    const decodedResolvedUriWithoutAnchor = decodeURIComponent(resolvedUriWithoutAnchor);
     const edge = node.outbound_edges.find(
       edge => edge.uri === decodedUrl || edge.uri === decodedResolvedUriWithoutAnchor,
     );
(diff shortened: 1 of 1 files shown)


packages/elements-dev-portal/src/components/NodeContent/index.tsx
```diff
@@ -69,7 +69,11 @@ const LinkComponent: React.FC<{ node: { url: string } }> = ({ children, node: {
     // Resolve relative file URI with
     const resolvedUri = resolve(dirname(node.uri), url);
     const [resolvedUriWithoutAnchor, hash] = resolvedUri.split('#');
-    const edge = node.outbound_edges.find(edge => edge.uri === url || edge.uri === resolvedUriWithoutAnchor);
+    const decodedUrl = window.decodeURI(url);
+    const decodedResolvedUriWithoutAnchor = window.decodeURI(resolvedUriWithoutAnchor);
```
marcelltoth, line 73: 1. You may want to do `decodeURIComponent` instead, don't you?
2. `window` will break SSR. Node does have a `decodeURIComponent` so either just getting rid of the prefix or preferrably `globalThis.decodeURIComponent` will do the job.

mpodlasin: Do I?

I was under the impression that `decodeURIComponent` was meant for a single elements within a path like `hello`, while `decodeURI` was for paths like `hello/world`?

Those are relative urls, but still urls nonetheless?

philsturgeon (approved):

philsturgeon:
Hey @audieleon this one should get links working for you when merged and tagged. It was reared to spaces as I thought.

packages/elements-dev-portal/src/components/NodeContent/index.tsx
marcelltoth: I think so. Whether they are URLs or not is pretty much irrelevant in this situation IMO. You want to simply URL-decode the entire string. (Which happens to contain a relative path, but that's a coincidence.) [`decodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI) does some magic which you probably don't want.

marcelltoth (approved):

marcelltoth:
If you have the chance, adding some tests would be nice, but we can let it slide.

PR #1378, opened 2021-06-07
fix(try-it): contrast issue of CORS link

### LINKED ISSUE #1131: Contrast issue on network error response
When the send request option is used and there is an error the following response happens

The CORS word is styled in a way that is very low contrast compared to the background

image

URL to reproduce 
https://elements-demo.stoplight.io/?spec=https://assets.zuora.com/zuora-documentation/swagger.yaml#/operations/GET_AllAccountingCodes

marcelltoth:
Resolves #1131 

Nothing too fancy. Unfortunately mosaic's `Link` doesn't have any default stying, so to still show that it may be clickable I applied the same semibold weight we used to have on there.

image


mpodlasin (approved):
🎉

PR #1379, opened 2021-06-07
feat: better copy for errors

### LINKED ISSUE #837: Common Errors Pages
**User Story**

As somebody potentially new to OpenAPI and giving it a try with a handy CLI/React tool, I would like it to be very forgiving if I open up invalid OpenAPI, completely unknown files, so that I can learn and get more experience with doing OpenAPI correctly.

**Details**

Error pages need to exist for when things go wrong:

- **Unknown File** - When something is not OAS2/3 what do we do? 
 
- **Not found** - when any URL is called that is not an operation, model, article, or something else valid. Note that there's two paths to handle here: 1. The API description document itself cannot be fetched 2. the document could be fetched but the subpath doesn't exist.

**Additional Information**

Currently in API component we are checking for a specific shape of JSON object passed as document to API component:

```ts
  if (isOas3(apiDescriptionDocument)) {
    uriMap = computeOas3UriMap(apiDescriptionDocument);
    documentTags = apiDescriptionDocument.tags?.map(tag => tag.name) || [];
  } else if (isOas2(apiDescriptionDocument)) {
    uriMap = computeOas2UriMap(apiDescriptionDocument);
    documentTags = apiDescriptionDocument.tags?.map(tag => tag.name) || [];
  } else {
    console.error('Document type is unknown');
  }
```

If we are not supporting a given document, we are just bailing out with a `console.error` and a "loading screen":

Screenshot from 2021-01-05 13-07-43

Also, current design quirk, when theres an error the powered by Stoplight shows up at the top of the page.

Screen Shot 2021-05-06 at 4 49 35 PM

marcelltoth:
Resolves #837 

As indicated in the issue, pretty much everything is covered already. This makes the error message a bit nicer, but that's it.

### DIFF
--- a/packages/elements/src/containers/API.tsx
+++ b/packages/elements/src/containers/API.tsx
@@ -69,17 +69,17 @@ const APIImpl: React.FC<APIProps> = props => {
 
   const { data: fetchedDocument, error } = useQuery(
     [apiDescriptionUrl],
-    () => fetch(apiDescriptionUrl).then(res => res.text()),
+    () =>
+      fetch(apiDescriptionUrl).then(res => {
+        if (res.ok) {
+          return res.text();
+        }
+        throw new Error(`Unable to load spec, status code: ${res.status}`);
+      }),
     {
       enabled: apiDescriptionUrl !== '' && !apiDescriptionDocument,
     },
   );
-
-  React.useEffect(() => {
-    if (error) {
-      console.error('Could not fetch spec', error);
-    }
-  }, [error]);
 
   const parsedDocument = useParsedValue(apiDescriptionDocument || fetchedDocument);
   const bundledDocument = useBundleRefsIntoDocument(parsedDocument, { baseUrl: apiDescriptionUrl });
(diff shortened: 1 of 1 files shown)


packages/elements/src/containers/API.tsx
```diff
@@ -69,18 +69,18 @@ const APIImpl: React.FC<APIProps> = props => {
 
   const { data: fetchedDocument, error } = useQuery(
     [apiDescriptionUrl],
-    () => fetch(apiDescriptionUrl).then(res => res.text()),
+    () =>
+      fetch(apiDescriptionUrl).then(res => {
+        if (res.ok) {
+          return res.text();
+        }
+        throw new Error(`Unable to load spec, status code: ${res.status}`);
```
philsturgeon, line 77: Can we say “description” or “document” or “description document” because spec means “OpenAPI itself.” We’ve been consistently moving towards this across the whole ecosystem but spec pops up incorrectly now and then.

mallachari (approved):
Just the `spec` fix and it's good to go

packages/elements/src/containers/API.tsx
marcelltoth: My bad, fixed

PR #1380, opened 2021-06-08
feat(docs): examples for standalone schemas

### LINKED ISSUE #1041: Schema Page Example
## User story.

As a docs viewer looking at schemas, I would like more information, what like how it might look in JSON if it were to show up as JSON somewhere.

## Is your feature request related to a problem?

Response example are great on operations view, but we don't get to see any examples of the schema on the schema view. 

image

Stripe do it well:

image

We could copy the two-col approach we use for operations. Keeping the title, description, then JSV in the left, we can then add a right col with the "Response Example" logic, but just for the model.

marcelltoth:
Resolves #1041


philsturgeon (approved):
Looking good!

mallachari (approved):
Looks good!

PR #1389, opened 2021-06-10
fix(Docs): docs crashing when viewing endpoint

mmiask:
A part of: https://github.com/stoplightio/platform-internal/issues/5372

## **Before**

Screenshot 2021-06-11 at 13 34 07

## **After**

Screenshot 2021-06-11 at 13 36 12


mpodlasin (approved):
Looks good!

PR #1391, opened 2021-06-10
[Not serious - Do not merge] TypeScript metalanguage $ref resolution

marcelltoth:
@mpodlasin  I was annoyed that we can't type the $ref resolver correctly, so I tried... 😇 

And it looks like it works. Doesn't handle everything yet - I'll probably not do the rest because this entire PR was just a random challenge - but it passes the 2 basic tests and also the circular one.

TS 4.3 needed because https://github.com/microsoft/TypeScript/issues/43060


mpodlasin:
Closing with sadness. @marcelltoth still waiting for that article about what happened here! :D

PR #1394, opened 2021-06-11
fix: align jsv and parameters

### LINKED ISSUE #1196: Response Body left padding is too large
Padding should align with padding of Request Parameters and Headers for example. Currently it doesn't.

mpodlasin:
Fixes #1196 

Before:

Zrzut ekranu 2021-06-11 o 14 53 21

After:

Zrzut ekranu 2021-06-11 o 14 50 03


marcelltoth (approved):
@noor-ahmad accepted the visuals, and this is good to go on the technical side, too.

Issue #1268, opened 2021-05-14
Write tests for new dev-portal components and/or hooks


mpodlasin:
- Bring over old tests from network error handling 
- Tests for everything in DevPortal package
- Mostly react testing perhaps, but figure it out as you go

mmiask:
I'm unassigning myself from this for now as we agreed in the team to first finish `platform` stories

PR #1397, opened 2021-06-11
test: deduplicate jest configs

marcelltoth:
Related to #1268 

Currently the jest config and the setup file are duplicated. This wouldn't be that bad but while adding `jest` to `dev-portal` I am about to duplicate it one more time even.

This is a chore that lifts the config to the root level and eliminates duplication.


mpodlasin (approved):
Love it.

Issue #1361, opened 2021-06-02
StoplightProject Platform JWT


philsturgeon:
In elements-dev-portal, we’ll need to update the fetch handlers/hooks to accept an authorization header so Platform can pass in the user’s JWT.

Issue #1400, opened 2021-06-11
Prefilled Auth token for "Try It" feature


bzmw:
## User story.

As an owner of API documentation, I can pass in an Authorization Token in React, so that Customers of my API Documentation can use the "TryIt" functionality without having to search for an API key of their own.

## Describe the solution you'd like

On instantiation of the `StoplightProject` component I would like to pass in a value. 
Ex:
```
<StoplightProject
  sampleAuth="my-sample-auth-token"
  ...
</StoplightProject>
```

## Comments

This is definitely something that I need for my API Reference, our clients are external to my company and definitely do not want to go digging for an API Token.
If this isn't going to prioritized I think I'll end up hacking it in through some React magic for my specific use case.

philsturgeon:
This is a great idea, but it will be tricky to plan out. 

There can be many different types of auth, maybe multiple used at the same time. Give this OAS:

```
components:
  securitySchemes:
    BasicAuth:
      type: http
      scheme: basic
    BearerAuth:
      type: http
      scheme: bearer
    ApiKeyAuth:
      type: apiKey
      in: header
      name: X-API-Key
```

What would the HTML look like for it?

I dont know if we can do dynamic attribute names:

```html
<elements-stoplight-project
  ApiKeyAuthValue="my-sample-auth-token"
  ...
</elements-stoplight-project>
```

Maybe we could try some sort of special syntax:

```html
<elements-stoplight-project
  parameters="ApiKeyAuth:my-sample-auth-token"
  ...
</elements-stoplight-project>
```

Or... something like that. That would predefine values for _any_ parameters then, but it looks a bit odd, and might have a rough time with multiple values. 

The JS interface is easier, we'd just pass an object:

```js
<StoplightProject
  parameters={{ ApiKeyAuth: "my-sample-auth-token" }}
  ...
</StoplightProject>
```

Anyway, curious what you think about this.

@marcelltoth any ideas on your end?

bzmw:
Personally I prefer the special syntax in your second example because it matches closest to the JS interface.

-----
It's not super awesome but could we throw in some custom fields into the OAS to support something like this?
```
components:
  securitySchemes:
    BasicAuth:
      type: http
      scheme: basic
      x-sample-token: my-sample-auth-token
...
```

marekhrabe:
I was trying to achieve this and it seems like the app is already loading the security values from the `localStorage`. 

[...]

PR #1422, opened 2021-06-14
feat: stoplight project authorization

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1361

Adds `authToken` property to `StoplightProject` component and uses it with fetch handlers to platform endpoints.
Tested with private projects on running platform.


marcelltoth (commented):
I'd like to stop for a second to think about the naming of this prop. There are some proposals (https://github.com/stoplightio/elements/issues/1400) to add API related auth functionality to `StoplightProject / API`. 

We may or may not decide to implement such, but in order to protect us from possible name collisions we could consider naming this something like `platformAuthToken` or similar, lining up with `platformUrl`. @philsturgeon WDYT?

mmiask (approved):
Awesome stuff! Looks good, has a test covering the expected behavior - let's merge it!

marcelltoth (changes requested):
To wait for @philsturgeon 's feedback on the prop name

philsturgeon:
I’ve got to admit I’m a bit confused about the auth work we need to do. I’m not sure I fully understand the technical requirements which has an impact on the product requirements.

I thought we wanted to look for JWTs being returned, and pass them back to the service in subsequent requests so that “logged in users” could get relevant content for them, internal, private, etc. It seems like this might not have been correct. 

The most vague suggestion I could make is that if the authToken is the only type of authToken we’ll need then there’s no need to make the variable name longer for clarification.

philsturgeon:
Btw the api interactions will always be limited to StoplightProject. The API component should not ever need any Stoplight API interactions to run, it should always be standalone.

mpodlasin:
> I’ve got to admit I’m a bit confused about the auth work we need to do. I’m not sure I fully understand the technical requirements which has an impact on the product requirements.
> 
> I thought we wanted to look for JWTs being returned, and pass them back to the service in subsequent requests so that “logged in users” could get relevant content for them, internal, private, etc. It seems like this might not have been correct.
> 
> The most vague suggestion I could make is that if the authToken is the only type of authToken we’ll need then there’s no need to make the variable name longer for clarification.

You are essentially correct, but we will not perform authentication (getting a token for the user in platform), because platform already does it.

Platform already has a token and it will pass it to dev-portal components.

@philsturgeon is however kind of correct here - this prop shouldn't be on the `StoplightProject` component itself, because this component is not used in platform. It should be passed only via `DevPortalProvider`, which will be used by platform when using components such as `TableOfContents`, `NodeContent` etc.

This also means that there will be no auth token naming conflict - `authToken` on `DevPortalProvider` is completely enough, but `platformAuthToken` is probably better, because it's more explicit what kind of token is that.

Second auth token that @marcelltoth mentions *will be* on StoplightProject component (and API for that matter) - given we accept that request from the user. But that's a whole separate issue.

mpodlasin (changes requested):
Look at my other comment.

There is no reason for `authToken` to be a `StoplightProject` prop.

Also cc @lottamus to make sure what we are doing here makes sense.

mallachari:
So we don't want StoplightProject to be able to access private projects for external users?

mpodlasin:
"so Platform can pass in the user’s JWT." - this is what the issue says. To achieve that prop on the StoplightProject is unnecessary.

I guess we could allow private docs to be displayed in StoplightProject, but how to achieve that is a whole separate conversation IMHO.

mallachari:
Well then let's have it just in provider. I assumed that people using docs might be in possession of token by doing some kind of external login. But if we don't need that now that's fine, let's focus on making it usable for platform.

mpodlasin:
> Well then let's have it just in provider. I assumed that people using docs might be in possession of token by doing some kind of external login. But if we don't need that now that's fine, let's focus on making it usable for platform.

And note that this will have a big effect on your unit tests.

There should be more tests that hooks pass the authorization token when they are in the `DevPortalProvider` context.

lottamus:
> There is no reason for authToken to be a StoplightProject prop.

Correct, we don't need `authToken` on `StoplightProject` (yet, see below). Platform doesn't use `StoplightProject`, instead it uses the data fetching handlers and passes data directly into TableOfContents and NodeContent components. So we just need to support `authToken` (or some way to pass the user's auth token) in the handlers, so we can send the Authorization header in the HTTP request to Elements API.

> So we don't want StoplightProject to be able to access private projects for external users?

Yeah, this is a bit trickier. We don't expose a way for folks to retrieve an auth token right now. Plus we don't want folks to expose their auth tokens to the public if providing it via a prop on the component. We'll need to think through the best approach to supporting private projects, but it will likely come with some sort of Login/Auth component where folks are required to log into their Stoplight account.

packages/elements-dev-portal/src/hooks/dataFetching.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+  beforeEach(() => {
+    fetchMock.mockResponse(() => Promise.resolve({ status: 200 }));
+    fetchMock.enableMocks();
+  });
+
+  afterEach(() => {
+    fetchMock.disableMocks();
+  });
+
+  it.each<[string, Function, Record<string, string>]>([
```
mpodlasin, line 23: Oh, this looks great!

mpodlasin (approved):
Looks great!

PR #1424, opened 2021-06-14
test: add tests for dev-portal

### LINKED ISSUE #1268: Write tests for new dev-portal components and/or hooks

marcelltoth:
Resolves #1268


philsturgeon:
Have you considered using a VCR approach instead of fixtures? I'm not sure if you've ever done anything like this, but a common problem for API clients is hardcoded fixtures into their tests, the API changes, their tests continue to say they're fine, then you've got a broken application. I have seen this happen 100 times, and its entirely part of the "dont mock what you dont own" mindset. 

https://apisyouwonthate.com/blog/testing-api-client-applications 

We don't have to do this now, but I wanted to know if it was considered. We can make an issue to switch to this in the future if you're interested, but a lot of people are not even aware of the concept.

marcelltoth:
Good call. Fortunately we also have a set of e2e tests that are ran against the real production API, which mostly gets us the confidence that the API is not breaking. We own it and shouldn't be breaking it though 😄 The VCR approach looks interesting, but I don't think it's worth the added complexity at this point.

mpodlasin (approved):
I think that's a good start. 👍

philsturgeon:
Looks good to me!

--
Phil Sturgeon
Product @ Stoplight.io

PR #1426, opened 2021-06-14
chore: bump mosaic to 60, update for breaking Menu changes

marbemac:
Related to https://github.com/stoplightio/mosaic/issues/140

The big breaking change when moving to react-aria ecosystem for menu component is that we provide menu structure as an `items` object, rather than react children. That is the majority of this PR.

One thing you might like @marcelltoth - nice news is that performance of menus has improved significantly :). Still plenty to improve down the line, but it's pretty good for now.

Before:

After (and this is with development react build which is much slower than normal, while the "before" capture above was captured from prod site). Also note that it looks kind of jumpy because the frame rate on the gif is not high enough:


packages/elements-core/src/components/TryIt/MockingButton.tsx
```diff
@@ -23,88 +22,95 @@ export const MockingButton: React.FC<MockingButtonProps> = ({
...
+      };
+
+      return menuItem;
+    }
+
+    return items;
+  }, [code, dynamic, example, isEnabled, operationResponses, setMockingOptions, toggleEnabled]);
 
   return (
     <Box>
```
marbemac, line 97: Here's what it looks like in practice with the updated menu.

packages/elements-core/src/components/Docs/HttpOperation/Responses.tsx
```diff
@@ -43,14 +43,14 @@ export const Responses = ({ responses: unsortedResponses, onStatusCodeChange, on
           <TabList>
             {responses.map(({ code }) => (
               <Tab key={code} id={code}>
-                <Text color={code === activeResponseId ? 'primary' : 'muted'}>{code}</Text>
```
marbemac, line 46: This hasn't been necessary for a while, just cleaned it up as I went I hope that's ok.

packages/elements-core/src/components/TryIt/TryIt.spec.tsx
```diff
@@ -811,12 +828,12 @@ describe('TryIt', () => {
 
     describe('OAuth2 Component', () => {
       it('allows to send a OAuth2 request', async () => {
-        render(<TryItWithPersistence httpOperation={putOperation} />);
+        const { getByLabelText } = render(<TryItWithPersistence httpOperation={putOperation} />);
```
marcelltoth, line 831: Why do you keep switching up `screen.doSomething` for this?

marbemac: Good question - as I was fixing the tests I briefly thought that perhaps the screen thing was causing an issue, so I switched it and did not switch it back. Might not be the case, I'll switch it back NP.

marbemac:
Alrighty so this is good to go except for the create react app and angular examples. Haven't had time to look into it yet for more than 15 min, but getting this after building and serving, on all pages (https://reactjs.org/docs/error-decoder.html/?invariant=130&args[]=undefined&args[]=):

Screen Shot 2021-06-16 at 12 35 53 AM

Not occurring in the demo site or the other examples - just create react app and angular. Before I dig in, @stoplightio/undefined how have you guys been debugging / working with these example apps? I started trying to figure it out but the basic iteration loop of make change, run elements build, run copy example, run build example, run serve (or start), takes forever. Am I missing a better way?

mpodlasin:
I'll take a look at those builds.

They are indeed cumbersome to debug, however I have loooooots of experience by now. 🌊

mpodlasin:
(sorry, accidentally closed) 😂

mpodlasin:
I tested around and it works very well.

My only problem is that the menu expands downwards always. On small screens this gives an illusion of the menu being broken. Scroll works there, so it's usable, but it looks strange.

Zrzut ekranu 2021-06-16 o 13 37 35

marbemac:
> My only problem is that the menu expands downwards always. On small screens this gives an illusion of the menu being broken. Scroll works there, so it's usable, but it looks strange.

Weird - works OK on the mosaic docs site, perhaps something to do with the scrolling containers in this context 🤔. In any case, nice catch! Will see if I have time today to take a peek, otherwise IMHO we might consider tackling this in a separate PR when we have bandwidth.

---

Ooo checks passing - thanks! What was the build issue in the end?

mpodlasin:
@marbemac Sure I would be okay with merging, it's not a huge issue.

Regarding the build - it was an old issue in JSV that we fixed already (caused by those infamous `.ejs` suffixes).

It was not an issue with your PR - it was on main also, because we didn't update JSV yet at that point.

mpodlasin (approved):
I am approving then. 

@marbemac just remember to create an issue for those menus! :)

marbemac:
Ok so that was kindd of a pain to track down ha. @mpodlasin I ended up bumping to mosaic `^1`, but that didn't quite fix it. I ended up having to remove the lock files from the example folders, since something about how yarn resolves packages with local file paths (example-dev elements packages) results in it not cleaning up the old `react-aria/*` packages from the lock file that was copied over from `examples`. This results in multiple react-aria packages being installed - this is bad because they use react context providers, which don't play well when multiple versions of the package are loaded.

It's kind of confusing, happy to hop on a call or continue to chat async if you'd like. But basically since last you reviewed, the main change is to bump to mosaic v1, and delete the example yarn lock files. LMK what you think.

packages/elements-core/src/components/TryIt/RequestBody.tsx
```diff
@@ -43,3 +26,34 @@ export const RequestBody: React.FC<RequestBodyProps> = ({ examples, requestBody,
     </Panel>
   );
 };
+
+function ExampleMenu({ examples, requestBody, onChange }: RequestBodyProps) {
+  const handleClick = React.useCallback(
+    (example: INodeExample | INodeExternalExample) => {
+      onChange(safeStringify('value' in example ? example.value : example.externalValue, undefined, 2) ?? requestBody);
```
philsturgeon, line 33: Do we ever want externalValue in there? Unless http-spec is doing something special, externalValye is a URI to a file. We'd need to follow the URI, see if we understand the file, and return the data (which is already going to be JSON or similar). 

Or we can ignore externalValue for examples, but we cannot use them interchangeably with value.

marbemac: I didn't change the logic of this code in this PR - just updated for the new Menu interface. If ya'll think this is important I can do it now or you guys could tackle in separate PR - cc @marcelltoth @mpodlasin

mallachari (approved):
Looks great!

PR #1430, opened 2021-06-15
feat: hide schemas in toc

### LINKED ISSUE #709: Show/hide "Schemas"
As an Elements user, I can pass a prop into the API component, so I can hide the "Schems" in the left navigation.

<p></p>

- Add a property `hideSchemas` to the API component

- When enabled, we should hide the "Schemas" part of the generated ToC

mpodlasin:
Closes #709

### DIFF
--- a/packages/elements/src/components/API/__tests__/utils.test.ts
+++ b/packages/elements/src/components/API/__tests__/utils.test.ts
@@ -388,7 +388,7 @@ describe('computeAPITree', () => {
     ]);
   });
 
-  it('hides schemas from ToC', () => {
+  it('allows to hide schemas from ToC', () => {
     const apiDocument: OpenAPIObject = {
       openapi: '3.0.0',
       info: {
--- a/packages/elements/src/components/API/__tests__/utils.test.ts
+++ b/packages/elements/src/components/API/__tests__/utils.test.ts
@@ -1,7 +1,7 @@
 import { OpenAPIObject } from 'openapi3-ts';
 
 import { transformOasToServiceNode } from '../../../utils/oas';
-import { computeTagGroups } from '../utils';
+import { computeAPITree, computeTagGroups } from '../utils';
 
 describe('computeTagGroups', () => {
   it('orders endpoints according to specificed tags', () => {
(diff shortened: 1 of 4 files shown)


packages/elements/src/components/API/utils.ts
```diff
@@ -51,7 +51,19 @@ export const computeTagGroups = (serviceNode: ServiceNode) => {
...
+
+const defaultComputerAPITreeConfig = {
+  hideSchemas: false,
+};
+
+export const computeAPITree = (serviceNode: ServiceNode, config: ComputeAPITreeConfig = {}) => {
+  const mergedConfig = {
+    ...defaultComputerAPITreeConfig,
+    ...config,
+  };
```
marcelltoth, line 66: This is a bit dangerous as a config like `{hideSchemas: undefined}` can override the default. Not a huge deal here as the default is `false`, but as a pattern in general, be aware of the problems. A safer tool is something like [_.defaults](https://lodash.com/docs/4.17.15#defaults) or similar property-by-property logic e.g. with a `??`.

packages/elements/src/components/API/__tests__/utils.test.ts
```diff
@@ -326,3 +326,117 @@ describe('computeTagGroups', () => {
...
+        id: '/schemas/ImportantSchema',
+        slug: '/schemas/ImportantSchema',
+        title: 'ImportantSchema',
+        type: 'model',
+        meta: '',
+      },
+    ]);
+  });
+
+  it('hides schemas from ToC', () => {
```
marcelltoth, line 391: *when instructed* or something. It doesn't hide them by default.

marcelltoth (approved):
Only that pattern below is concerning but also it happens to be correct in this case, so 👍.

packages/elements/src/components/API/utils.ts
mpodlasin: I used defaults.

packages/elements/src/components/API/__tests__/utils.test.ts
mpodlasin: Yup, done 👍

PR #1431, opened 2021-06-15
feat(http-service): hide Base URL box when empty

### LINKED ISSUE #1111: Hide API Base URL when there's garbage input
## Describe the bug

The window with URLs is empty in some cases, especially when the URL specified is malformed. It shouldn't appear at all.

## To Reproduce

Visit https://api.apis.guru/v2/specs/britbox.co.uk/3.730.274-ref-1-39-0/openapi.yaml in elements demo.

## Expected behavior

No window when there are no contents.

**Screenshots**
Screenshot from 2021-04-22 10-22-50

Screenshot from 2021-04-22 10-15-40

marcelltoth:
Resolves #1111 

More general than the issue: we would like to hide the box any time it would be empty, as an empty box looks broken.

To break it you can just use `https:///something` (note the triple forward slash).

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpService/ServerInfo.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/ServerInfo.tsx
@@ -7,15 +7,20 @@ import { isProperUrl } from '../../../utils/guards';
 import { getServerUrlWithDefaultValues } from '../../../utils/http-spec/IServer';
 
 interface ServerInfoProps {
-  servers?: IServer[];
+  servers: IServer[];
   mockUrl?: string;
 }
 
 export const ServerInfo: React.FC<ServerInfoProps> = ({ servers, mockUrl }) => {
   const mocking = React.useContext(MockingContext);
   const showMocking = !mocking.hideMocking && mockUrl && isProperUrl(mockUrl);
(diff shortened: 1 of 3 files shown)


philsturgeon (approved):

packages/elements-core/src/components/Docs/HttpService/ServerInfo.tsx
```diff
@@ -7,23 +7,28 @@ import { isProperUrl } from '../../../utils/guards';
...
 
 export const ServerInfo: React.FC<ServerInfoProps> = ({ servers, mockUrl }) => {
   const mocking = React.useContext(MockingContext);
   const showMocking = !mocking.hideMocking && mockUrl && isProperUrl(mockUrl);
-  const productionServer = servers?.[0];
+  const productionServer = servers[0];
   const productionUrl = productionServer && getServerUrlWithDefaultValues(productionServer);
+  const showProductionUrl = productionUrl && isProperUrl(productionUrl);
+
+  if (!showMocking && !showProductionUrl) {
```
mpodlasin, line 21: Yup, that's exactly what I had in mind when I was writing that issue :D

mpodlasin (approved):
Looks good!

Issue #1368, opened 2021-06-02
StoplightProject Stacked Layout


philsturgeon:
## User story.

As a of StoplightProject, I can do use the stacked layout, so that I can  get rid of the sidebar and integrate into a CMS that already has a sidebar.

marbemac:
Couple questions:

1. A stoplight project has many articles and APIs, and an indeterminate depth to the nesting of all of it. What adjustments do you think we need to make to stacked layout UI to accommodate this (and do we think a stacked layout is even reasonable for this much info?)?
2. Thoughts re priority versus everything else?

PR #1437, opened 2021-06-16
StoplightProject Doesn't Have Layout (Yet)

philsturgeon:
- docs: fixed README link to React
- docs: removed layout from StoplightProject It does not support layout yet.

We might bring in a stacked layout for StoplightProject but we have to figure out how it looks first. https://github.com/stoplightio/elements/issues/1368


mpodlasin (approved):
Yup, 💯 correct

README.md
```diff
@@ -58,12 +58,11 @@ import { API } from "@stoplight/elements";
 
 <API
   apiDescriptionUrl="https://api.apis.guru/v2/specs/github.com/1.1.4/openapi.yaml"
-  layout="traditional"
```
mmiask, line 61: Although there is no `traditional` layout in the `API` component, we still have `sidebar | stacked`

Another thing is that we don't have any description of `StoplightProject` in the main README yet. Do you think we should add it there?

mmiask (commented):
Good eye with catching this `layout` inaccuracy  in `StoplightProject` docs. I just left one comment but overall this looks good 👍

README.md
philsturgeon: Nah, we push people to docs and docs takes care of that.

PR #1446, opened 2021-06-17
fix: read/write only params in generated content

### LINKED ISSUE #1369: No readOnly properties in "Response Example"
## Describe the bug

When the property is readOnly it's not showing in "Response Example". If I change it to "Read/Write" - will show.

## To Reproduce

1. Use https://unpkg.com/@stoplight/elements@beta/web-components.min.js 
2. Use yaml file below.

## Expected behavior

Read-only properties should be shown in response examples.

## test.yaml

```
openapi: 3.1.0
paths:
  /websites:
    get:
      responses:
        '200':
          $ref: '#/components/responses/WebsitesNames'
      operationId: get-websites
components:
  responses:
    WebsitesNames:
      content:
        application/json:
          schema:
            type: array
            items:
              type: object
              properties:
                id:
                  type: integer
                  minimum: 1
                  example: 1
                  readOnly: true
                name:
                  type: string
                  example: My Website
                  minLength: 1
              required:
                - id
                - name
```

## Screenshot

Screenshot from 2021-06-03 17-32-32

mpodlasin:
Fixes #1369 

Before:

120624486-54fc7580-c493-11eb-9445-776c2c6f0e62

After:

Zrzut ekranu 2021-06-17 o 15 01 09

### DIFF
--- a/packages/elements-core/src/components/TryIt/useTextRequestBodyState.ts
+++ b/packages/elements-core/src/components/TryIt/useTextRequestBodyState.ts
@@ -12,7 +12,8 @@ import { useGenerateExampleFromMediaTypeContent } from '../../utils/exampleGener
 export const useTextRequestBodyState = (
   mediaTypeContent: IMediaTypeContent | undefined,
 ): [string, React.Dispatch<React.SetStateAction<string>>] => {
-  const initialRequestBody = useGenerateExampleFromMediaTypeContent(mediaTypeContent);
+  const options = React.useMemo(() => ({ skipReadOnly: true }), []);
+  const initialRequestBody = useGenerateExampleFromMediaTypeContent(mediaTypeContent, undefined, options);
 
   const [textRequestBody, setTextRequestBody] = React.useState<string>(initialRequestBody);
 
--- a/packages/elements-core/src/components/TryIt/useTextRequestBodyState.ts
+++ b/packages/elements-core/src/components/TryIt/useTextRequestBodyState.ts
@@ -12,8 +12,9 @@ import { useGenerateExampleFromMediaTypeContent } from '../../utils/exampleGener
 export const useTextRequestBodyState = (
   mediaTypeContent: IMediaTypeContent | undefined,
 ): [string, React.Dispatch<React.SetStateAction<string>>] => {
-  const options = React.useMemo(() => ({ skipReadOnly: true }), []);
-  const initialRequestBody = useGenerateExampleFromMediaTypeContent(mediaTypeContent, undefined, options);
+  const initialRequestBody = useGenerateExampleFromMediaTypeContent(mediaTypeContent, undefined, {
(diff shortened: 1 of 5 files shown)


philsturgeon (approved):

mmiask (approved):
Looks good 👍

packages/elements-core/src/utils/exampleGeneration.ts
```diff
@@ -6,21 +6,25 @@ import React from 'react';
...
   mediaTypeContent: IMediaTypeContent | undefined,
   chosenExampleIndex?: number,
+  options?: GenerateExampleFromMediaTypeContentOptions,
 ) => {
   const document = useDocument();
   return React.useMemo(
-    () => generateExampleFromMediaTypeContent(mediaTypeContent, document, chosenExampleIndex),
-    [mediaTypeContent, document, chosenExampleIndex],
+    () => generateExampleFromMediaTypeContent(mediaTypeContent, document, chosenExampleIndex, options),
+    [mediaTypeContent, document, chosenExampleIndex, options],
```
marcelltoth, line 19: I see...
Well unfortunately the objects pattern doesn't play nice with hooks, but you still can't expect consumers to just memoize them all the time I think. In this case the workaround is probably to split up the options, add them all to the dependency array, then reassemble it when calling the function.

packages/elements-core/src/components/TryIt/useTextRequestBodyState.ts
```diff
@@ -12,7 +12,8 @@ import { useGenerateExampleFromMediaTypeContent } from '../../utils/exampleGener
 export const useTextRequestBodyState = (
   mediaTypeContent: IMediaTypeContent | undefined,
 ): [string, React.Dispatch<React.SetStateAction<string>>] => {
-  const initialRequestBody = useGenerateExampleFromMediaTypeContent(mediaTypeContent);
+  const options = React.useMemo(() => ({ skipReadOnly: true }), []);
```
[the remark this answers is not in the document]
marcelltoth: It could just be a global constant if you want to keep reference identity.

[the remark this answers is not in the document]
mpodlasin: Yeah, but that's the thing. If I take is far away from this place it becomes less readable.

I just wanted to have it inlined, but of course reference identity keeps me from doing so.

Don't complain to me, complain to how hooks work :P

packages/elements-core/src/utils/exampleGeneration.ts
mpodlasin: That... that sounds horrible 😂

marcelltoth: That is

marcelltoth (approved):

PR #1449, opened 2021-06-18
feat: add option to collapse toc

### LINKED ISSUE #1370: Collapse Top Level ToC
ToC is currently automatically expanding everything, which in projects with a lot of content takes up a huge amount of space, as APIs with lots of operations, or articles with lots of sub folders, can be nested to pretty much infinite levels. 

<p></p>

Different people have different preferences over whether they should be expanded to save clicks, or collapsed to save space/scrolling.

<p></p>

Add a new option for StoplightProject of `collapseToc: true` which will collapse everything in a group to the highest level.

- API's will be closed so tags and untagged operations will not show up until expanded.

- Nested folders will be collapsed, so no nested articles will show up.

- Update documentation for Elements (Web and React), and Dev Portal (Web and React).

<p></p>

The project used for this example is <https://philsturgeon.stoplight.io/docs/dev-portal-demo/> which has a projectId of "cHJqOjY1MzEw".

mpodlasin:
Closes #1370 

Before:

Zrzut ekranu 2021-06-18 o 12 49 24

After (with option turned on):

Zrzut ekranu 2021-06-18 o 12 44 35

### DIFF
--- a/packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.spec.tsx
+++ b/packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.spec.tsx
@@ -39,6 +39,35 @@ describe('TableOfContents', () => {
       unmount();
     });
 
+    it('it renders group contents if maxDepthOpenByDefault > 0', () => {
+      const { unmount } = render(
+        <TableOfContents
+          activeId=""
+          tree={[
+            {
+              title: 'Root',
+              items: [
+                {
+                  id: 'targetId',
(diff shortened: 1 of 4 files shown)


philsturgeon (approved):

philsturgeon:
@audieleon

mallachari (approved):
Works great. I was thinking that we might also add a test for StoplightProject as we now have those fetch mocks with fixtures, but if that's too much work the toc one should be sufficient.

PR #1450, opened 2021-06-18
fix: example project id

mallachari:
Updates dev-portal `projectId` for storybook and examples as the current one no longer exists. 
New one comes from `https://elements-examples.stoplight.io` workspace.


mpodlasin (approved):
Yes please 🙏

Issue #1266, opened 2021-05-14
Hide Internal Operations / Models [API Component]


marbemac:
## User story

As a Stoplight Studio user, I can mark operations and models as internal, so that they will be hidden from external users.

## Is your feature request related to a problem?

Studio recently added an Internal toggle, which adds `x-internal: true` to operations and models that should be hidden from logged out users in our hosted docs solution. <https://github.com/stoplightio/platform-internal/pull/6340/>

We should replicate this functionality in elements v7, so that our hosted docs users continue to have this feature, and it will make a handy feature for our OSS users too. 

## Describe the solution you'd like

Elements OSS will not know anything about a user being logged in or not, so it cannot know if it should display internal or not, so we can leave that up to the caller to decide. 

- Add a new component argument: `hideInternal: true/false`

- Add to both API component and StoplightProject component

- Documented anywhere arguments are documented (READMEs and docs/).

- Available for React/Web Component users

- Remove from ToC

- Prevent from routing 

- For StoplightProject: Filter/hide results that have internal: true in the node data property. Look to see if internal: true/false is in node data from API, if not we'll to add it.

marbemac:
Link to feature: https://stoplightio.productboard.com/feature-board/planning/features/7983337

mnaumanali94:
@philsturgeon Think we can take it up considering we did it in the current platform too?

philsturgeon:
@mnaumanali94 i would really appreciate that!

mpodlasin:
I've done part for the API component.

StoplightProject still needs to be implemented.

PR #1461, opened 2021-06-21
feat: option to hide internal operations in API component

mpodlasin:
Related to #1266 (but not closes!!!).

This PR only allows to hide both operations and models.

It only applies to API component.

I've added a story with a component with an internal operation.

Testing approach was taken directly from https://testing-library.com/docs/example-react-router/

Before:

Zrzut ekranu 2021-06-21 o 14 30 36

After (with `hideInternal` option turned on):

Zrzut ekranu 2021-06-21 o 14 30 29


philsturgeon:
@mnaumanali94 heads up, we accidentally started some work on this after that issue had been sat in our Next Sprint column. Your team can either grab this as a starter, or we can close it if you're already further ahead.

mallachari (approved):
Looks good, well tested and works great

PR #1478, opened 2021-06-24
chore: refactor examples layouts & fix stoplight toc layout

### LINKED ISSUE #1443: Refactor examples to use "limited height" layout & add a section in docs about layout.
## Summary

Currently the best way to experience elements is to limit height of the container. That's how demo is written. This creates "two scrolls" layout, where menu and content have separate scrolls.

On the other hand, currently examples don't utilize that. This second, unlimited layout is still valuable, so we can leave it, but let's update all the examples to use "limited height" layout. This will lead users to use the most optimal version.

Also let's create a section in docs explaining this "layout" situation and encourage users to limit the height of the elements container.

## Tasks

- [ ] update all examples to use the same layout as demo
- [ ] add documentation section about making layouts with elements

mpodlasin:
Resolves #1443 

This issue involves small fix in StoplightProject ToC layout, making it behave the same as API component.

If you run examples as copy (`copy:react-cra` etc.) and compare it to currently published examples, you will see that change.

### DIFF
--- a/docs/getting-started/usage/layout.md
+++ b/docs/getting-started/usage/layout.md
@@ -2,11 +2,9 @@
 
 The following section applies both to `API` and `StoplightProject` components. We will focus on `API` as an example.
 
-When you render `API` component on your page without any additional styling, it behaves like a regular `div` element. In particular this means that it will grow in height indefinitely, depending on how much content it has inside. 
+When you render `API` component on your page without any additional styling, it behaves like a regular `div` element. In particular this means that it will grow in height indefinitely, depending on how much content it has inside. We did this because we wanted the component to behave as predictably as possible.
 
-We did this because we wanted the component to behave as predictably as possible, emulating typical HTML `div` behavior.
-
-There exists however a second layout option, which we recommend. `API` has two scrolls built-in inside - one on Table of Contents and one on actual content page. This makes scrolling through massive API docs much more pleasant.
+There exists however a second layout option, which we recommend. `API` has two scrollable areas built-in inside - one on Table of Contents and one on the actual content. This makes scrolling through massive API docs much more pleasant.
 
 In order to achieve that effect, you just have to limit the height of `API` component. 
 
@@ -18,7 +16,7 @@ For example if you have an empty page displaying *only* the `API` component, you
 </body>
 ```
 
-If you have some additional components above and/or below `API`, you can use a flexbox layout This is in fact the layout that we use in majority of our examples:
+If you have some additional components above and/or below `API`, you can use a flexbox layout. This is in fact the layout that we use in majority of our examples:
 
 ```html
 <body>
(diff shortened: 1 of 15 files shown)


philsturgeon (approved):

docs/getting-started/usage/layout.md
```diff
@@ -0,0 +1,47 @@
+# Layout
+
+The following section applies both to `API` and `StoplightProject` components. We will focus on `API` as an example.
+
+When you render `API` component on your page without any additional styling, it behaves like a regular `div` element. In particular this means that it will grow in height indefinitely, depending on how much content it has inside. 
+
+We did this because we wanted the component to behave as predictably as possible, emulating typical HTML `div` behavior.
+
+There exists however a second layout option, which we recommend. `API` has two scrolls built-in inside - one on Table of Contents and one on actual content page. This makes scrolling through massive API docs much more pleasant.
```
marcelltoth, line 9: *two scrolls* -> *two scrollable areas* or *two scrollbars* ?

Maybe add an image here?

examples/bootstrap/index.html
```diff
@@ -62,5 +62,18 @@
...
+        display: flex;
+        flex-direction: column;
+        height: 100vh;
+      }
+
+      main {
+        flex: 1 0 0;
+        overflow: hidden;
+      }
+    </style>
```
marcelltoth, line 76: Can we move this to the `head`? Looks weird here.

examples/react-cra/src/index.css
```diff
@@ -32,3 +32,14 @@ code {
...
+.app-container {
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+}
+
+.main-content {
+  flex: 1 0 0;
+  overflow: hidden;
+}
```
marcelltoth, line 45: Just use the `main` tag filter maybe?

examples/react-gatsby/src/components/layout.css
```diff
@@ -6,3 +6,14 @@ html {
...
   margin: 0;
 }
+
+.Layout {
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+}
+
+.MainContent {
```
marcelltoth, line 16: These are called in 3 different ways across the examples (`.MainContent`, `.main-content`, `main`). Can we maybe unify it?

docs/getting-started/usage/layout.md
```diff
@@ -0,0 +1,47 @@
+# Layout
```
marcelltoth, line 1: Also maybe consider making this page a bit shorter, it has a lot of text which is all great, but I'm afraid people will just be *tl;dr* Nothing major, just maybe merging some paragraphs or some redundant sentences. @philsturgeon ?

marcelltoth (approved):

examples/react-gatsby/src/components/layout.css
mpodlasin: I was just using the conventions that I already encountered in each of the projects :P

Sure, I'll try to unify that, but I don't believe that we have to keep those unified. They are literally different apps. They have even completed different headers. :D

marcelltoth: Good point. 👍

docs/getting-started/usage/layout.md
```diff
@@ -41,7 +39,7 @@ If you have some additional components above and/or below `API`, you can use a f
 </style>
 ```
 
-Note that those are only our recommendations. You can use any other HTML/CSS method to limit the height of `API` component.
+Note that those are only our recommendations. You can use any other HTML/CSS approaches to limit the height of `API` component.
 
-Also, you don't *have to* limit the height. It's just our recommendation, but the `API` component without any additional styling will still work 100% correctly and look aesthetic. You just won't get the "two scrolls" functionality.
+Also, you don't *have to* limit the height. Tthe `API` component without any additional styling will still work 100% correctly and look aesthetic. You just won't get the "two scrollbars" functionality.
```
marcelltoth, line 44: Too late to the party I know, but typo: `Tthe`

Issue #782, opened 2021-01-05
Display API / model version


marbemac:
As a docs consumer, I can see the API or schema version number in the docs sidebar, so that I can distinguish between two APIs with the same name but different versions.

<p></p>

- Add a light gray version identifier to the right end of the API or Schema file's TableOfContents item

- If the node does not have a version or is v0.0 then don't display the version identifier

<p></p>

![](https://pb-files.s3.amazonaws.com/production/attachments/8b405c50886c1c0e0847c1a2fd36f3b3615dab132d7d25d5e92bd3768c77c7c1/image.png)

<p></p>

mpodlasin:
Calculating version for each API/model should happen on the backend - in elements service.

PR #1483, opened 2021-06-25
feat: display api/model version in ToC

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/782

Displays version badge in ToC for items with `version` property.

Requires https://github.com/stoplightio/platform-internal/pull/6879 to fully test with dev-portal 

The design is up to discussion.

Screenshot from 2021-06-25 13-01-25
Screenshot from 2021-06-25 14-19-42


packages/elements-core/src/components/MosaicTableOfContents/utils.ts
```diff
@@ -70,25 +70,10 @@ export function isGroup(item: TableOfContentsItem): item is TableOfContentsGroup
   return Object.keys(item).length === 2 && 'title' in item && 'items' in item;
 }
 export function isNodeGroup(item: TableOfContentsItem): item is TableOfContentsNodeGroup {
-  return (
-    Object.keys(item).length === 6 &&
```
mmiask, line 74: What is the reason for removal of this check?

mmiask (commented):
Looks good! I checked it with the latest version of Platform and it works well 👍

packages/elements-core/src/components/MosaicTableOfContents/utils.ts
mallachari: Optional version means that we can have 6 or 7 items

mmiask (approved):

packages/elements-core/src/components/MosaicTableOfContents/utils.ts
mmiask: Let's make sure we update the `TableOfContents` in Platform, because the current version checks for the length of this item. Currently ToC breaks in v2

noor-ahmad:
Small tweak to the badge: change it from Outline Orange -> **Text/Light - Filled**

mallachari:
The color has been changed. Also it's possible to test it with platform project now

philsturgeon (approved):

PR #1497, opened 2021-06-28
fix: toc node check

mallachari:
Branched out from https://github.com/stoplightio/elements/pull/1483
Removing length check as added optional `version` property in elements service toc result makes it fail the check


marcelltoth:
Note: I applied zenhub metadata based on my best assumptions, @mallachari please make sure it's correct.

marcelltoth (approved):
Go for it 🚀

mmiask (approved):

PR #1501, opened 2021-06-29
feat: make Docs styling more granular

mmiask:
This is needed for https://github.com/stoplightio/platform-internal/issues/6253 in order to hide headings.

Aside from that, it make the styling of Docs a bit more granular and removes unnecessary logic in `HttpService`

The visual outcome is the same as before, this only allows a more precise manipulation of styling.


packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.tsx
```diff
@@ -35,73 +43,56 @@ const HttpOperationComponent = React.memo<HttpOperationProps>(
 
     const hasBadges = isDeprecated || securitySchemes.length > 0 || isInternal;
 
-    if (!headless)
-      return (
-        <Box bg="transparent" className={cn('HttpOperation', className)} w="full">
+    return (
+      <Box bg="transparent" className={cn('HttpOperation', className)} w="full">
+        {!customStyle.noHeading && (
```
marcelltoth, line 48: Wow this is much nicer!

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.tsx
```diff
@@ -18,7 +18,15 @@ import { Responses } from './Responses';
...
 const HttpOperationComponent = React.memo<HttpOperationProps>(
-  ({ className, data: unresolvedData, headless, uri, hideTryIt, hideTryItPanel, allowRouting = false }) => {
+  ({
+    className,
+    data: unresolvedData,
+    uri,
+    hideTryIt,
+    hideTryItPanel,
+    allowRouting = false,
+    customStyle = { noHeading: false },
```
marcelltoth, line 28: Unfortunately you can't default options like this. The problem is that anyone who passes even one prop will override this. I mean if `customStyle={{asd: true}}` then you lose `noHeading: false`.

The solution is usually to use [_.defaults](https://lodash.com/docs/4.17.15#defaults) inside the function.

marcelltoth: Also you don't need that default as your prop is negative. Undefined acts pretty much the same, it's also falsey.

packages/elements-core/src/components/Docs/Docs.tsx
```diff
@@ -50,6 +50,22 @@ interface BaseDocsProps {
    * Allows to use internal routing (requires wrapping with Router component)
    */
   allowRouting?: boolean;
+
+  /**
+   * Allows to customize the layout of Docs
+   */
+  customStyle?: {
```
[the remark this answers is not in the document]
mmiask: That's an absolutely valid point. I'm not happy with `customStyle` as well, `styleOptions` or `layoutOptions` will be better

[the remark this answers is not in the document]
marcelltoth: `layoutOptions` is fine by me, too. I wrote these before I read your discussion

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.tsx
mmiask: Also, I will change `hidePoweredBy` to `showPoweredBy`- first of all hiding is the default behavior, and secondly it allows to get rid of setting these default values

marcelltoth (approved):
Looks good now!

PR #1515, opened 2021-06-30
Layouts Docs were hiding

philsturgeon:
Moved this to guides and added to ToC. Don't forget to add new docs to ToC!


mallachari (approved):
Good point. Easy to forget with custom ToC.

Issue #1291, opened 2021-05-18
Dynamically add elements-dev-portal version


lottamus:
## Summary

Right now we're hardcoding the [elements-dev-portal version](https://github.com/stoplightio/elements/pull/1195/files#diff-3ea33606917a9eee1215ba3bcc2d46f28c2600363729eb1a51456f8ccfb1fbdbR28) when passing it to the Stoplight API. Ideally we could grab this dynamically or at build time.

## Why

The "Stoplight-Elements-Version" header is used by the API to determine the shape of the response. This makes it easier to make changes in the future without breaking folks who are using older versions of Elements.

PR #1523, opened 2021-07-01
feat: dynamic dev-portal version

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1291

Updating elements dev portal version that is sent with a header to elements service.

I went with a update version during build approach as doing it dynamically in runtime caused problems with loading json files in some environments.

The idea is to replace a version value in a js file before build (or while running storybook). No need to set it back to old value in such case. No problem with committing that file as well.


packages/elements-dev-portal/src/handlers/getBranches.ts
```diff
@@ -11,7 +12,7 @@ export const getBranches = async ({
 }): Promise<Branch[]> => {
   const response = await fetch(`${platformUrl}/api/v1/projects/${projectId}/branches`, {
     headers: {
-      'Stoplight-Elements-Version': '1.0.0',
+      'Stoplight-Elements-Version': appVersion,
```
mpodlasin, line 15: At this point I am thinking we want some kind of wrapper for fetch, which will encapsulate this and getting a token for the request. 🤷

mpodlasin (changes requested):
Doesn't work on a Mac

Screenshot 2021-07-01 at 10 27 49

packages/elements-dev-portal/src/handlers/getBranches.ts
philsturgeon: middleware! middleware! 

Maybe something like https://github.com/benjamine/fetch-wrap or any other middleware implementation that looks good. This will let us add all sorts of cool generic functionality, like generic http caching, sunset sniffing, auto-retries, circuit breakers, etc.

mallachari: We could use either some middleware package or just wrap it in a simple module, but we would still need to pass the auth token along url.

mallachari:
> Doesn't work on a Mac
> 

Good catch. Looks like `sed` slightly differs on mac and linux within `-i` argument so I've used platform agnostic `replace`.
It does work now.

packages/elements-dev-portal/src/version.ts
```diff
@@ -0,0 +1,2 @@
+// auto-updated during build
+export const appVersion = '1.0.0';
```
mpodlasin, line 2: I would just put some purposefully stupid version here, to catch if the generation breaks. Like `0.0.0` or something.

mpodlasin (approved):
Works now 🎉 

My other comment is optional

packages/elements-dev-portal/src/version.ts
mallachari: We could but that would get updated anyway during first run anyway. And if we do it would be changed and committed within any next unrelated pr which would look weird as the version wouldn't change then. 
It would make more sense if we were unsetting that value to a placeholder after build, but I'm not sure if that's actually needed. Now we are always keeping that value up to date with actual version. Only scenario when it gets out of sync is when a release maker doesn't run build or storybook at least once, but then it would just get updated within next PR. 
Though I'm fine with adding that unsetting step in postbuild if you think it's better.

mpodlasin: Ah, I see. So the idea is that we *do* update this file in PRs as well. Okay. 👍

PR #1529, opened 2021-07-01
chore(release): elements-core v7.1.0

mmiask:
Needed for: https://github.com/stoplightio/platform-internal/issues/6253


packages/elements-dev-portal/package.json
```diff
@@ -47,7 +47,7 @@
   },
   "dependencies": {
     "@fortawesome/free-solid-svg-icons": "^5.10.2",
-    "@stoplight/elements-core": "^7.0.0",
+    "@stoplight/elements-core": "~7.0.0",
```
marcelltoth, line 50: Can we enforce >= 7.0.1 here? According to our best knowledge 7.0.0 does **not** work, so would you make this `~7.0.1` instead? Same for `elements`.

marcelltoth (changes requested):

marcelltoth (approved):
You're right, it's not 7.0.1 but 7.1.0. Looks great now, thank you!

mpodlasin (commented):
LGTM

mpodlasin:
Lol, I thought it's still open for some reason :D

Issue #1398, opened 2021-06-11
Enhance e2e suite


mpodlasin:
## Summary

Currently our suite of e2e tests is *extremely* minimalistic. Meanwhile we had a fair dose of arguably app breaking bugs. We want to catch the most critical errors with our e2e suite.

## Tasks

Create more tests.

In the spike https://github.com/stoplightio/elements/issues/1141 we created following test ideas:

- each page type - article, service, operation, model, etc. - should have a test if this page type is correctly displayed.
- send a request with TryIt
- check generated RequestSample content
- validate that JSV is working
- basic test if ref resolving works in Docs, examples etc.
- test if mocking in stoplight project works

You can of course add more tests, if you come up with something.

Remember that e2e tests should be highly general. Preferably each aspect of application (for example TryIt) gets no more than 1, *max* 2 tests.

mpodlasin:
There also appeared issues with props not being passed correctly to web components.

Therefore perhaps it would be valuable to make e2e tests for props like `hideTryIt, `hideSchemas`?

Up to debate.

PR #1540, opened 2021-07-04
test: enhance e2e suite

mallachari:
Fixes https://github.com/stoplightio/elements/issues/1398

Adds bunch of missing e2e tests.

### DIFF
--- a/cypress/integration/api.ts
+++ b/cypress/integration/api.ts
@@ -31,16 +31,13 @@ describe('API component', () => {
 
     it('displays request sample correctly', () => {
       cy.visit('/zoom-api/operations/userCreate');
-      cy.findByLabelText(
-        'curl --request POST \\ --url \'https://api.zoom.us/v2/users?access_token=\' \\ --header \'Content-Type: application/json\' \\ --data \'{ "action": "create", "user_info": { "email": "string", "first_name": "string", "last_name": "string", "password": "string", "type": 1 } }\'',
-      ).should('exist');
+      cy.findByLabelText(/curl/i).should('exist');
+      cy.findByLabelText(/--request post/i).should('exist');
     });
 
     it('displays example correctly', () => {
-      cy.visit('/zoom-api/operations/userCreate');
-      cy.findByLabelText(
-        '{ "email": "string", "first_name": "string", "id": "string", "last_name": "string", "type": 1 }',
-      ).should('exist');
+      cy.visit('/zoom-api/operations/users');
+      cy.findByLabelText(/{.+}/).should('exist');
     });
 
     it('displays schema correctly', () => {
@@ -52,7 +49,7 @@ describe('API component', () => {
     it('invokes TryIt request (receives network error)', () => {
       cy.visit('/zoom-api/operations/userCreate');
       cy.findByRole('button', { name: /Send Request/i }).click();
-      cy.findByText('Network Error occured.').should('exist');
+      cy.findByRole('button', { name: 'Error' }).should('exist');
     });
 
     it('navigates on auth badge', () => {
@@ -72,9 +69,7 @@ describe('API component', () => {
 
     it('displays example correctly', () => {
       cy.visit('/zoom-api/schemas/Account');
-      cy.findByLabelText(
-        '{ "email": "string", "first_name": "string", "last_name": "string", "options": { "meeting_connectors": "string", "pay_mode": "master", "room_connectors": "string", "share_mc": false, "share_rc": false }, "password": "string" }',
-      ).should('exist');
+      cy.findByLabelText(/{.+}/).should('exist');
     });
   });
 
--- a/cypress/integration/api.ts
+++ b/cypress/integration/api.ts
@@ -4,8 +4,78 @@ describe('API component', () => {
     cy.redirectUnpkgToLocalDistFolder('elements');
   });
 
-  it('loads correctly', () => {
-    loadZoomApiPage();
+  describe('Api page', () => {
+    it('loads api page correctly', () => {
+      loadZoomApiPage();
+      cy.findByText('https://api.zoom.us/v2').should('exist');
+      cy.findByRole('heading', { name: /API Key/ }).should('exist');
+      cy.findByText('API Description').should('exist');
+    });
+  });
+
+  describe('Operation page', () => {
+    it('loads operation docs correctly', () => {
+      cy.visit('/zoom-api/operations/groupCreate');
+      cy.findByRole('heading', { name: /Create a group/i }).should('exist');
+      cy.findByRole('heading', { name: /Request/i }).should('exist');
+      cy.findByRole('heading', { name: /post \/groups/i }).should('exist');
+      cy.findByRole('heading', { name: /Response/i }).should('exist');
+    });
+
+    it('does not break on select dropdown', () => {
+      loadZoomApiPage();
+      cy.visit('/zoom-api/operations/groups');
+      cy.findByRole('button', { name: /request sample/i }).click();
+      cy.findByText('Obj-C').should('exist');
+    });
+
+    it('displays request sample correctly', () => {
+      cy.visit('/zoom-api/operations/userCreate');
+      cy.findByLabelText(
+        'curl --request POST \\ --url \'https://api.zoom.us/v2/users?access_token=\' \\ --header \'Content-Type: application/json\' \\ --data \'{ "action": "create", "user_info": { "email": "string", "first_name": "string", "last_name": "string", "password": "string", "type": 1 } }\'',
+      ).should('exist');
+    });
+
+    it('displays example correctly', () => {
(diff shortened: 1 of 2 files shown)


cypress/integration/stoplight-project.ts
```diff
@@ -3,32 +3,122 @@ describe('Stoplight component', () => {
...
+      cy.findByText('403 Forbidden').should('exist');
+      cy.findByText('"NOT AUTHORIZED"').should('exist');
+    });
+
+    it('mocks response correctly', () => {
+      loadListTodosPage();
+      cy.findByRole('button', { name: /mocking/i }).click();
+      cy.findByRole('menuitemcheckbox', { name: /enabled/i }).should('exist');
+      cy.findByRole('menuitemcheckbox', { name: /enabled/i }).then(enabled => {
+        enabled.click();
```
mallachari, line 64: This is a kind of hack to make it work as the  `menuitemcheckbox` element is detached from dom when invoking regular `click` by chained cypress api.

cypress/integration/stoplight-project.ts
```diff
@@ -3,32 +3,121 @@ describe('Stoplight component', () => {
...
+      cy.findByRole('button', { name: /send request/i }).click();
+      cy.findByText('403 Forbidden').should('exist');
+      cy.findByText('"NOT AUTHORIZED"').should('exist');
+    });
+
+    it('mocks response correctly', () => {
+      loadListTodosPage();
+      cy.findByRole('button', { name: /mocking/i }).click();
+      cy.findByRole('menuitemcheckbox', { name: /enabled/i }).then(enabled => {
+        enabled.trigger('click');
```
mallachari, line 63: This kind of hack makes it work as the `menuitemcheckbox` element is detached from dom when invoking regular `click` by chained cypress api.

cypress/integration/api.ts
```diff
@@ -4,27 +4,85 @@ describe('API component', () => {
...
+      loadZoomApiPage();
+      cy.visit('/zoom-api/operations/groups');
+      cy.findByRole('button', { name: /request sample/i }).click();
+      cy.findByText('Obj-C').should('exist');
+    });
+
+    it('displays request sample correctly', () => {
+      cy.visit('/zoom-api/operations/userCreate');
+      cy.findByLabelText(
+        'curl --request POST \\ --url \'https://api.zoom.us/v2/users?access_token=\' \\ --header \'Content-Type: application/json\' \\ --data \'{ "action": "create", "user_info": { "email": "string", "first_name": "string", "last_name": "string", "password": "string", "type": 1 } }\'',
```
marcelltoth, line 35: This may be too specific to the implementation. I'm thinking: There's about a million ways to send the same request via `curl`, but enough if you just think about the order of arguments. This will break even when the result is still perfectly correct. This may be a compromise we accept of course, but I'm thinking that as this is an e2e that is supposed to be high level (and we already have a unit test for the same topic) it may be enough here to assert some basics that are less fragile

cypress/integration/api.ts
```diff
@@ -4,27 +4,85 @@ describe('API component', () => {
...
+    it('displays schema correctly', () => {
+      cy.visit('/zoom-api/operations/userCreate');
+      cy.findAllByText('action').should('exist');
+      cy.findAllByText('password').should('exist');
+    });
+
+    it('invokes TryIt request (receives network error)', () => {
+      cy.visit('/zoom-api/operations/userCreate');
+      cy.findByRole('button', { name: /Send Request/i }).click();
+      cy.findByText('Network Error occured.').should('exist');
```
marcelltoth, line 55: hm... Again, this is testing the copy. Maybe checking for `/error/i` would be enough?

cypress/integration/api.ts
```diff
@@ -4,27 +4,85 @@ describe('API component', () => {
...
+    it('loads schema page correctly', () => {
+      cy.visit('/zoom-api/schemas/Account');
+      cy.findByRole('heading', { name: /Account/i }).should('exist');
+      cy.findAllByText('first_name').should('exist');
+      cy.findAllByText('pay_mode').should('exist');
+    });
+
+    it('displays example correctly', () => {
+      cy.visit('/zoom-api/schemas/Account');
+      cy.findByLabelText(
```
marcelltoth, line 75: Same as above regarding the `curl` test.

cypress/integration/stoplight-project.ts
```diff
@@ -3,32 +3,121 @@ describe('Stoplight component', () => {
...
-  it('loads correctly', () => {
-    loadStoplightProjectPage();
+  describe('Api page', () => {
+    it('loads api page correctly', () => {
+      loadStoplightProjectPage();
+      cy.findByText('To-dos').click();
+      cy.findByText('https://todos.stoplight.io/').should('exist');
+      cy.findByText('https://stoplight.io/mocks/elements-examples/studio-demo/389434').should('exist');
+      cy.findByRole('heading', { name: /API Key/ }).should('exist');
+      cy.findByText('💫 Overview').should('exist');
```
marcelltoth, line 13: Copy. I'd vote for removing this one, we probably won't lose much.

cypress/integration/stoplight-project.ts
```diff
@@ -3,32 +3,121 @@ describe('Stoplight component', () => {
...
+
+    it('does not break on select dropdown', () => {
+      loadCreateTodoPage();
+      cy.findByRole('button', { name: /request sample/i }).click();
+      cy.findByText('Obj-C').should('exist');
+    });
+
+    it('displays request sample correctly', () => {
+      loadCreateTodoPage();
+      cy.findByLabelText(
```
marcelltoth, line 34: As above.

cypress/integration/stoplight-project.ts
```diff
@@ -3,32 +3,121 @@ describe('Stoplight component', () => {
...
+        'curl --request POST \\ --url \'https://todos.stoplight.io/todos?apikey=\' \\ --header \'Content-Type: application/json\' \\ --data \'{ "name": "string", "completed": false }\'',
+      ).should('exist');
+    });
+
+    it('displays example correctly', () => {
+      loadCreateTodoPage();
+      cy.findByLabelText(
```
marcelltoth, line 41: As above.

cypress/integration/stoplight-project.ts
marcelltoth: These Cypress things... 🤔 

If we are doing manual events though, have you considered actually using `user-event`? For the same reasons why we're doing it in jest. (e.g. This doesn't pass a proper Event object to our handler, it'd crash if we accessed it. Or there's no `mouseover` before)

cypress/integration/stoplight-project.ts
```diff
@@ -3,32 +3,121 @@ describe('Stoplight component', () => {
...
+
+    it('mocks response correctly', () => {
+      loadListTodosPage();
+      cy.findByRole('button', { name: /mocking/i }).click();
+      cy.findByRole('menuitemcheckbox', { name: /enabled/i }).then(enabled => {
+        enabled.trigger('click');
+      });
+      cy.findByRole('button', { name: /send request/i }).click();
+      cy.findByText('200 OK').should('exist');
+      cy.findAllByText('"2019-08-24T14:15:22Z"').should('exist');
```
marcelltoth, line 67: Why this date? Isn't this an implementation detail of the mocker? (Or is it from an example?)

cypress/integration/stoplight-project.ts
```diff
@@ -3,32 +3,121 @@ describe('Stoplight component', () => {
...
+    it('loads schema page correctly', () => {
+      loadTodoSchemaPage();
+      cy.findByRole('heading', { name: /Todo/i }).should('exist');
+      cy.findByText('id').should('exist');
+      cy.findByText('completed_at').should('exist');
+    });
+
+    it('displays example correctly', () => {
+      loadTodoSchemaPage();
+      cy.findByLabelText(
```
marcelltoth, line 81: Again, maybe just `{` to be less specific here?

marcelltoth (approved):
Lots of tiny improvement ideas, but it's looking great in general, I'm approving. Please consider what I wrote below before merging.

cypress/integration/api.ts
mallachari: That's true. I simplified it

cypress/integration/stoplight-project.ts
mallachari: Deleted

cypress/integration/stoplight-project.ts
mallachari: It comes from an example. I removed it, 200 OK should be enough.

cypress/integration/stoplight-project.ts
mallachari: Do you mean to use `user-event` from `testing-library`? Would it work with cypress?

marcelltoth (approved):
Awesome! Thank you!

Issue #1383, opened 2021-06-08
Consider replacing Dependabot with a custom script for examples, too


marcelltoth:
## Summary

@mpodlasin proposed that currently maintaining dependencies of examples takes a lot of manual labor. This is mainly because of the sheer amount of total dependencies we have (3 examples with around 10 deps each comes around 30 deps, most with frequent updates). Additionally, dependabot does not update transient dependencies unless something higher up in the tree gets updated. This means we are still not always testing against the latest frameworks.

The idea is to replace dependabot with the same custom weekly CircleCI script we have in place for main-project updates. 

There should be a separate job for each example, but a single job for every single dep of that example.

## Tasks

- [ ] Implement automatic CircleCI job
- [ ] Disable Dependabot's update PRs.

PR #1549, opened 2021-07-05
chore: less noisy dependabot

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1383

Dependabot got a recent [update](https://github.blog/changelog/2021-05-21-dependabot-version-updates-can-now-ignore-major-minor-patch-releases/) that allows to filter out bumps based on semver scope.
This PR makes it ignore all the patch and minor bumps leaving only major ones. The rest should be installed anyway in a latest version because of `^` usage.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,9 @@ version: 2
 updates:
   - package-ecosystem: npm
     directory: /examples/angular
+    ignore:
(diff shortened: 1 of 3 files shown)


philsturgeon (approved):

mpodlasin (approved):
🙏

Issue #1526, opened 2021-07-01
With new mosaic versions, badges look weird


mpodlasin:
## Describe the bug

After upgrading to mosaic 1.1.1, badges on service page look like this:

<img width="158" alt="Screenshot 2021-07-01 at 13 07 34" src="https://user-images.githubusercontent.com/7916523/124115819-87879700-da6e-11eb-825c-4426269cecb7.png">

That's because default styling changed to:

<img width="122" alt="Screenshot 2021-07-01 at 13 14 53" src="https://user-images.githubusercontent.com/7916523/124115853-90786880-da6e-11eb-9c8e-04c7474f2a1b.png">

And we were by hand overwriting colors.

## To Reproduce

1. Upgrade mosaic to 1.1.1
2. Run storybook in packages/elements

## Expected behavior

The font color and border should match Figma.

- Overview: Version, Security?
- Operation: Deprecated, Internal, Security, any other badges anywhere.

## Additional context
Would be good to rethink how we style badges, so that those errors don't appear in the future. Perhaps mosaic should have more colors or have some way to allow safe styling. Thoughts @marbemac ?

PR #1563, opened 2021-07-06
fix: badges

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1526

Brings back previous badges styles.

Screenshot 2021-07-06 at 20 23 38
Screenshot 2021-07-06 at 20 23 19
Screenshot 2021-07-06 at 20 26 40

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpOperation/Badges.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/Badges.tsx
@@ -4,7 +4,7 @@ import { HttpSecurityScheme } from '@stoplight/types';
 import React from 'react';
 import { Link } from 'react-router-dom';
 
-import { badgeDefaultColor } from '../../../constants';
+import { badgeDefaultBackgroundColor, badgeDefaultColor } from '../../../constants';
 import { getReadableSecurityName } from '../../../utils/oas/security';
 
 export const DeprecatedBadge: React.FC = () => (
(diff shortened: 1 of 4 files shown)


philsturgeon:
@mallachari i think we might have some funny padding here. Looks like more on the top and left than the bottom and right.

mallachari:
> @mallachari i think we might have some funny padding here. Looks like more on the top and left than the bottom and right.

Hmm, if so that must have changed in mosaic as we don't control that. We could hard code that too, but that doesn't make much sense. I'll check it in mosaic.

mallachari:
@philsturgeon I think that was just my browser in zoom mode, it looks fine now.
Screenshot 2021-07-06 at 20 59 46
 
There's nothing in mosaic that changed regarding that. Check out https://deploy-preview-1563--stoplight-elements-dev-portal-storybook.netlify.app

mmiask (approved):
Looks good! 🚀

philsturgeon:
Sorry @mallachari but when I view in Firefox (v89) im seeing this and its certainly not evenly spaced.

Screen Shot 2021-07-07 at 12 21 55 PM

philsturgeon:
I'll approve this because currently things are super broken and this is less broken, if you want to make a new issue to dig into why it looks funny later? Is that a fair trade?

philsturgeon (approved):

mallachari:
Yup, good enough. I'll create issue for that.

PR #1567, opened 2021-07-07
fix: do not navigate with empty slug

mallachari:
Fixes: https://github.com/stoplightio/platform-internal/issues/6995
Requires https://github.com/stoplightio/platform-internal/pull/7031 to run as intended but is also required by the same PR to avoid breaking change.

Blocks navigation for toc nodes with empty `''` slugs.
Doesn't force groups to open if navigation is blocked.

Blocking an empty slug might have a downside if someone intentionally sets it this way (f.e. in custom toc) to navigate to the root of the project (which opens the first item). Though I don't see it as something probable.

Instead of empty slug we could:
- set it as `undefined` - that would require modifying guards to accept optional slug
- add `shouldNavigate` property - would make it more clear though adds yet another property which might be unnecessary 

For testing in dev portal storybook you may use pr-env from linked platform PR:
platform URL: https://x-7031.stoplight-dev.com
projected: cHJqOjU5
 
https://user-images.githubusercontent.com/14196079/124758974-6a037300-df2f-11eb-8d32-b1b7382c62ce.mov

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.1.1",
+  "version": "7.1.2",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**"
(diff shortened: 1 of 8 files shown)


mpodlasin (approved):
👍

PR #1571, opened 2021-07-08
fix: gatsby dependencies and build process

### LINKED ISSUE #1518: Unable to generate sample code due to util.format undefined
Customer reported sample code still not working. "Unable to generate sample code" is showing, and this is the error in the console.

mpodlasin:
Fixes #1518 

- `process` and `util` are moved to elements-core dependencies. Those are peer dependencies of httpsnippet. Without them build doesn't work in environments which don't stub node.js modules themselves (like webpack 5).
- `jsdom` and `canvas` were required by markdown-viewer for SSR. I tested removing those packages and everything works, so it seems with new versions they are not necessary anymore.
- I also double-check our `gatsby-node` config and removed `http` and `https` stubs. I am not sure which library required those, but it seems that they are not needed anymore - everything works without them.

### DIFF
--- a/examples/react-gatsby/gatsby-node.js
+++ b/examples/react-gatsby/gatsby-node.js
@@ -1,9 +1,7 @@
-exports.onCreateWebpackConfig = ({ stage, loaders, actions, plugins }) => {
+exports.onCreateWebpackConfig = ({ stage, actions, plugins }) => {
   actions.setWebpackConfig({
     resolve: {
       fallback: {
-        "http": false,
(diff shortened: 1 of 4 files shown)


philsturgeon:
@audieleon this will fix your latest "Unable to generate request sample" issue.

philsturgeon (approved):

mallachari (approved):
All works fine.
Not sure about `http` and `https` either but it seems to not interfere.

PR #1578, opened 2021-07-08
fix(toc): find first node

mallachari:
Addresses: stoplightio/platform-internal#6995 and follows https://github.com/stoplightio/elements/pull/1567

Ignores group nodes with an empty slug when searching for the first node. This way we can skip http services with no description (empty slug) and redirect to the next existing node. 

For testing in dev portal storybook you may use pr-env from linked platform PR:
platform URL: https://x-7031.stoplight-dev.com

https://user-images.githubusercontent.com/14196079/124975015-d1075180-e02d-11eb-844b-418f369a3a91.mov

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.1.3",
+  "version": "7.1.4",
   "sideEffects": [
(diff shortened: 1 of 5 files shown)


philsturgeon (approved):

mpodlasin (approved):
🎉

PR #1584, opened 2021-07-09
chore: quiet dependabot

marcelltoth:
Followup to #1549 

This should hopefully solve the issue. 

IDK why the ignore approach wasn't working TBH, this is just a different - hopefully more supported - way of doing the same thing. What the `increase-if-necessary` strategy does is that it only changes `package.json` if the new package version falls out of the current range constraint. That should mean that wherever we have carets, we'll only get notified of major updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,9 +2,6 @@ version: 2
 updates:
   - package-ecosystem: npm
     directory: /examples/angular
-    ignore:
-      - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


mallachari (approved):
I'm not sure about that ignore case. It is exactly as documented. Perhaps that is related to how dependabot works and its schedule? 
Anyway, this should also be helpful so if that works then let's have it

marcelltoth:
As I said, I don't know either, what you did looked good to me. But maybe this one will do the trick. 🤷

PR #1592, opened 2021-07-12
chore(release): elements-core v7.1.5

mmiask:
Needed for: https://github.com/stoplightio/platform-internal/issues/7052

Also bumps `markdown-viewer` to `5.1.0`(this changes the `MarkdownViewer` components class, which affects the styling) - for details see: https://github.com/stoplightio/markdown-viewer/pull/100

Releases:
- `elements-core: 7.1.4 -> 7.1.5`
- `elements: 7.0.2 -> 7.0.3`
- `elements-dev-portal: 1.0.4 -> 1.0.5`

### DIFF
--- a/packages/elements-dev-portal/package.json
+++ b/packages/elements-dev-portal/package.json
@@ -48,7 +48,7 @@
   "dependencies": {
     "@fortawesome/free-solid-svg-icons": "^5.10.2",
     "@stoplight/elements-core": "~7.1.5",
-    "@stoplight/markdown-viewer": "^5.0.0",
+    "@stoplight/markdown-viewer": "^5.1.0",
     "@stoplight/mosaic": "^1",
     "@stoplight/path": "^1.3.2",
     "@stoplight/types": "^12.0.0",
--- a/packages/elements-dev-portal/package.json
+++ b/packages/elements-dev-portal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-dev-portal",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "description": "UI components for composing beautiful developer documentation.",
   "keywords": [],
   "sideEffects": [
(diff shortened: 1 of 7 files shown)


marcelltoth (commented):
1. Why is this a `feat`? (Why a minor bump?)

2. Do we intentionally not bump the required elements-core versions in elements and dev-portal? Because of *1.* and the `~` constraint we agreed on, this change actually will *not* surface in elements and dev-portal. Is that expected?

 3. I guess it's related to the above, but why is the old MDV not gone from the yarn lock? Because of the *2*?

marcelltoth:
I didn't mean to tell you to release an elements / dev-portal version. 😅  Only that the dep needs updating. But I guess, it's OK, we wanted to do a release soon anyway.

philsturgeon (approved):

packages/elements-dev-portal/package.json
```diff
@@ -47,8 +47,8 @@
   },
   "dependencies": {
     "@fortawesome/free-solid-svg-icons": "^5.10.2",
-    "@stoplight/elements-core": "~7.1.4",
-    "@stoplight/markdown-viewer": "^5.0.0",
+    "@stoplight/elements-core": "~7.1.5",
+    "@stoplight/markdown-viewer": "^5.1.0",
```
mpodlasin, line 51: Does dev-portal depend directly on markdown-viewer? 🤔

mpodlasin (approved):
🚀

mallachari (approved):

mmiask:
I've the width of `sl-markdown-viewer` to `width: 0` in the stylesheet due to the bug mentioned here: https://bugs.chromium.org/p/chromium/issues/detail?id=464210

packages/elements-dev-portal/package.json
mmiask: `CustomComponentMapping` from `markdown-viewer` is used in `NodeContent`

PR #1593, opened 2021-07-12
chore: use stable version

philsturgeon:
We've been stable a while but didnt update the docs. Oops!

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ The examples will hopefully help show Elements working in close to real world si
 ### React Component
 
 ```bash
-$ npm install @stoplight/elements@beta
+$ npm install @stoplight/elements
 ```
 
 ```js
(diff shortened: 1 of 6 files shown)


mpodlasin (approved):
Yup 😅

bzmw:
@philsturgeon the tags and releases on this repo also do not reflect that 7.0 is the latest and greatest.

philsturgeon:
@bzmw we no longer use tags, as this is a monorepo now and that gets a bit funny. I've made a note to look into lerna + releases again, but for now I've just turned Releases off. Thanks!

philsturgeon:
@bzmw we no longer use tags, as this is a monorepo now and that gets a bit funny. I've made a note to look into lerna + releases again, but for now I've just turned Releases off. Thanks!

PR #1594, opened 2021-07-13
fix: markdown-viewer width

mmiask:
Fixes broken `markdown-viewer` width in Docs

image

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
@@ -1,4 +1,4 @@
-import { Box, Heading, VStack } from '@stoplight/mosaic';
+import { Box, Flex, Heading, VStack } from '@stoplight/mosaic';
 import { withErrorBoundary } from '@stoplight/react-error-boundary';
 import { IHttpService } from '@stoplight/types';
 import cn from 'classnames';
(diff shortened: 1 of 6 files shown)


packages/elements-core/src/styles/_Docs.scss
```diff
@@ -15,6 +15,6 @@
   --fs-code: 12px;
 }
 
-.sl-markdown-viewer {
+.sl-markdown-viewer-wrapper {
```
mpodlasin, line 18: I would just inline width: 0, whenever possible, for consistency's sake

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
```diff
@@ -38,7 +38,13 @@ const HttpServiceComponent = React.memo<HttpServiceProps>(({ className, data, lo
     </>
   );
 
-  const description = data.description && <MarkdownViewer className="sl-mb-5" markdown={data.description} />;
+  const description = data.description && (
+    <Flex>
```
mpodlasin, line 42: Nope, the solution should be **in** `TwoColumnLayout`.

mpodlasin (changes requested):
See comments above.

mpodlasin (approved):
👍 Firefox, safari and chrome look good.

PR #1597, opened 2021-07-14
fix: scroll on toc click

### LINKED ISSUE #1442: Make scrolling after clicking on ToC work in all examples
## Describe the bug

Currently in demo, when we click on an option in ToC, it scrolls main page to the top.

This doesn't happen in examples however, because elements container is not limited to screen width there.

## To Reproduce

1. Go to react-cra. Scroll down. Click on ToC option.

## Expected behavior

It scrolls to the top even when the elements container height is not limited.

So in other words - works in demo *and* examples at the same time.

mpodlasin:
Fixes #1442 

You can test the behaviour in storybooks for `API` and `StoplightProject`, since those don't have limited height.

https://user-images.githubusercontent.com/7916523/125603360-6028431a-377c-4fa2-8015-ffd4eb4c8845.mov

Also fixes an additional problem - when loading a page ToC tries to scroll to the active element in the table of contents.

This is great when ToC has its own scroll, but bad when it doesn't - then the whole webpage is being scrolled to seemingly random places.

### DIFF
--- a/packages/elements-core/src/components/Layout/SidebarLayout.tsx
+++ b/packages/elements-core/src/components/Layout/SidebarLayout.tsx
@@ -6,49 +6,47 @@ type SidebarLayoutProps = {
   sidebar: React.ReactNode;
   maxContentWidth?: number;
   sidebarWidth?: number;
+  children?: React.ReactNode;
 };
 
 const MAX_CONTENT_WIDTH = 1800;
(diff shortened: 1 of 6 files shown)


mmiask (approved):
LGTM, great job! 🎉

philsturgeon (approved):

PR #1598, opened 2021-07-14
fix: relative links

mmiask:
Fixes: https://github.com/stoplightio/platform-internal/issues/7098

### DIFF
--- a/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
+++ b/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
@@ -36,7 +36,7 @@ export const NodeContent = ({ node, Link, hideTryIt, hideTryItPanel, hideMocking
   return (
     <PersistenceContextProvider>
       <NodeLinkContext.Provider value={[node, Link]}>
-        <MarkdownComponentsProvider value={{ link: LinkComponent }}>
+        <MarkdownComponentsProvider value={{ a: LinkComponent }}>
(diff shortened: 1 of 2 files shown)


mpodlasin (approved):
🎉

marcelltoth (commented):
lol, was that it? 😮 I thought it was going to be much more complicated 😄 Good job @mmiask !!!

marbemac:
@marcelltoth @mmiask quick note - `markdown-viewer` has this sort of markdown component override provider built in now (https://github.com/stoplightio/markdown-viewer/pull/95). Could switch to using it rather than the custom one here in elements, and get a nice code reduction win here. The one built into markdown-viewer also has benefit that the custom components will be leveraged even when markdown viewer is rendered in downstream components (for example markdown rendered in json schema viewer descriptions, etc).

mmiask:
> @marcelltoth @mmiask quick note - `markdown-viewer` has this sort of markdown component override provider built in now ([stoplightio/markdown-viewer#95](https://github.com/stoplightio/markdown-viewer/pull/95)). Could switch to using it rather than the custom one here in elements, and get a nice code reduction win here. The one built into markdown-viewer also has benefit that the custom components will be leveraged even when markdown viewer is rendered in downstream components (for example markdown rendered in json schema viewer descriptions, etc).

Oh, nice - I'll definitely check it out, thanks! No need in having custom ones if there's one already in MDV 👍

PR #1607, opened 2021-07-16
fix: reduce react query cache time

mpodlasin:
Related to https://github.com/stoplightio/platform-internal/issues/7001

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.1.7",
+  "version": "7.1.8",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**"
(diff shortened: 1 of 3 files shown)


packages/elements-core/src/hoc/withQueryClientProvider.tsx
```diff
@@ -6,7 +6,7 @@ import { getDisplayName } from './utils';
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
-      staleTime: 1000 * 60 * 5, // 5 minutes
```
mallachari, line 9: That is 15 ms, is it intended?

mpodlasin: Ah damn it, you are right 😂

packages/elements-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.1.7",
```
mallachari, line 3: Might be worth bumping dev-portal too as it's relevant to ninja, not studio only.

mpodlasin: God damnit, you are right again 😂

mallachari (approved):

Issue #1100, opened 2021-04-20
"Folderize" TryIt


mpodlasin:
Currently, TryIt is just one big directory with a flat file list.

Divide those files into subdirectories. For example "Auth", "Mocking"... etc.

![Screenshot from 2021-04-20 16-37-22.png](https://images.zenhubusercontent.com/5f857aaa311cec7e20b4f258/27e1c09c-d900-46dd-bbb2-1f31675b36d8)

PR #1609, opened 2021-07-16
chore: folderize tryit

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1100

Reducing entropy

Screenshot 2021-07-16 at 11 33 44


marcelltoth (approved):
What are we going to swap now? 😢

mallachari:
> What are we going to swap now? 😢

Babel warnings 😉

PR #1624, opened 2021-07-20
chore: no scss

### LINKED ISSUE #964: Remove SCSS
We use node-sass which a C++ extension for SASS, and this can cause install problems. It's especially causing trouble Elements as node-sass does not yet support Node 16. 

To support NodeJS we need to remove node-sass. If we still needed SASS or SCSS then we could use [sass](https://www.npmjs.com/package/sass) instead, but with ui-kit switched over to use pure CSS we can get rid of this entirely.

**Acceptance Criteria**

- [ ] node-sass is gone
- [ ] Swap `scripts/build-styles.js` with PostCSS and rollup. 
- [ ] Quick tidy up of unused CSS rules.
- [ ] NodeJS 16 added to CI checks.

marcelltoth:
Resolves #964 

Removes SCSS. Uses postcss, postcss-import and cssnano instead. In order to do so, I also pulled the storybook config up to the root directory, otherwise I would have had to do it 3 times. (We duplicate the SB config in every project.)

Very likely there's also a slight reduction in CSS bundle size, as `cssnano` is a proper css optimizer tool, while `node-sass` just happens to have a compressed output format, it's not designed for that use-case. 

Also as we have `postcss` set up, now we can add some plugins if users look for it (like autoprefixer, or css modules)


demo/src/components/ElementsAPI.tsx
```diff
@@ -1,4 +1,4 @@
-import '@stoplight/elements-core/styles/elements.scss';
+import '@stoplight/elements-core/styles.css';
```
mpodlasin, line 1: Whoa, whoa.

So does it mean that css endpoint changed for the users?

Aaah... I see that that's core.

Why is demo using core? Can we change that?

mpodlasin (approved):
🚀

Issue #1392, opened 2021-06-11
Create an ES module build


marcelltoth:
## Summary

In order to reduce bundle size and allow consumers to only pull in some of the dependencies (tree shaking) we would like to provide them with a tree-shakable ES module build. This is very important for us as well, `elements` and `elements-dev-portal` only uses pieces of `elements-core` and their dependencies. For example if `elements-core` was an ESM, we could tree-shake FontAwesome in our Web Components build, shaving off about 1MB of bundle size.

## Tasks

This is blocked by https://github.com/stoplightio/scripts/issues/89 so when that is done, it should just be the matter of adopting that `sl-scripts` command. We already eliminated deep-imports into `core` so that should be fine too.

- [ ] Use `sl-scripts bundle` consistently in `elements`.
- [ ] (Optional) Try to propagate the changes upwards in the chain, as much as practical. e.g. if some library we author doesn't have an ESM build yet but could use one, apply the same `build -> bundle` change. **Warning:** that's a breaking change for libraries we deeply import. Only do it for ones that don't support deep imports.

## Additional context

This likely (at least partially) solves #961 #376

PR #1625, opened 2021-07-21
chore: create ES module build

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1392

Builds ES module along with CJS using sl-scripts bundling. 
Adds `styles.min.css` to exports to include it in the encapsulated bundle.

Tested with examples and platform (requires replacing deep imports for elements-core).

### DIFF
--- a/packages/elements-core/src/hooks/useComputeMarkdownHeadings.ts
+++ b/packages/elements-core/src/hooks/useComputeMarkdownHeadings.ts
@@ -1,20 +1,19 @@
 import { MDAST } from '@stoplight/markdown';
 import * as React from 'react';
 import { Parent } from 'unist';
+import { selectAll } from 'unist-util-select';
 
 import { IArticleHeading } from '../types';
-
-const selectAll = require('unist-util-select').selectAll;
 
 export function useComputeMarkdownHeadings(tree: MDAST.Root) {
   return React.useMemo(() => computeMarkdownHeadings(tree), [tree]);
 }
 
 export function computeMarkdownHeadings(tree: MDAST.Root): IArticleHeading[] {
-  return selectAll(':root > [type=heading]', tree)
+  return (selectAll(':root > [type=heading]', tree) as MDAST.Heading[])
     .map((heading: MDAST.Heading) => ({
       title: findTitle(heading),
-      id: heading.data && (heading.data.id as string | undefined),
+      id: (heading.data?.id as string | undefined) || '',
       depth: heading.depth - 1,
     }))
     .filter((item: IArticleHeading) => item.depth >= 1 && item.depth <= 2);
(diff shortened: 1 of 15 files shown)


packages/elements-core/src/hooks/useComputeMarkdownHeadings.ts
```diff
@@ -1,20 +1,19 @@
...
   return React.useMemo(() => computeMarkdownHeadings(tree), [tree]);
 }
 
 export function computeMarkdownHeadings(tree: MDAST.Root): IArticleHeading[] {
-  return selectAll(':root > [type=heading]', tree)
+  return (selectAll(':root > [type=heading]', tree) as MDAST.Heading[])
     .map((heading: MDAST.Heading) => ({
       title: findTitle(heading),
-      id: heading.data && (heading.data.id as string | undefined),
+      id: (heading.data?.id as string | undefined) || '',
```
marcelltoth, line 16: What is this change? Is this related?

packages/elements/package.json
```diff
@@ -96,5 +96,8 @@
   },
   "release": {
     "extends": "@stoplight/scripts/release"
+  },
+  "exports": {
```
marcelltoth, line 100: Aren't you missing the equivalent for `dev-portal`?

marcelltoth (commented):
Looks great, 2 blockers, and I'd have to try installing it somewhere. Have you?

packages/elements-core/src/hooks/useComputeMarkdownHeadings.ts
mallachari: I've replaced `unist-util-select`, earlier imported by `require` and having no typing. This change was to have some typing instead using `any`

packages/elements/package.json
mallachari: Yeah, right. We also export styles in dev-portal. I'll add it

mallachari:
I've installed it in examples and platform. In platform we're using deep imports so it has to be updated there (but all components are accessible directly from index file)

marcelltoth:
> I've installed it in examples and platform. In platform we're using deep imports so it has to be updated there (but all components are accessible directly from index file)

Thank you! I marked your issue to be blocked by stoplightio/platform-internal#7072 It's in our sprint anyway, so you might as well just commit your test code and complete two issues with the same effort ;)

mallachari:
I forgot about that issue. Yup, I'll do that too

mallachari:
Unblocked, ready to go.

marcelltoth:
Doing one more check, then ✅

marcelltoth (approved):
Looks good!

mallachari:
Should we also bump the packages within this PR?

marcelltoth:
Let's do that separately, remember last time:)

Issue #1615, opened 2021-07-20
Always show the overview


mnaumanali94:
Currently we only show overview page if a description exists. This leads to a few concerns:
- The navigation changes
- No way to easily copy baseurl / mockurl
- No way for users to know the overview page even exists

Let's show the overview page always irrespective of the description/summary being there or not.

This applies to both elements standalone and in platform.

mpodlasin:
The page without description does look very, very empty - borderline broken.

I would advocate for having some placeholder content there.

Maybe even something as simple as:

"This API doesn't have any description."

Or maybe @noor-ahmad can help us here. (Ignore other quirks in the screenshot - it's an old one)

![Screenshot 2021-07-05 at 15.14.58.png](https://images.zenhubusercontent.com/5f857aaa311cec7e20b4f258/58a088a0-15ce-4d88-a9b2-97e987a0d001)

PR #1639, opened 2021-07-23
fix: always show overview page in API component

mmiask:
A part of: #1615 

It makes the API component show HTTP Service page at all times, even if there is not description in it (in contrary to the previous behavior)

Doing the same but in `StoplightProject` will happen in https://github.com/stoplightio/platform-internal/pull/7266

**Screenshots**

https://user-images.githubusercontent.com/58433203/126777828-3c6e2433-30a2-4ddc-80cd-48d0acf794d3.mov


mallachari (approved):
Works well. I've added some comments cleanup.

PR #1640, opened 2021-07-23
chore: webpack 4 -> 5 migration

### LINKED ISSUE #1528: Use webpack 5 for building web components
## Summary

While investigating #1518 I realised that we likely use webpack 4 *by accident* to build web components.

Version of webpack was not provided explicitly in package.json and we likely just got the storybook version by accident (which is 4).

Or perhaps it was on purpose and I just don't remember anymore. 😆 

Using 5 would be better because:

- it might have a bit smarter tree-shaking
- it expects us to be more explicit with some dependencies - like it is the case with node dependencies in #1518 which webpack 4 was adding by itself. This made us "forget" to add them to dependencies of elements
- better support going forward

## Tasks

- install webpack 5 in elements and elements-dev-portal packages
- rewrite web-components generating scripts to work with webpack 5 (Gatsby config might help, since Gatsby uses w5)

marcelltoth:
Closes #1528 

Replaces Webpack 4 with 5 everywhere:
- Demo
- Storybooks
- Web-component builds

Not included: examples (they use whatever build system their frameworks pack)


.storybook/main.js
```diff
@@ -12,8 +17,26 @@ module.exports = {
...
+
+    config.module.rules.push({
+      test: /\.m?js/,
+      resolve: {
+        fullySpecified: false
+      }
+    });
+
+    config.plugins.push(new ProvidePlugin({
+      process: require.resolve('process/browser'),
```
mallachari, line 37: Why is it needed to provide `process` here?

mallachari (approved):
Looks and works great. Good job with the transition!

.storybook/main.js
marcelltoth: Well, unfortunately way too many libraries we depend on are not really browser-compatible, mainly they use some node internals, in this case usually the `process.env.*` global. While we can try to eliminate this in the future, this plugin provides a `process` global so the code works for the time being. 

AFAIK we have the same provide plugin in the gatsby config (that's where I took the idea from)

Issue #1621, opened 2021-07-20
Enabling mocking should update the code sample URL


marcelltoth:
@mnaumanali94 commented on [Tue Jul 20 2021](https://github.com/stoplightio/platform-internal/issues/7192)

Enabling mocking should update the code sample URL

![Screenshot 2021-07-20 at 1.18.27 PM.png](https://images.zenhubusercontent.com/5ea9a7f489044245dbd23d46/83ea9fc9-7b22-4b26-8768-d14f7b627359)

---

@marcelltoth commented on [Tue Jul 20 2021](https://github.com/stoplightio/platform-internal/issues/7192#issuecomment-883274232)

I'd say this is an enhancement.

While I can see why this is useful for our customers, such an interaction between `TryIt` and `RequestSamples` was never considered.

---

@marcelltoth commented on [Tue Jul 20 2021](https://github.com/stoplightio/platform-internal/issues/7192#issuecomment-883274595)

Also transfering to `@stoplightio/elements`

PR #1644, opened 2021-07-25
fix: update request samples when mocking

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1621

https://user-images.githubusercontent.com/14196079/126910208-1ed5ed8b-9da0-4422-947c-904769421e3d.mov


marcelltoth (changes requested):
Would you add a test for this please?

Looks great otherwise, thanks for including support for the `Prefer` header as well!

packages/elements-core/src/components/TryIt/TryItWithRequestSamples.spec.tsx
```diff
@@ -28,6 +29,29 @@ describe('TryItWithRequestSamples', () => {
...
+    const TryItWithPersistence = withMosaicProvider(withPersistenceBoundary(TryItWithRequestSamples));
+    render(<TryItWithPersistence httpOperation={putTodosOperation} mockUrl="https://mock-todos.stoplight.io" />);
+
+    const mockingButton = screen.getByRole('button', { name: /mocking/i });
+
+    userEvent.click(mockingButton);
+
+    // enable mocking
+    let enableItem = await screen.getByRole('menuitemcheckbox', { name: 'Enabled' });
+    expect(enableItem).toBeInTheDocument();
```
marcelltoth, line 42: I wouldn't `expect` this, as it's not the main focus of the test. Also `getByRole` will throw anyway, so it's likely just adding noise.

packages/elements-core/src/components/TryIt/TryItWithRequestSamples.spec.tsx
```diff
@@ -28,6 +29,29 @@ describe('TryItWithRequestSamples', () => {
     await waitFor(() => expect(codeViewer).toHaveTextContent(/todos\/123456789/));
   });
 
+  it('reacts to mocking', async () => {
+    const TryItWithPersistence = withMosaicProvider(withPersistenceBoundary(TryItWithRequestSamples));
```
marcelltoth, line 33: This could get pulled out to top level

marcelltoth (approved):
Looks great, merge it!

I put some suggestions but feel free to skip them

PR #1645, opened 2021-07-25
chore: no description overview cleanup

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1615

Revomes no-slug logic. Empty node slug was removed in https://github.com/stoplightio/platform-internal/pull/7266 so we don't need it anymore.

Note: The change in elements-service that always sets slug is not yet in prod - hence storybook still receives items with empty slug which causes it to redirect within this pr.


marcelltoth (approved):
LGTM

Issue #1524, opened 2021-07-01
ToC is not expanded even when its child is active


mpodlasin:
## Describe the bug

When I go directly to url of a specific operation, like `http://localhost:8000/b3A6MTQ-list-all-pets` I see the page correctly, however the ToC is not expanded and because of that the active ToC item is not visible:

<img width="700" alt="Screenshot 2021-07-01 at 12 26 32" src="https://user-images.githubusercontent.com/7916523/124109829-daaa1b80-da67-11eb-814e-46a9ff70ea8d.png">

After expanding ToC (how it should look like):

<img width="700" alt="Screenshot 2021-07-01 at 12 28 20" src="https://user-images.githubusercontent.com/7916523/124109840-dc73df00-da67-11eb-9462-37b1941b682a.png">

## To Reproduce

1. Initiate Gatsby starter for elements-dev-portal: https://github.com/stoplightio/gatsby-starter-elements-dev-portal
2. Run `npm run start` in created directory.
2. Go directly to `http://localhost:8000/b3A6MTQ-list-all-pets` in the browser

## Expected behavior

ToC with active item is expanded.

PR #1648, opened 2021-07-26
fix: expand node when child is active

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1524

ToC groups are initialized during the first render when data might not yet be fetched, which causes `activeId` to be undefined and a group collapsed. Using node id from node slug (which we know already) fixes the issue.


marcelltoth (approved):
Good fix 👏

PR #1650, opened 2021-07-27
chore: update deps, fix warnings - 2

marcelltoth:
## Summary

I got annoyed by all the `yarn install` & `yarn build` warnings, so I went and fixed most of them.

## Fixes

- Much less warnings while `yarn install`ing (some of which are legit hiding-problems).
   Before: image
   After: image

- TypeScript 4.3, jotai 1.0+ (these were needed to solve the above)

- No more circular dependency warning in `elements-core`. 
image

- No more size warning for WC builds (under 2M) 
   Before: image
   After: image

- Added missing dependency to `elements-utils` (legit warning but we ignored it as it was lost among the rest)

## Notes

You might need to `rm -rf node_modules`, there's some yarn bug I ran into.

Draft because built on top of the WP5 branch (to resolve some problems), will rebase once that is merged.


mpodlasin:
dev-portal storybook doesn't work for me locally. 🤷 

Screenshot 2021-07-27 at 16 58 29

marcelltoth:
> dev-portal storybook doesn't work for me locally. 🤷
> 
> Screenshot 2021-07-27 at 16 58 29

Works for me.
image

Have you tried the `rm -rf` I suggested in the description? Yarn is weird.

mpodlasin (approved):
Ah, you were right, it works 💯

PR #1651, opened 2021-07-28
feat: single column overview page

### LINKED ISSUE #1616: Show the URLs / Auth Info in the middle on the overview page
Currently we show the URL info and Auth info on the right side of the overview page. This makes for a very blank overview page in case of empty /small descriptions. 

Let's move the URLs / Auth Info in the middle on the overview page.

The structure should look like this:
- Title / Badges
- URLs / Auth Info
- Description

Additional Info:
- Make the panels wide to reflect the width of the markdown content

marcelltoth:
Resolves #1616

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
@@ -41,7 +41,7 @@ const HttpServiceComponent = React.memo<HttpServiceProps>(({ data, location = {}
           )}
         </Box>
       </VStack>
-      {data.description && <MarkdownViewer className="sl-mb-5" markdown={data.description} />}
+      {data.description && <MarkdownViewer className="sl-my-5" markdown={data.description} />}
     </Box>
   );
 });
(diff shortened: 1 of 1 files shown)


marcelltoth:
I'm investigating why netlify doesn't report success, but you can see from the comments that it did actually deploy just fine.

mnaumanali94 (approved):
Looks awesome!

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
```diff
@@ -1,71 +1,48 @@
...
+      )}
+      <VStack spacing={6}>
+        <ServerInfo servers={data.servers ?? []} mockUrl={mocking.mockUrl} />
+        <Box>
+          {data.securitySchemes?.length && (
+            <SecuritySchemes schemes={data.securitySchemes} defaultScheme={query.get('security') || undefined} />
+          )}
+        </Box>
+      </VStack>
+      {data.description && <MarkdownViewer className="sl-mb-5" markdown={data.description} />}
```
mallachari, line 44: How about adding some top padding/margin here? It's positioned quite close to security schemas.

mallachari (approved):
Looks better for APIs without description. Also I think it looked better before for apis with description. But that's the design so I accept it :)

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
marcelltoth: We certainly can

marcelltoth: I added some margin on top, it certainly looks much better, nice catch.

Issue #707, opened 2020-11-05
Add OpenAPI Export/Download


marbemac:
Stoplight Documentation and other OSS tools have an option to download/export the OpenAPI document being used to power the documentation. 

<img width="1425" alt="image" src="https://user-images.githubusercontent.com/67381/118528489-4e9f9700-b73a-11eb-9926-ecd8ad3a1e6a.png">

**API Component**

- Similar button design to hosted docs but more mosaic-ish
- Show button it on the overview page only.
- Add a property `hideExport`
- Offer "Original@" and "Bundled References" using our fork of json-schema-ref-parser.
- Ignore Dereferenced because difficulties with circular refs.

**StoplightProject Component**

- Similar button design to hosted docs but more mosaic-ish
- Show button it on the overview page only.
- Add a property `hideExport`
- Look for `links.export_url` in the node response and if it shows up and hideExport=false then show the Export button

**Platform/Ninja**

- Delete old Export OpenAPI buttons.

PR #1652, opened 2021-07-28
feat: openapi export

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/707
Requires https://github.com/stoplightio/platform-internal/pull/7315 to make export button visible in StoplightProject (it's on v2 but not yet in prod)

Adds export/download feature to API and StopolightProject components.

Screenshot 2021-07-28 at 16 04 42
Screenshot 2021-07-28 at 16 04 51

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
@@ -281,7 +281,7 @@ describe('HttpService', () => {
       expect(exportButton).not.toBeInTheDocument();
     });
 
-    it('should not render if hno exportProps are present', () => {
+    it('should not render if no exportProps are present', () => {
       const wrapper = render(
         <Router>
           <MosaicProvider>
--- a/packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
@@ -1,5 +1,6 @@
 import 'jest-enzyme';
 
+import { Provider as MosaicProvider } from '@stoplight/mosaic';
 import {
   IOauth2AuthorizationCodeFlow,
   IOauth2ClientCredentialsFlow,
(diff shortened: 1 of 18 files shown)


marcelltoth:
I'd make the button more neutral, maybe outlined, as in platform. But @mnaumanali94 will decide

mpodlasin (approved):
I agree with @marcelltoth 

However all the rest looks and works good.

mallachari:
`minimal` appearance doesn't look good but `default` does. Switched to:
Screenshot 2021-07-28 at 22 19 34

packages/elements/src/containers/API.tsx
```diff
@@ -90,10 +98,44 @@ export const APIImpl: React.FC<APIProps> = props => {
...
+    const originalDocument = typeof document === 'object' ? JSON.stringify(document, null, 2) : document || '';
+    exportDocument(originalDocument);
+  }, [document, exportDocument]);
+
+  const exportBundledDocument = React.useCallback(() => {
+    const stringifiedDocument = isJsonDocument
+      ? JSON.stringify(bundledDocument, null, 2)
+      : safeStringify(bundledDocument);
+    exportDocument(stringifiedDocument);
+  }, [bundledDocument, isJsonDocument, exportDocument]);
```
marcelltoth, line 128: WDYT about extracting these into a custom hook, to clean up the container?

packages/elements/src/utils/oas/index.ts
```diff
@@ -146,3 +146,12 @@ function findMapMatch(key: string | number, map: ISourceNodeMap[]): ISourceNodeM
     }
   }
 }
+
+export function isJson(value: string) {
+  try {
+    JSON.parse(value);
+  } catch (e) {
+    return false;
+  }
```
marcelltoth, line 155: Lol, I like this one 😄

marcelltoth (dismissed):
Actually, can you write some tests for this? It's a significant piece of functionality (`hideExport`, handling json / yaml), we should cover it.

packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
```diff
@@ -241,4 +242,56 @@ describe('HttpService', () => {
...
+            />
+          </MosaicProvider>
+        </Router>,
+      );
+
+      const exportButton = wrapper.queryByRole('button', { name: 'Export' });
+      expect(exportButton).not.toBeInTheDocument();
+    });
+
+    it('should not render if hno exportProps are present', () => {
```
mpodlasin, line 284: typo

mpodlasin (approved):
👍

cancan101:
Is there any documentation on what exporting "Bundled References" means (as opposed to a exporting "Original")?

mnaumanali94:
@cancan101 While there is no documentation on this original export doesn't resolve references while bundled resolves references and places them in the components section so that there is no duplication. Generally bundled export would be your best bet.

mnaumanali94:
@pamgoodrich probably a good topic to document at some point

PR #1653, opened 2021-07-29
chore: upgrade node in CircleCI

marcelltoth:
Fixes the install problem with the gatsby example hopefully.

Unfortunately Cypress no longer seems to maintain its 12.x images, so I had to move up to 14.x (no 12.20+ image which the dependency requires).


mallachari:
@marcelltoth fixed in https://github.com/stoplightio/elements/pull/1655

Issue #1032, opened 2021-04-08
Examples in Parameter Objects can appear in multiple places


philsturgeon:
## Describe the bug

An example for a parameter could show up in the schema, or in the parameter object (next to schema). Either are equally valid.

```
/params:
  get:
    parameters:
      - name: single-example-good
        description: Valid to its schema
        in: query
        schema: 
          type: string
          enum: [foo, bar]
        example: foo

      - name: single-schema-example-good
        description: Valid to its schema
        in: query
        schema: 
          type: string
          enum: [foo, bar]
          example: foo
```

We currently do not support these sorts of examples, which are showing up in the Digital Ocean example. 

![image](https://user-images.githubusercontent.com/67381/114068707-192fa000-9896-11eb-870e-908a1639818f.png)

## To Reproduce

1. Given the Digital Ocean example https://elements-demo.stoplight.io/?spec=https://raw.githubusercontent.com/digitalocean/openapi/main/specification/DigitalOcean-public.v2.yaml
2. Look at the rate limit headers
3. Should see example, but its not there

## Expected behavior

- ratelimit-limit should have example 5000
- ratelimit-remaining should have example 4816

marcelltoth:
We have worked with `http-spec` before, I think we would be comfortable taking this in any sprint, without external help even (with a non-particularly-high point number). As `http-spec` does not have a dedicated owner, IMO this should be decided based on the expected workload of the teams. If you'd like us to take it, @philsturgeon please place it on our _Product backlog_ where appropriate.

mpodlasin:
This is partially solved in https://github.com/stoplightio/elements/pull/1654

```
      - name: single-example-good
        description: Valid to its schema
        in: query
        schema: 
          type: string
          enum: [foo, bar]
        example: foo
```

will now work.

However this:

```
      - name: single-schema-example-good
        description: Valid to its schema
        in: query
        schema: 
          type: string
          enum: [foo, bar]
          example: foo
```

is a bit trickier. Spec for JSON Schema only mentions "examples" parameter (plural). https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.9.5

There is no mention of "example" (singular).

Even in TypeScript we don't have access to "example" property, since `http-spec` claims we are getting a `JSONSchema7` instance.

So should "example" (singular) parameter in JSON Schema be even supported? 

@marbemac @philsturgeon

philsturgeon:
[...]

PR #1654, opened 2021-07-29
fix: openapi and schema examples validations rendering

mpodlasin:
Closes https://github.com/stoplightio/platform-internal/issues/6720 and https://github.com/stoplightio/elements/issues/1032

Makes sure that `parameter.example`, `parameter.examples` and `parameter.schema.examples` are displayed correctly.

Doesn't display `parameter.schema.example`, since it doesn't exist on the typings nor is mentioned in JSON Schema docs - https://json-schema.org/understanding-json-schema/reference/generic.html

Unifying `parameter.example` and `parameter.examples` was done in `http-spec`, hence it is not visible in tests here. That's what `http-spec` update is for.

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
@@ -62,17 +62,23 @@ export const Parameter: React.FunctionComponent<IParameterProps> = ({ parameter,
   // TODO (JJ): schema.deprecated is used in platform - to be removed once it's updated https://github.com/stoplightio/platform-internal/issues/2267
   const deprecated = get(parameter, 'deprecated') || get(parameter, 'schema.deprecated', false);
 
+  const parameterExamples =
+    parameter.examples?.map(example => {
+      if (isNodeExample(example)) {
+        return example.value;
+      }
+
+      return example.externalValue;
+    }) || [];
+
+  const schemaExamples = parameter.schema?.examples;
+  const schemaExamplesArray = Array.isArray(schemaExamples) ? schemaExamples : [];
+
   const validations = omitBy(
     {
       ...omit(parameter, ['name', 'required', 'deprecated', 'description', 'schema', 'style', 'examples']),
       ...omit(get(parameter, 'schema'), ['description', 'type', 'deprecated']),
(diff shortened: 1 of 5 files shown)


packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
```diff
@@ -62,17 +62,23 @@ export const Parameter: React.FunctionComponent<IParameterProps> = ({ parameter,
...
+    parameter.examples?.map(example => {
+      if (isNodeExample(example)) {
+        return example.value;
+      }
+
+      return example.externalValue;
+    }) || [];
+
+  const schemaExamples = parameter.schema?.examples;
+  const schemaExamplesArray = Array.isArray(schemaExamples) ? schemaExamples : [];
```
marcelltoth, line 75: Why not just `const schemaExamples = parameter.schema?.examples ?? []`, like above? Are we ever interested in these variables by themselves?

marcelltoth (approved):
Looks good! One suggestion, but not to block this.

packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
mpodlasin: Because TS complains, since typings for that JSONSchema7 type are not obeying the spec and are claiming that something else other than array can be there. 

Meanwhile the spec clearly states this must be an array of values. 🤷

marcelltoth: I get it 🤔

philsturgeon:
Please make sure schema.example is supported because it’s an important and regularly used OAS3 keyword.  

https://phil.tech/2020/openapi-examples/

--
Phil Sturgeon
Product @ Stoplight.io

marcelltoth:
I think that's a candidate for simplification by `http-spec`

philsturgeon:
Yes and I believe it already is somewhat. 

--
Phil Sturgeon
Product @ Stoplight.io

mpodlasin:
It already works as needed, see my comments in https://github.com/stoplightio/elements/issues/1032

PR #1655, opened 2021-07-30
fix: gatsby build

mallachari:
Fixes gatsby build by:
- bumping MDV
- using node >=14.13 for cypress workflow - gatsby has a dependency that requires usage of node in version >=12.20 which is missing in cypress docker images. 

Looks like @marcelltoth first approach in https://github.com/stoplightio/elements/pull/1653 was correct.


mpodlasin (approved):
🙏

PR #1667, opened 2021-08-04
fix: jsv padding

### LINKED ISSUE #1562: Json Schema in markdown looks clunky
There is no padding on the right of required:

It can also get awkwardly wide:

marcelltoth:
Resolves #1562 

Also fixes some broken CSS that still has the old classname (`MarkdownViewer` instead of `sl-markdown-viewer`).

### DIFF
--- a/packages/elements-core/src/styles.css
+++ b/packages/elements-core/src/styles.css
@@ -85,6 +85,10 @@
   line-height: 1.5em;
 }
 
+.sl-elements .sl-markdown-viewer .JsonSchemaViewer {
+  padding-right: 10px;
+}
+
 .sl-elements .HttpOperation__gutter {
   --fs-code: 12px;
 }
--- a/packages/elements-core/src/styles.css
+++ b/packages/elements-core/src/styles.css
@@ -69,18 +69,7 @@
 
 /* HttpOperation */
 
-.sl-elements .HttpOperation .JsonSchemaViewer .MarkdownViewer p {
-  font-size: 12px;
-  line-height: 1.5em;
-}
-
-.sl-elements .HttpOperation__Parameters .MarkdownViewer p {
-
-  font-size: 12px;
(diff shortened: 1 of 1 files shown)


packages/elements-core/src/styles.css
```diff
@@ -69,20 +69,13 @@
 
 /* HttpOperation */
 
-.sl-elements .HttpOperation .JsonSchemaViewer .MarkdownViewer p {
+.sl-elements .HttpOperation .JsonSchemaViewer .sl-markdown-viewer p, .sl-elements .HttpOperation__Parameters .sl-markdown-viewer p, .sl-elements .Model .JsonSchemaViewer .sl-markdown-viewer p {
```
marcelltoth, line 72: This is just fixing the 3 rules on the left (they had the old mdv classname), the real "fix" is the `padding-right`

marcelltoth:
> Why do we have to overwrite these anyway? Can we just "fix" the CSS upstream in JSE and Markdown viewer?

It's tricky. You'd never guess where the lightgrey border (see issue) originates from: [yes, *that* library](https://github.com/stoplightio/mosaic/blob/main/libs/core/src/components/Prose/style.css#L347). It's just pure insanity IMO, mosaic isn't supposed to know about JSV's existence... but if you look through that file, actually a significant amount if MDV's styling is implemented there. (Now I get why MDV doesn't need a css file...)

Anyhow, avoiding to go down that rabbit hole, I just added some CSS here that fixes the problem.

packages/elements-core/src/styles.css
wmhilton: Consolidating 3 rules into 1 instead of simply fixing the classnames is the kind of unhelpful refactoring that means I spend 20 minutes resolving a merge conflict 5 days later instead of being able to Just Merge It.

wmhilton (approved):

PR #1668, opened 2021-08-05
chore: don't show duplicate schema description for non-objects

paulatulis:
Addresses #7386

Gets rid of duplicate description for schemas that aren't objects. 
If data.type is an object, should show description. If it's not, don't show description. 

Screen Shot 2021-08-05 at 2 33 21 PM

vs. 

Screen Shot 2021-08-05 at 2 33 48 PM

### DIFF
--- a/packages/elements-core/src/components/Docs/Model/Model.spec.tsx
+++ b/packages/elements-core/src/components/Docs/Model/Model.spec.tsx
@@ -6,6 +6,7 @@ import { Model } from './Model';
 
 const exampleSchema: JSONSchema7 = {
   type: 'object',
+  description: 'example schema description',
   properties: {
     propA: {
       type: 'string',
@@ -20,5 +21,17 @@ describe('Model', () => {
 
     expect(screen.getByRole('heading', { name: /example/i })).toBeInTheDocument();
     expect(container).toHaveTextContent('"propA": "valueA"');
+  });
+
+  it('displays description at top of doc for objects', async () => {
+    const { container } = render(<Model data={exampleSchema} />);
+
+    expect(container).toHaveTextContent(/example schema description/i);
+  });
+
+  it('does not display description at top of doc for non-objects', async () => {
+    const description = screen.queryByRole('textbox');
+
+    expect(description).not.toBeInTheDocument();
   });
 });
--- a/packages/elements-core/src/components/Docs/Model/Model.spec.tsx
+++ b/packages/elements-core/src/components/Docs/Model/Model.spec.tsx
@@ -15,6 +15,11 @@ const exampleSchema: JSONSchema7 = {
   },
 };
 
+const exampleStringSchema: JSONSchema7 = {
+  type: 'string',
+  description: 'example schema description that should only show once :)',
+};
+
 describe('Model', () => {
   it('displays examples', async () => {
     const { container } = render(<Model data={exampleSchema} />);
@@ -24,12 +29,16 @@ describe('Model', () => {
   });
 
   it('displays description at top of doc for objects', async () => {
-    const { container } = render(<Model data={exampleSchema} />);
+    render(<Model data={exampleSchema} />);
+    const description = screen.queryAllByText('example schema description');
+    const textboxDescription = screen.getByRole('textbox');
 
-    expect(container).toHaveTextContent(/example schema description/i);
+    expect(description).toHaveLength(1);
+    expect(textboxDescription).toHaveTextContent('example schema description');
   });
 
   it('does not display description at top of doc for non-objects', async () => {
+    render(<Model data={exampleStringSchema} />);
     const description = screen.queryByRole('textbox');
 
     expect(description).not.toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


mmiask (commented):
Looks good! Let's make sure we add a test case for this, and it's good to go!

paulatulis:
@mmiask forgive me for my lack of unit testing expertise. Let me know if these simple ones are okay!

packages/elements-core/src/components/Docs/Model/Model.spec.tsx
```diff
@@ -21,4 +22,16 @@ describe('Model', () => {
     expect(screen.getByRole('heading', { name: /example/i })).toBeInTheDocument();
     expect(container).toHaveTextContent('"propA": "valueA"');
   });
+
+  it('displays description at top of doc for objects', async () => {
+    const { container } = render(<Model data={exampleSchema} />);
```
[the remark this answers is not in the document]
mmiask: I'd suggest the following:
```ts
    render(<Model data={exampleSchema} />);

    // query by text and check that we display the description once
    const description = screen.queryAllByText('example schema description');

    expect(description).toHaveLength(1);

    // query by role, and check if the description is in the textbox
    
    const textboxDescription = screen.getByRole('textbox');
    
    expect(textboxDescription).toHaveTextContent('example schema description');
```

packages/elements-core/src/components/Docs/Model/Model.spec.tsx
```diff
@@ -21,4 +22,16 @@ describe('Model', () => {
...
+
+    expect(container).toHaveTextContent(/example schema description/i);
+  });
+
+  it('does not display description at top of doc for non-objects', async () => {
```
mmiask, line 32: You're not rendering anything here. First you need create another schema, that is not an object (e.g. string)
```json
{
"type": "string",
"description": "I am a string!"
}
```

Then render Model component using it, and then do the assertion - check that the `textbox` containing `I am a string!` isn't there

I don't think that we need to check if the description is present in schema's row though - it should be covered by JSV's test suite.

packages/elements-core/src/components/Docs/Model/Model.tsx
```diff
@@ -45,7 +45,9 @@ const ModelComponent: React.FC<ModelProps> = ({ data: unresolvedData, className,
 
   const description = (
     <>
-      {data.description && <MarkdownViewer className="sl-mb-6" markdown={data.description} />}
+      {data.description && data.type === 'object' && (
+        <MarkdownViewer className="sl-mb-6" role="textbox" markdown={data.description} />
```
mmiask, line 49: I'm personally not a fan of specifying a `role` since there are ARIA-roles embedded in DOM elements by default, although I understand the need of accesibility here since `MarkdownViewer` is a `div`.

mmiask (commented):
The scope of these tests is precisely what we need 👍 I left some comments on how we could adjust them, but the overall premise is great! Once they're addressed, this is good to go! 🚀

packages/elements-core/src/components/Docs/Model/Model.tsx
mmiask: General thought, not limited to this PR - as a part of tech debt, maybe we should remove this hardcoded roles and implement semantic HTML a bit more

packages/elements-core/src/components/Docs/Model/Model.spec.tsx
paulatulis: AHA yes I was trying to go based off of the original first test but if I just render the screen it works how I want it to. Thank you for pointing this out!

packages/elements-core/src/components/Docs/Model/Model.spec.tsx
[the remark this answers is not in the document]
paulatulis: So helpful, thank you!

packages/elements-core/src/components/Docs/Model/Model.tsx
paulatulis: Is it okay for me to leave this as is for now? I'm happy to update it, just couldn't think of another way to grab only this part.

packages/elements-core/src/components/Docs/Model/Model.spec.tsx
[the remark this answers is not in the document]
paulatulis: Addressed in 4158e71

packages/elements-core/src/components/Docs/Model/Model.spec.tsx
paulatulis: Addressed in 4158e71

paulatulis:
@mmiask okay let's see if I did better this round? Addressed your feedback in 4158e71. Let me know if it looks okay to you!

packages/elements-core/src/components/Docs/Model/Model.tsx
mmiask: Let's leave this here for now, it doesn't affect the PR itself directly

mmiask (approved):
Looks perfect now! Let's merge it once tests pass 🎉

PR #1669, opened 2021-08-06
feat: display multiple servers in service overview page

### LINKED ISSUE #1617: Show all servers in API overview
Currently we just show 1 server with the other being a mock server.

In OAS you can define multiple servers i.e. Staging, Production e.t.c which will can all make sense for a consumer.

We should show all servers here with a divider between Servers from OpenAPI and Mock Server. If a name is not defined just call it `Server`

mpodlasin:
Closes #1617 

Screenshot 2021-08-06 at 15 03 29
Screenshot 2021-08-06 at 15 03 35

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
@@ -53,14 +53,8 @@ describe('HttpService', () => {
   it('generates names for servers without names', () => {
     render(<ServerInfo servers={httpServiceWithUnnamedServers.servers ?? []} />);
 
-    const serverUrl = screen.getByLabelText('Production API');
-    expect(serverUrl).toHaveTextContent('https://api.stoplight.io');
-
     const secondServerUrl = screen.getByLabelText('Server 2');
     expect(secondServerUrl).toHaveTextContent('https://api.staging.stoplight.io');
-
-    const thirdServerUrl = screen.getByLabelText('Integration API');
-    expect(thirdServerUrl).toHaveTextContent('https://api.int.stoplight.io');
 
     const fourthServerUrl = screen.getByLabelText('Server 4');
     expect(fourthServerUrl).toHaveTextContent('https://localhost:4060');
--- a/packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
@@ -14,6 +14,7 @@ import { BrowserRouter as Router } from 'react-router-dom';
 
 import { apiKey, oauth } from '../../../__fixtures__/security-schemes';
 import httpService from '../../../__fixtures__/services/petstore';
+import { httpServiceWithUnnamedServers } from '../../../__fixtures__/services/with-unnamed-servers';
 import { httpServiceWithUrlVariables } from '../../../__fixtures__/services/with-url-variables';
 import { httpServiceWithoutOrigin } from '../../../__fixtures__/services/without-origin';
 import { HttpService } from './index';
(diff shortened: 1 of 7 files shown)


packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
```diff
@@ -31,36 +32,60 @@ describe('HttpService', () => {
...
+    expect(fourthServerUrl).toHaveTextContent('https://localhost:4060');
+
+    expect(screen.queryByLabelText('Mock Server')).not.toBeInTheDocument();
+  });
+
+  it('generates names for servers without names', () => {
+    render(<ServerInfo servers={httpServiceWithUnnamedServers.servers ?? []} />);
+
+    const serverUrl = screen.getByLabelText('Production API');
+    expect(serverUrl).toHaveTextContent('https://api.stoplight.io');
```
marcelltoth, line 57: I'd not assert these here again. It's already tested above, and in general one test should have one reason to fail. Just test what the test title says.

packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
```diff
@@ -31,36 +32,60 @@ describe('HttpService', () => {
...
+
+    const secondServerUrl = screen.getByLabelText('Server 2');
+    expect(secondServerUrl).toHaveTextContent('https://api.staging.stoplight.io');
```
marcelltoth, line 60: Not that it matters really, but just noticed that the assertions are a bit backwards here. Really what you're testing for is whether the server at `https://api.staging.stoplight.io` has the name *Server 2* or not, right? Your code is written the other way around though. I don't mind it too much but consider it.

packages/elements-core/src/components/Docs/HttpService/ServerInfo.tsx
```diff
@@ -11,14 +11,23 @@ interface ServerInfoProps {
   mockUrl?: string;
 }
 
+const getServersToDisplay = (originalServers: IServer[]): IServer[] => {
```
marcelltoth, line 14: I like that you extracted this here, neat.

packages/elements-core/src/components/Docs/HttpService/ServerInfo.tsx
```diff
@@ -27,24 +36,25 @@ export const ServerInfo: React.FC<ServerInfoProps> = ({ servers, mockUrl }) => {
...
-              <Box whitespace="nowrap">
-                {showMocking && (
-                  <Text pr={2} fontWeight="bold">
-                    Production:
-                  </Text>
-                )}
-                <Text aria-label="production-server">{productionUrl}</Text>
+          <Panel.Content w="full" className="sl-flex sl-flex-col">
+            {serversToDisplay.map(({ name, url }) => (
+              <Box whitespace="nowrap" mb={1} key={url}>
```
marcelltoth, line 41: Suuuuuure about that `key`? I know I'm extra cautious around keys, but IMO anything that's user-driven can (and will) have duplications. You don't even need a weird use-case for it: don't forget that this is here driving the preview in studio, too. It's very likely that someone looking to add another server will start by duplicating an existing one. At that point you are in undefined behavior mode. 

I propose we err on the safe side and use indexes for most things, unless there's a native ID (I mean like a postrgres key, something that's guaranteed by our API to be unique).

marcelltoth (commented):
image

This doesn't look right, why do I see `To-Dos` twice [here](https://deploy-preview-1669--stoplight-elements-demo.netlify.app/?utm_source=github&utm_campaign=bot_dp#/)?

Weird as it's not even listed a single time here (from the Export): 
image

mpodlasin:
> image
> 
> This doesn't look right, why do I see `To-Dos` twice [here](https://deploy-preview-1669--stoplight-elements-demo.netlify.app/?utm_source=github&utm_campaign=bot_dp#/)?
> 
> Weird as it's not even listed a single time here (from the Export):
> image

Aaaah, you are right. I used wrong property (`http-spec`'s `name` instead of `description`). 🤷

packages/elements-core/src/components/Docs/HttpService/HttpService.spec.tsx
mpodlasin: Idk if I agree. For me the test simulates user behaviour of "let me check what is a *Production* url". So first I am looking for label and then I am checking the url. 🤷

mnaumanali94 (approved):

wmhilton (approved):

Issue #1618, opened 2021-07-20
Increase padding for markdown articles


mnaumanali94:
Padding around markdown docs is very less compared to the actual designs:

![Screenshot 2021-07-20 at 1.23.22 PM.png](https://images.zenhubusercontent.com/5ea9a7f489044245dbd23d46/57a62105-7037-449d-ae6a-12ee5c0ec5aa)

This really boils down to max width, as people can't properly read very wide lines. Align Docs with the Figma designs:

https://www.figma.com/file/PzxLMVm3NmDuKX1y3UR4x9/Workspace?node-id=1330%3A40569

PR #1670, opened 2021-08-06
Fix: increase padding in docs

mmiask:
A part of: https://github.com/stoplightio/elements/issues/1618

This adds some class names to the `Docs/Article` component and later applies custom styling to this component in `elements-dev-portal`. I decided not to modify the `Article` component directly in order to make it more customizable for other usages.

**Before**

https://user-images.githubusercontent.com/58433203/128506785-cd574528-1501-4fc8-9ac8-d723a9c66efe.mov

**After**

https://user-images.githubusercontent.com/58433203/128506554-fefa9215-f744-45cc-948f-971cb5d27a75.mov

### DIFF
--- a/packages/elements-dev-portal/src/styles.css
+++ b/packages/elements-dev-portal/src/styles.css
@@ -1,12 +1,13 @@
 @import "../../elements-core/src/styles.css";
 
 .sl-elements-article {
-    justify-content: space-evenly;
-    padding-left: 60px;
-    padding-right: 60px;
-  }
+  justify-content: space-evenly;
+}
   
-  .sl-elements-article-content {
-    max-width: 800px;
-  }
-  
+.sl-elements-article-content {
+  max-width: 700px;
+}
+
+.ArticleHeadings {
+  max-width: 300px;
+}
--- a/packages/elements-dev-portal/src/styles.css
+++ b/packages/elements-dev-portal/src/styles.css
@@ -1,1 +1,12 @@
 @import "../../elements-core/src/styles.css";
+
+.sl-elements-article {
+    justify-content: space-evenly;
+    padding-left: 60px;
+    padding-right: 60px;
+  }
+  
+  .sl-elements-article-content {
+    max-width: 800px;
+  }
+  
--- a/packages/elements-core/src/components/Docs/Article/index.tsx
+++ b/packages/elements-core/src/components/Docs/Article/index.tsx
@@ -17,8 +17,8 @@ const ArticleComponent = React.memo<ArticleProps>(({ data }) => {
   if (tree === null) return null;
 
   return (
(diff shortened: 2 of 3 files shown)


packages/elements-dev-portal/src/styles.css
```diff
@@ -1 +1,12 @@
 @import "../../elements-core/src/styles.css";
+
+.sl-elements-article {
+    justify-content: space-evenly;
```
mpodlasin, line 4: Idk. I find it a bit weird.

First of all, we should keep CSS to bare minimum and use it only if mosaic's styling solution really can't solve our problems.

On top of that, this will only affect dev-portal. Meanwhile I think the biggest importance was for that to look good in platform. But it doesn't matter - for me the layout should be the same in platform, dev-portal and elements.

mnaumanali94:
Screenshot 2021-08-06 at 7 12 23 PM


Thoughts on just updating the max-width to 600 to solve this problem? 🤔

mmiask:
> Screenshot 2021-08-06 at 7 12 23 PM
> 
> Thoughts on just updating the max-width to 600 to solve this problem? 🤔

@mnaumanali94 I personally found this width to be too narrow, but if that's the suggested value let's go with it

mnaumanali94:
Can you share a screenshot of how it looks? 🙏🏼

packages/elements-dev-portal/src/styles.css
mmiask: Ok, so there are few things here that need addressing:

> First of all, we should keep CSS to bare minimum and use it only if mosaic's styling solution really can't solve our problems.

Most definitely. Mosaic applies to MDV though, and I feel that enforcing a specific width or padding there isn't a perfect solution. It would reduce the flexibility of customization of MDV outside of Elements. Unless it's directly required by Design or Product person, I think we should aim for Elements styling to be as customizable as possible.

> On top of that, this will only affect dev-portal. Meanwhile I think the biggest importance was for that to look good in platform. But it doesn't matter - for me the layout should be the same in platform, dev-portal and elements.

The thing is that Platform uses `NodeContent`. If we change this for `dev-portal` it will be present in Platform after Elements bump. Even if I'd want to style this directly within Platfrom, I will still need specific selectors for these containers.

One thing this PR will do though is that it will enforce styling for `NodeContent` - ideally we probably want to have a custom styling within `dev-portal` and `Platform` but not for `NodeContent` directly - so that people can still import that component and style it as they want

mmiask:
> Can you share a screenshot of how it looks? 🙏🏼

The above was discussed on a call, I'll change the max-width to `700px` for article content and `300px` for `ArticleHeadings`

mpodlasin (approved):
I also find it a bit narrow-ish, but code looks good. 👍

Issue #1588, opened 2021-07-09
Docs: Disambiguate Sceurity Schemes with same types


philsturgeon:
## Use Story

As somebody describing an API with multiple security schemes using the same type, whether together or separately, I would expect them to be disambiguated so its clear to the user what badges are what. 

<img width="1202" alt="image" src="https://user-images.githubusercontent.com/67381/125072584-102fb400-e0bb-11eb-87a9-99efbfd8f4dc.png">

## Is your feature request related to a problem?

Calendly and Plaid have the issue where there are multiple security schemes that need disambiguation. 

Lets look at Plaid's example reduced to nothing.

```
openapi: 3.0.0
info:
  contact:
    name: Plaid Developer Team
    url: 'https://plaid.com'
  description: 'The Plaid REST API. Please see https://plaid.com/docs/api for more details.'
  termsOfService: 'https://plaid.com/legal/'
  title: The Plaid API
  version: 1.0.0
security:
  - clientId: []
    plaidVersion: []
    secret: []
tags:
  - description: The Plaid API
    name: plaid
paths:
  /accounts/balance/get:
    post:
      operationId: accountsBalanceGet
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: string
 
components:
  securitySchemes:
    clientId:
      in: header
      name: PLAID-CLIENT-ID
      type: apiKey
    plaidVersion:
      in: header
      name: Plaid-Version
      type: apiKey
    secret:
      in: header
      name: PLAID-SECRET
      type: apiKey
```

## Describe the solution you'd like

Lets update the badge to contain the key that securityScheme was defined by (e.g.: `API Key (clientId)`, `API Key (plaidVersion)`, `API Key (secret)`). 

Also add this info on the overview page and security selector in try-it.

## Additional context

https://elements-demo.stoplight.io/?spec=https://api.apis.guru/v2/specs/plaid.com/2020-09-14_1.8.0/

philsturgeon:
Link to feature: https://stoplightio.productboard.com/feature-board/planning/features/8608198

mnaumanali94:
Don't show the scopes if we do.

PR #1671, opened 2021-08-06
feat: same type security schemes

mallachari:
Fixes: https://github.com/stoplightio/elements/issues/1588

Shows security keys for auth schemes that occur duplicated.

Unique schemes:
Screenshot 2021-08-06 at 12 30 54
Screenshot 2021-08-06 at 12 31 03

Duplicated schemes:
Screenshot 2021-08-06 at 12 40 22
Screenshot 2021-08-06 at 12 43 22

### DIFF
--- a/packages/elements-core/src/components/TryIt/Auth/Auth.tsx
+++ b/packages/elements-core/src/components/TryIt/Auth/Auth.tsx
@@ -3,7 +3,7 @@ import { HttpSecurityScheme } from '@stoplight/types';
 import { flatten } from 'lodash';
 import * as React from 'react';
 
-import { getReadableSecurityName } from '../../../utils/oas/security';
+import { getReadableSecurityName, shouldIncludeKey } from '../../../utils/oas/security';
 import { APIKeyAuth } from './APIKeyAuth';
 import { HttpSecuritySchemeWithValues } from './authentication-utils';
 import { BasicAuth } from './BasicAuth';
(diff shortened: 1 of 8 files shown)


mnaumanali94:
Beautiful!

mnaumanali94 (approved):

packages/elements-core/src/components/TryIt/Auth/Auth.tsx
```diff
@@ -41,7 +41,7 @@ export const TryItAuth: React.FC<TryItAuthProps> = ({ operationSecurityScheme: o
         title: 'Security Schemes',
         children: filteredSecurityItems.map(auth => ({
           id: `security-scheme-${auth.key}`,
-          title: getReadableSecurityName(auth),
+          title: getReadableSecurityName(auth, shouldIncludeKey(filteredSecurityItems, auth.type)),
```
mmiask, line 44: What do you think about changing the menu display name to include keys as well instead of only `API Key`? (I guess that would be line 26 in this file
Screen Shot 2021-08-10 at 12 59 53 PM

mmiask (approved):
Code looks good, works great as well 👍  I left two questions, but they shouldn't block anything

packages/elements-core/src/utils/oas/security.ts
```diff
@@ -5,33 +5,31 @@ import {
...
-      return 'Mutual TLS';
+      name = 'Mutual TLS';
+      break;
   }
-}
 
-const getOauthScopeMapper = (securityScheme: IOauth2SecurityScheme) => (flow: string) => {
-  if (!['implicit', 'password', 'clientCredentials', 'authorizationCode'].includes(flow)) return [];
-  return keys(securityScheme.flows[flow]?.scopes);
-};
```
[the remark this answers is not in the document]
mallachari: We're removing scopes from the badges. We don't need them there.

packages/elements-core/src/components/TryIt/Auth/Auth.tsx
mallachari: I intentionally didn't add it there as I thought it might make it messy and that it may be batter to kepp button with generic name. But we could do that if that's something worth to do.

PR #1672, opened 2021-08-06
fix: all elements menus

mallachari:
For some reason, the new mosaic version breaks the usage of all Menu components. 
Reverting to 1.1.5 to unblock other tasks.


marcelltoth:
Still failing unfortunately :\

PR #1675, opened 2021-08-10
Update CODEOWNERS

wmhilton:


mallachari (approved):
Maybe worth adding Nauman too?

PR #1677, opened 2021-08-10
chore: weekly maintenance PR should always run on latest 'main' commit

wmhilton:
Normally this doesn't matter, but I wanted to be able to "re-run" failed lockfile maintenance jobs, in which case it matters that you're running on the latest commit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,6 +92,7 @@ jobs:
       - run:
           command: |
             git checkout main
+            git pull --ff-only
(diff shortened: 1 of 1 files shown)


paulatulis:
@wmhilton I have a dumb question and that is: what is a weekly maintenance PR?

mpodlasin:
@paulatulis We've made a simple script that refreshes yarn.lock. It creates PR like this once a week https://github.com/stoplightio/elements/pull/1691

It was needed, because when users install elements as a library, our lock gets ignored. Hence we need to make sure that elements works with the newest available dependencies.

mpodlasin (approved):
👍

PR #1684, opened 2021-08-12
feat: In MarkdownViewer TryIt, parse path parameters based on provided url

mmiask:
A part of: https://github.com/stoplightio/platform-internal/issues/7388

This PR adds a posibility to enter the value of path parameters, which are now available in inline TryIt/RequestMaker (in MarkdownViewer) based on the names extracted from url e.g.

```
$baseUrl/hello/{param1}/welcome/{path2}
```

Will allow to pass values for `param1` and `param2`

All of this is utilizing and modifying existing `parseHttpRequest` function which expects data of type `PartialHttpRequest` - path parameters are not specified there per se, as they normaly are in OAS document (where you need path param both in url and a corresponding path param object).

There's a limitation here that we only allow to pass values, but do not show examples, expected data format (`string`, `number`, etc.) or prepopulate the `TryIt` with a path param value based on the `json` file passed to it.

Please let me know if you think I should update the `parseHttpRequest` to accept more robust data than `PartialHttpRequest` or is this good enough.

### **Before**

https://user-images.githubusercontent.com/58433203/129169151-35c9445b-2085-4fc2-b869-2ed20ca703f5.mov

### **After**

https://user-images.githubusercontent.com/58433203/129169196-cb192160-6cc9-4757-9b47-a6f766c1e5e0.mov

### DIFF
--- a/packages/elements-core/src/components/MarkdownViewer/CustomComponents/CodeComponent.spec.ts
+++ b/packages/elements-core/src/components/MarkdownViewer/CustomComponents/CodeComponent.spec.ts
@@ -54,7 +54,7 @@ describe('parseHttpRequest', () => {
   it('parses HttpRequest parameters', () => {
     const request: IHttpRequest = {
       method: 'post',
-      url: '/todos',
+      url: '/todos/{id}',
       baseUrl: 'http://test',
       query: {
         limit: ['10'],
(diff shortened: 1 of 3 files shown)


packages/elements-core/src/components/MarkdownViewer/CustomComponents/CodeComponent.spec.ts
```diff
@@ -70,11 +70,12 @@ describe('parseHttpRequest', () => {
     expect(httpOperation).toMatchObject({
       id: '?http-operation-id?',
       method: 'post',
-      path: '/todos',
+      path: '/todos/{id}',
       servers: [{ url: 'http://test' }],
       request: {
         query: [{ name: 'limit', style: HttpParamStyles.Form, schema: { default: '10' } }],
         headers: [{ name: 'apikey', style: HttpParamStyles.Simple, schema: { default: '123' } }],
+        path: [{ name: 'id', style: HttpParamStyles.Simple, required: true }],
```
mpodlasin, line 78: hmm, maybe this is `type: string`, since it has to be in the url? 🤔

mpodlasin (approved):
Works and looks good, just one comment to consider.

mpodlasin:
Do we have an issue created for large labels in TryIt on markdown? 🤔

packages/elements-core/src/components/MarkdownViewer/CustomComponents/CodeComponent.spec.ts
mmiask: Well, technically yes - but if we display `string` all the time, and there's a path param called `{id}` and it should be a `number`, we won't show that

PR #1687, opened 2021-08-13
feat: don't show big response examples (>500 lines) by default

mmiask:
Big response examples vs our current implementation

asd

### The solution:

After some discussion on Slack about large response examples causing long loading times/page hanging, it was agreed upon limiting the size of response examples being displayed by default. This PR does that (limit of 500 lines) and adds a loading button if users want to display that response example anyway.

**Before**
Screen Shot 2021-08-13 at 4 46 36 PM

**After**
Screen Shot 2021-08-13 at 4 46 41 PM

**Screenshots**

https://user-images.githubusercontent.com/58433203/129375699-b7bc014c-e0e2-404c-86bc-fe1a28acbeb0.mov


packages/elements-core/src/hoc/withQueryClientProvider.tsx
```diff
@@ -7,7 +7,8 @@ const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       retry: false,
-      staleTime: 15 * 1000,
+      refetchOnWindowFocus: false,
+      staleTime: Infinity,
```
mpodlasin, line 11: this will break editing.

When users add new article for example in studio, and go back to platform, they will never see it.

packages/elements-core/src/components/ResponseExamples/ResponseExamples.tsx
```diff
@@ -40,18 +40,34 @@ export const ResponseExamples = ({ httpOperation, responseMediaType, responseSta
...
           maxHeight="400px"
           language="json"
-          value={responseExample}
+          value={example}
           showLineNumbers
         />
+      );
+
+    return (
+      <Box p="auto" onClick={() => setShow(true)}>
```
mpodlasin, line 60: If it's clickable, we should make some kind of button out of it, also for accessibility reasons.

packages/elements-core/src/hoc/withQueryClientProvider.tsx
mmiask: This was just an experiment. Changing it back!

packages/elements-core/src/components/ResponseExamples/ResponseExamples.spec.tsx
```diff
@@ -74,4 +76,31 @@ describe('Response Examples', () => {
     expect(container).not.toHaveTextContent('writeOnlyParamter');
     expect(container).toHaveTextContent('someOtherParameter');
   });
+
+  it('does not show examples >500 lines by default', async () => {
```
mpodlasin, line 80: <3

mpodlasin (approved):
Works nicely. 👍

noor-ahmad:
Looks good Michael! 👍

Small little nitpick, could we change  `...Loading` to be `Loading...`

Maybe later on we can include a circular loading indicator above `Loading` to give it some animation! But no rush, something we can tackle in a new ticket!

mnaumanali94 (approved):

packages/elements-core/src/components/ResponseExamples/ResponseExamples.tsx
```diff
@@ -44,14 +47,34 @@ export const ResponseExamples = ({ httpOperation, responseMediaType, responseSta
...
+              aria-label="load-example"
+              onPress={() => {
+                setLoading(true);
+                setTimeout(() => setShow(true), 50);
+              }}
+              appearance="minimal"
+              loading={loading}
+              disabled={loading}
+            >
+              {loading ? '...Loading' : 'Load examples'}
```
wmhilton, line 71:
```suggestion
              {loading ? 'Loading...' : 'Load examples'}
```

PR #1692, opened 2021-08-16
Feat: Stoplight Markdown support in Request Samples aka Codegen

mmiask:
Inspired by a conversation on Slack, I created a small PoC of a Stoplight Markdown code generator.

Only supports `application/json` for now and misses the ability to distinguish `baseurl` and `url`/`path` but it might be a starting point for future improvements.


packages/elements-core/src/components/RequestSamples/convertRequestToSample.ts
```diff
@@ -1,7 +1,27 @@
...
+
 export const convertRequestToSample = (language: string, library: string | undefined, request: HarFormatRequest) => {
+  // Stoplight Markdown support
+  if (language === 'StoplightMarkdown') {
+    let markdown;
+    if (library === 'yaml') {
+      markdown =
+        '```yaml http\n' +
+        `${safeStringifyYaml(toPartialHttpRequest(request), { indent: 2, noRefs: true, skipInvalid: true })}` +
+        '\n```';
```
mmiask, line 17: This can be basically anything, like:
```tsx
if (library) {
  return;
}
```

Just so we don't break the `convertRequestTosample`

packages/elements-core/src/components/RequestSamples/convertRequestToSample.ts
```diff
@@ -10,3 +30,16 @@ export const convertRequestToSample = (language: string, library: string | undef
     return null;
   }
 };
+
+function toPartialHttpRequest(request: HarFormatRequest): PartialHttpRequest {
+  const query = request.queryString.reduce((obj, item) => Object.assign(obj, { [item.name]: [item.value] }), {});
```
mpodlasin, line 35: is that just copying objects? 🤔

mpodlasin:
It works great!!! I vote to merge it.

packages/elements-core/src/components/RequestSamples/convertRequestToSample.ts
mmiask: Almost. Both of these types are pretty similar e.g. 

Screen Shot 2021-08-16 at 5 00 35 PM

vs

Screen Shot 2021-08-16 at 5 19 08 PM

mmiask: `PartialHttpRequest` interface which is the base of Stoplight Markdown syntax used in embedded `TryIt` is pretty loose. I guess we could add some guards here if needed

mnaumanali94:
Path parameters seems to be ignored. Is that what you mean by not distinguishing between base url and path?

mmiask:
> Path parameters seems to be ignored. Is that what you mean by not distinguishing between base url and path?

@mnaumanali94 

`IHttpRequest` type which is used in Markdown Viewer inlined `TryIt` looks as following (CodeComponent.tsx):

```tsx
export interface IHttpRequest<T = any> {
    method: HttpMethod;
    /** Can be relative or absolute. If relative, `baseUrl` must also be set. */
    url: string;
    /** `baseUrl` will be prepended to `url` unless `url` is absolute. */
    baseUrl: string;
    headers: HttpNameValue;
    query: HttpNameValues;
    body?: T;
}
```
This type is expected in Markdown Viewer and being parsed to an actual `IHttpOperation` format which is accepted by `TryIt`:

```tsx
export function parseHttpRequest(data: PartialHttpRequest): IHttpOperation {
```

So when generating code of Stoplight Markdown format (which is basically of `PartialHttpRequest` /`IHttpRequest` type), we don’t expect path at all. We generate path parameters based on the url:

```tsx
export function parseHttpRequest(data: PartialHttpRequest): IHttpOperation {
[...]
82 const pathParam = data.url.match(/[^{\}]+(?=})/g);
[...]
99 path: pathParam?.map(name => ({
        name,
        style: HttpParamStyles.Simple,
        required: true,
      })),
```

So in short - having the types we have right now for embedded `TryIt` (TryIt/Request Maker in Markdown), we don’t expect users to specify path parameters as separate properties, but rather generate them based on the provided url

mmiask:
Oooh, I see now that path parameter are not being enclosed in curly braces. This might be tricky, as HAR request format doesn't have an information about path parameters

Issue #1662, opened 2021-08-03
multiple examples for schemas don't show up in platform


mnaumanali94:
Currently we don't show multiple examples for schemas

![Screenshot 2021-08-03 at 5.37.53 AM.png](https://images.zenhubusercontent.com/5ea9a7f489044245dbd23d46/71aeb768-67cd-4838-bb25-8133fadb8606)

Lets convert the above to `Examples` and a dropdown like below if there are multiple examples:

![Screenshot 2021-08-03 at 5.37.49 AM.png](https://images.zenhubusercontent.com/5ea9a7f489044245dbd23d46/77d18d3a-7ed1-4c5e-aefb-b2e5eece5f19)

mallachari:
Examples don't show in platform because they are missing. The example that is visible in Docs doesn't come from the document - it is generated. Parsed document (HttpOperation) lacks `examples` property in media object. That's probably the result of incorrectly analyzing document as it is missing in node_snapshot as well. Transform operation in `http-spec` seems fine as it does the job in Elements API component for the same document.

domagojk:
Hm, I haven't find any issues in node_snapshot table (example was there), and I got it working by doing this https://github.com/stoplightio/elements/pull/1693
Am i missing something?

PR #1693, opened 2021-08-16
fix: support for multiple model examples

domagojk:
For #1662 

I've used a similar approach as in `elements-core/src/components/ResponseExamples/ResponseExamples.tsx` where examples length is checked and if there are multiple ones defined, it uses `Select` component


packages/elements-core/src/components/Docs/Model/Model.spec.tsx
```diff
@@ -30,6 +30,24 @@ describe('Model', () => {
     expect(container).toHaveTextContent('"propA": "valueA"');
   });
 
+  it('uses examples defined in the schema', async () => {
```
mpodlasin, line 33: would be awesome to have a test for switching the examples as well.

mpodlasin (approved):

Issue #687, opened 2020-11-02
API Overview: Additional Information Links


marbemac:
As a docs consumer, I can see various pieces of additional information, including license, contact information, TOS, and external docs link on the API overview page.

<p></p>

- For [contact](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#contactObject), we should display the "Contact ${info.contact.name}" and the link will be info.contact.url if it exists, or mailto: info.contact.email.

- For [license](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#license-object), we should display the name of the license, and link to license

  - If `license.url` is there then show it.

  - OAS3.1 introduced `license.identifier` which can be used to get a URL: <https://www.npmjs.com/package/spdx-license-list>

  - If both are there (they should not be but thats up to spectral to complain about) then show the URL 

- For [terms](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#info-object-example), we should display "Terms of Service" as a link to the info.termsOfService url.

- **External Docs** are handled separately (#718) and **are out of scope here**

![](https://pb-files.s3.amazonaws.com/production/attachments/8de5469dfb21418a38bbf5cff1a8b4d7b3b19699f89406baa17de5e8614cb75f/image.png)

<p></p>

PR #1696, opened 2021-08-16
chore: add additional info to docs overview

paulatulis:
Addresses #687 by adding Addditional Information section to docs API overview page. 
Screen Shot 2021-08-16 at 11 34 57 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,9 +13,7 @@
     "demo",
     "packages/*"
   ],
-  "dependencies": {
-    "spdx-license-list": "^6.4.0"
-  },
+  "dependencies": {},
   "devDependencies": {
     "@babel/core": "7.12.17",
     "@babel/plugin-syntax-object-rest-spread": "7.8.3",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,9 @@
     "demo",
     "packages/*"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "spdx-license-list": "^6.4.0"
+  },
   "devDependencies": {
     "@babel/core": "7.12.17",
     "@babel/plugin-syntax-object-rest-spread": "7.8.3",
@@ -23,16 +25,16 @@
     "@storybook/addon-postcss": "2.0.0",
     "@storybook/addon-toolbars": "6.3.5",
     "@storybook/addon-viewport": "6.3.5",
-    "@storybook/react": "6.3.5",
-    "@storybook/theming": "6.3.5",
     "@storybook/builder-webpack5": "6.3.5",
     "@storybook/manager-webpack5": "6.3.5",
+    "@storybook/react": "6.3.5",
+    "@storybook/theming": "6.3.5",
     "@testing-library/cypress": "^7.0.3",
     "@typescript-eslint/eslint-plugin": "4.28.4",
     "@typescript-eslint/parser": "4.28.4",
     "concurrently": "6.1.0",
-    "cssnano": "5.0.6",
     "css-loader": "6.2.0",
+    "cssnano": "5.0.6",
     "cypress": "^6.3.0",
     "eslint": "7.31.0",
     "eslint-plugin-import": "2.23.4",
--- a/packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
@@ -1,10 +1,9 @@
 import { Panel } from '@stoplight/mosaic';
 import { IHttpService } from '@stoplight/types';
 import React from 'react';
+import spdxLicenseList from 'spdx-license-list';
 
 import { MarkdownViewer } from '../../MarkdownViewer';
-
-const spdxLicenseList = require('spdx-license-list');
 
 interface AdditionalInfoProps {
   termsOfService?: string;
@@ -24,7 +23,7 @@ export const AdditionalInfo: React.FC<AdditionalInfoProps> = ({ termsOfService, 
   const licenseUrl =
     license?.url || (spdxLicenseList[`${license?.identifier}`] && spdxLicenseList[`${license?.identifier}`].url);
 
-  const licenseLink = license?.name && licenseUrl ? `[License: ${license.name}](${licenseUrl})` : '';
+  const licenseLink = license?.name && licenseUrl ? `[${license.name} License](${licenseUrl})` : '';
   const tosLink = termsOfService ? `[Terms of Service](${termsOfService})` : '';
   return (
     <>
--- a/packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
+++ b/packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
@@ -24,7 +24,7 @@ export const AdditionalInfo: React.FC<AdditionalInfoProps> = ({ termsOfService, 
   const licenseUrl =
     license?.url || (spdxLicenseList[`${license?.identifier}`] && spdxLicenseList[`${license?.identifier}`].url);
 
-  const licenseLink = license?.name && licenseUrl ? `[${license.name}](${licenseUrl})` : '';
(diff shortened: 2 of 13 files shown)


philsturgeon:
As some people will not know what "MIT" means, can we label things as `License:` or something?

paulatulis:
@philsturgeon yeah! Do you think I should change the other lines to be consistent with that format or does this look okay?

Screen Shot 2021-08-17 at 11 16 20 AM

packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
```diff
@@ -0,0 +1,45 @@
+import { Panel } from '@stoplight/mosaic';
+import { IHttpService } from '@stoplight/types';
+import React from 'react';
+
+import { MarkdownViewer } from '../../MarkdownViewer';
+
+const spdxLicenseList = require('spdx-license-list');
```
wmhilton, line 7: Did you try `import` instead of `require`? Just seems odd to have a `require` in here...

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
```diff
@@ -18,7 +19,7 @@ const HttpServiceComponent = React.memo<HttpServiceProps>(({ data, location = {}
   const { search, pathname } = location;
   const mocking = React.useContext(MockingContext);
   const query = new URLSearchParams(search);
-
+  console.log(data);
```
wmhilton, line 22:
```suggestion

```

package.json
```diff
@@ -13,7 +13,9 @@
     "demo",
     "packages/*"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "spdx-license-list": "^6.4.0"
+  },
```
wmhilton, line 18: I think this dependency needs to be in the `packages/elements-core/package.json` in order for the final package to work properly. The top-level `package.json` should only contain `devDependencies` needed for people who clone the elements repo.

mnaumanali94:
> @philsturgeon yeah! Do you think I should change the other lines to be consistent with that format or does this look okay?
> 
> Screen Shot 2021-08-17 at 11 16 20 AM

@paulatulis Lets say MIT License, Apache License e.t.c.

packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+      ? `[Contact ${contact.name}](${contact.url})`
+      : contact?.email
+      ? `[Contact ${contact.name || contact.email}](mailto:${contact.email})`
+      : '';
+
+  //use spdx to look up url for license identifier if available
+  const licenseUrl =
+    license?.url || (spdxLicenseList[`${license?.identifier}`] && spdxLicenseList[`${license?.identifier}`].url);
+
+  const licenseLink = license?.name && licenseUrl ? `[License: ${license.name}](${licenseUrl})` : '';
```
wmhilton, line 27: I don't think this works if `license.identifier` is present but not `license.name`. You could look up the `name` from `spdx-license-list` the same way you look up the `url`.

wmhilton: I take that back. I see `name` is "REQUIRED" by the OpenAPI 3.1 spec. So we should probably force customers to use it correctly 😆

wmhilton: p.s. The reason I think it's a bad design decision on OpenAPI's part is it means this is valid:

```json
{
  "name": "BSD 3-Clause",
  "identifier": "Apache-2.0"
}
```

and heavens knows what that means, except fun for lawyers. Particularly in our docs rendering, unless they click on the link, they wouldn't even notice the URL points to a different license than the name field.

paulatulis: @wmhilton maybe I am confusing myself, but I think that I intentionally made it so that it won't show anything if there's not a name AND a link (or a name AND an identifier to look up the link) based on the fact that Spectral will yell at someone if they don't include the name, so the onus is on the user to add one. 

Also, agree with why you think it's bad design! When I was testing it out, I was using random names with real identifiers and thought that seemed weird :D 

Do you think I should alter it?

packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
paulatulis: addressed with abc8aac

package.json
paulatulis: Hopefully addressed with abc8aac

paulatulis:
> @paulatulis Lets say MIT License, Apache License e.t.c.

@mnaumanali94  addressed with abc8aac

packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
wmhilton: > Do you think I should alter it?

Nope, let's leave it as is. I may disagree with the standard but if everyone writes code based on what they _wish_ the standard said, you end up undermining the standard. (And you get really awkward behavior like HTML parsers where some badly written HTML works in some browsers but is broke in others.) The standard says `name` is required, Spectral will yell if they don't include the name, Elements should treat name as required too then.

packages/elements-dev-portal/src/version.ts
```diff
@@ -1,2 +1,2 @@
 // auto-updated during build
-export const appVersion = '1.0.11';
+export const appVersion = '1.1.0';
```
wmhilton, line 2: ... lol. Looks right though!
image

paulatulis: Oh goodness this snuck in and I don't know exactly how. Should I revert?

wmhilton (changes requested):
Ooh. OK. I have one more suggestion 😆 

Instead of adding `spds-license-list` dependency, let's generate the URL automatically using just the identifier: `https://spdx.org/licenses/${identifier}.html` (for example, https://spdx.org/licenses/MIT.html)

That way we don't add [12KB](https://bundlephobia.com/package/spdx-license-list@6.4.0) to the size of elements just to be able to add license links.

Edit: I see `spdx-license-list` was Marc's suggestion, but I still think using `spdx.org` instead of adding a JSON listing every license ever is better 😛

packages/elements-dev-portal/src/version.ts
wmhilton: No, lets leave it. It probably should have happened as part of an earlier commit. The comment above makes me think it's some automated build step.

paulatulis:
> Ooh. OK. I have one more suggestion 😆
> 
> Instead of adding `spds-license-list` dependency, let's generate the URL automatically using just the identifier: `https://spdx.org/licenses/${identifier}.html` (for example, https://spdx.org/licenses/MIT.html)
> 
> That way we don't add [12KB](https://bundlephobia.com/package/spdx-license-list@6.4.0) to the size of elements just to be able to add license links.
> 
> Edit: I see `spdx-license-list` was Marc's suggestion, but I still think using `spdx.org` is better than adding a JSON file listing every license ever 😛

I feel like he might agree and I like this suggestion! Changed with 45c97bf

mnaumanali94 (approved):

wmhilton (approved):
Yay!

wmhilton:
> Thoughts?

We discussed it in Slack and I declared that adding a 404 check is not worth it. Some other reasons:

- if you do a 404 check, now you're downloading a biggish file again. That would increase page loading time.
- 99% of the time, nobody looking at the page is going to click the link. So 99% of the time it doesn't matter.
- If spdx.org returns a 404, how do we know whether that's a problem with the spec and not a problem with spdx.org being down? or a firewall problem?

paulatulis:
Thanks for the thoughts and time and help, @wmhilton!

Issue #1664, opened 2021-08-03
Request maker body and sample should have a max height


mnaumanali94:
The body in request maker should have a max height much like response example. This should be both in reference and embedded markdown

To replicate:
1. Go here: https://schoolyear.stoplight.io/docs/schoolyear-api-docs/b3A6OTM2NTkyMQ-create-new-exam
2. Run `__SL.state.flags.enabledFlags.elements_v7 = true` in console

![Screenshot 2021-08-03 at 4.46.12 PM.png](https://images.zenhubusercontent.com/5ea9a7f489044245dbd23d46/8fae6baa-4ffd-464e-a766-350be5337e4a)

paulatulis:
Hmmm I am trying to reproduce this locally and cannot... am I looking at the wrong component?

![2021-08-19 15.20.33.gif](https://images.zenhubusercontent.com/5dc060647a4d02000177a6d0/d8558935-93da-43cb-b055-05ed9953c7c2)

mnaumanali94:
Markdown Try-it has it for sure. This was fixed in some other PR i guess.

paulatulis:
Is it alright if I go ahead and close this issue in that case?

PR #1701, opened 2021-08-19
fix: maxHeight for panels

paulatulis:
Should address #1664 specifically request sample panel was unnecessarily long. Request body looks to already be fixed.

before: 
Screen Shot 2021-08-23 at 10 35 45 AM

after:
Screen Shot 2021-08-23 at 10 35 22 AM

### DIFF
--- a/packages/elements-core/src/components/RequestSamples/RequestSamples.tsx
+++ b/packages/elements-core/src/components/RequestSamples/RequestSamples.tsx
@@ -87,7 +87,7 @@ export const RequestSamples = React.memo<RequestSamplesProps>(({ request }) => {
         <CodeViewer
           aria-label={requestSample ?? fallbackText}
           noCopyButton
-          maxHeight="510"
+          maxHeight="400px"
           language={mosaicCodeViewerLanguage}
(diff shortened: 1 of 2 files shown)


mnaumanali94:
request sample fix still should be merged, no?

wmhilton (approved):
Ah good old "px"

wmhilton:
@paulatulis FYI I've changed both this one and #1710 from "chore" to "fix" because the distinction is not how small the PR is, but whether they have an impact on customers and thus need to be released. Chores might include, fixing typos in the README or changing the CI pipeline to be faster or something. Both of these changes make a difference to how the application works from a customer perspective!

paulatulis:
> @paulatulis FYI I've changed both this one and #1710 from "chore" to "fix" because the distinction is not how small the PR is, but whether they have an impact on customers and thus need to be released. Chores might include, fixing typos in the README or changing the CI pipeline to be faster or something. Both of these changes make a difference to how the application works from a customer perspective!

Thank you! I get into the habit of using chore and forget to update. Will try to make it a point to be more mindful going forward.

Issue #1694, opened 2021-08-16
TryIt in Markdown has too large font


mpodlasin:
![Screenshot 2021-08-16 at 16.35.34.png](https://images.zenhubusercontent.com/5f857aaa311cec7e20b4f258/b92e9024-4b3d-410b-8840-fb4bdfca3171)

To reproduce, create an article with TryIt in it:

```
` ` ` json http
{
  "method": "PATCH",
  "url": "https://api.zoom.us/v2/accounts/accountId/billing",
  "query": {
    "access_token": [
      ""
    ]
  },
  "headers": {
    "Content-Type": [
      "application/json"
    ]
  },
  "body": {
    "address": "string",
    "apt": "string",
    "city": "string",
    "country": "string",
    "email": "string",
    "first_name": "string",
    "last_name": "string",
    "phone_number": "string",
    "state": "string",
    "zip": "string"
  }
}
` ` ` 
```

mmiask:
Let's make sure we don't display the input field in the line below parameters name

PR #1703, opened 2021-08-20
Fix: TryIt in Markdown has too large font

mmiask:
Fixes: #1694 

To reproduce, follow the steps in connected issue

**Before**

Screen Shot 2021-08-20 at 10 55 41 AM

**After**

Screen Shot 2021-08-20 at 10 53 33 AM


wmhilton (changes requested):
Hmm. I'm sure this works, but I'd like to try a slightly different approach since this seems kinda fragile and I'd rather us not hard-code CSS font sizes in CSS files.

I've posted an alternative solution at #1705. What do you think?

mmiask:
Closing in favor of: #1705

PR #1705, opened 2021-08-20
Fix: TryIt in Markdown has too large font

wmhilton:
this is Will's version of #1703


mmiask (approved):
Yeah, yours is much better. Approving!

PR #1708, opened 2021-08-23
fix: support for model grouping by tag in TOC

domagojk:
For stoplightio/platform-internal#7438

The issue was actually not related to `x-tags` but the fact that in elements-utils, standalone models were never grouped by tags. So I've added grouping in `appendModelsToToc`, similar how it is done in `appendHttpServiceItemsToToC`.

Most of the tests in `toc.test.ts` have a tag defined for models, which is why I had to update the expected output. It would be good if someone (@mnaumanali94 I guess? 🙂) reviews it and confirms that this is what we want, because it is a bit strange to me that there were no model grouping before, so I wonder if there was any reason behind it?.

### DIFF
--- a/packages/elements-utils/src/toc/toc.ts
+++ b/packages/elements-utils/src/toc/toc.ts
@@ -344,11 +344,39 @@ export function appendHttpServicesToToC(toc: ITableOfContents) {
 export function appendModelsToToc(toc: ITableOfContents, schemaType: SchemaType = 'divider') {
   return (models: NodeData[]) => {
     if (models.length) {
-      const childItems: TableOfContentItem[] = models.map(model => ({
-        type: 'item',
-        title: model.name,
-        uri: model.uri,
-      }));
+      const { others, groups } = models.reduce<{
+        groups: { [key: string]: Group };
+        others: Item[];
+      }>(
+        (result, model) => {
+          const [tagName] = model.tags || [];
+
+          if (tagName) {
+            if (result.groups[tagName.toLowerCase()]) {
+              result.groups[tagName.toLowerCase()].items.push({
+                type: 'item',
+                title: model.name,
+                uri: model.uri,
+              });
+            } else {
+              result.groups[tagName.toLowerCase()] = {
+                type: 'group',
+                title: tagName,
+                items: [{ type: 'item', title: model.name, uri: model.uri }],
+              };
+            }
+          } else {
+            result.others.push({ type: 'item', title: model.name, uri: model.uri });
+          }
+
+          return result;
+        },
+        { groups: {}, others: [] },
+      );
+
+      const childItems: TableOfContentItem[] = [];
+      others.forEach(item => childItems.push(item));
+      Object.entries(groups).forEach(([, group]) => childItems.push(group));
 
       if (schemaType === 'divider') {
         toc.items.push({ type: 'divider', title: 'Schemas' });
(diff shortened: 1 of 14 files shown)


packages/elements-utils/src/toc/toc.ts
```diff
@@ -344,11 +344,39 @@ export function appendHttpServicesToToC(toc: ITableOfContents) {
...
-        title: model.name,
-        uri: model.uri,
-      }));
+      const { others, groups } = models.reduce<{
+        groups: { [key: string]: Group };
+        others: Item[];
+      }>(
+        (result, model) => {
+          const [tagName] = model.tags || [];
+
```
wmhilton, line 353:
```suggestion
const item = { type: 'item', title: model.name, uri: model.uri };
```

packages/elements-utils/src/toc/toc.ts
```diff
@@ -344,11 +344,39 @@ export function appendHttpServicesToToC(toc: ITableOfContents) {
...
+          if (tagName) {
+            if (result.groups[tagName.toLowerCase()]) {
+              result.groups[tagName.toLowerCase()].items.push({
+                type: 'item',
+                title: model.name,
+                uri: model.uri,
+              });
```
wmhilton, line 360:
```suggestion
              result.groups[tagName.toLowerCase()].items.push(item);
```

packages/elements-utils/src/toc/toc.ts
```diff
@@ -344,11 +344,39 @@ export function appendHttpServicesToToC(toc: ITableOfContents) {
...
+            } else {
+              result.groups[tagName.toLowerCase()] = {
+                type: 'group',
+                title: tagName,
+                items: [{ type: 'item', title: model.name, uri: model.uri }],
```
wmhilton, line 365:
```suggestion
                items: [item],
```

packages/elements-utils/src/toc/toc.ts
```diff
@@ -344,11 +344,39 @@ export function appendHttpServicesToToC(toc: ITableOfContents) {
...
+              };
+            }
+          } else {
+            result.others.push({ type: 'item', title: model.name, uri: model.uri });
```
wmhilton, line 369:
```suggestion
            result.others.push(item);
```

wmhilton (approved):
Looks correct to me! I left some suggestions that go against my usual rule of "just copy and paste, DRY is over-rated" because I'm dehydrated and in a DRY mood I guess, feel free to ignore them lol. I'm sure applying them directly will make the linter unhappy.

wmhilton:
@domagojk Can you add an example to copy-paste or link to a page that @mnaumanali94 can look at to see how the look grouped?

mnaumanali94:
Screenshot 2021-08-23 at 11 30 00 PM

@domagojk I can't seem to see any effect here. What am i missing.

mnaumanali94:
I tested it using storyboook for dev-portal and a projectID.

mallachari:
If I remember correctly the initial idea was that standalone models are not grouped. Just listed in `SCHEMAS` section. Which doesn't mean that they shouldn't be, but that's the product question.

mnaumanali94:
> If I remember correctly the initial idea was that standalone models are not grouped. Just listed in `SCHEMAS` section. Which doesn't mean that they shouldn't be, but that's the product question.

Models inside APIs aren't being grouped either.

domagojk:
>  I left some suggestions that go against my usual rule of "just copy and paste, DRY is over-rated" because I'm dehydrated and in a DRY mood I guess, feel free to ignore them lol.

I agree with the suggestion. Sometimes, even DRY code can be consider better than the alternative 😄 

> Edit: wait, elements-utils... does that mean this needs to be yalced into platform to see it?

Oh yea. That is why it doesn't work @mnaumanali94.

I think we can merge this and test it in version bump PR

PR #1717, opened 2021-08-24
fix: external scrollbar

mallachari:
Addresses: https://github.com/stoplightio/platform-internal/issues/7661

ToC in the platform doesn't use its own scrollbar like the elements or elements-dev-portal version, it rather uses higher-level one. This means that the default mechanism to detect whether ToC is using scrollbar doesn't affect here. 

This PR adds an optional property to ToC to indicate that there's an external scrollbar and we can scroll to active item.
Platform PR will need to add that property to ToC container.


mpodlasin (approved):
Ah!

I didn't expect that solution, but I am ok with it. 😄

PR #1718, opened 2021-08-25
feat: add cors proxy for try it

### LINKED ISSUE #1037: Add proxy support to Try It
## Overview

When using TryIt we should support interacting with APIs that have not enabled CORS headers for whatever domain the elements based documentation is running on.

- elements users can provide their own CORS proxy as a property, otherwise its all done directly.
- elements-dev-portal can also provide their own, but by default will use the hosted CORS proxy we set up for the demo at `https://cors.stoplight.io/`.

## Requirements

Let's add a `proxy` prop to both public components, which takes a string. As we are using `fetch()` try and find some portable solution that works in the browser. node-fetch for instance can do this:

```
const fetch = require('node-fetch');
const HttpsProxyAgent = require('https-proxy-agent');

(async () => {
    const proxyAgent = new HttpsProxyAgent('http://46.250.171.31:8080');
    const response = await fetch('https://httpbin.org/ip?json', { agent: proxyAgent});
    const body = await response.text();
    console.log(body);
})();
```

If that is not going to work, prepend it to the URL as we do for the demo:

```
const url = `https://cors.stoplight.io/${actualUrl}`;
```

## Example

```
<API apiDescriptionUrl="https://some-domain/oas.yaml" proxy="https://cors-anywhere.example.io/" />` 
```

## Out of Scope

This will eventually be replaced with hosted Prism Proxy, but that will require some work to do right. Let's unblock people with this existing CORS proxy then move over to Prism Proxy for elements-dev-portal when we can.

mpodlasin:
Closes #1037

### DIFF
--- a/packages/elements/src/containers/API.tsx
+++ b/packages/elements/src/containers/API.tsx
@@ -73,6 +73,13 @@ export interface CommonAPIProps extends RoutingProps {
    * @default false
    */
   hideExport?: boolean;
+
+  /**
+   * Url of a CORS proxy that will be used to send requests in TryIt.
+   * Provided url will be prepended to an URL of an actual request.
+   * @default false
+   */
+  tryItCorsProxy?: string;
 }
 
 const propsAreWithDocument = (props: APIProps): props is APIPropsWithDocument => {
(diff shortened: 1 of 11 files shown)


mpodlasin:
Since this PR doesn't have stoplight proxy included, I would merge it as is. Sounds good? @mnaumanali94 @wmhilton

packages/elements/src/containers/API.tsx
```diff
@@ -73,14 +73,30 @@ export interface CommonAPIProps extends RoutingProps {
...
-  const { layout, apiDescriptionUrl = '', logo, hideTryIt, hideSchemas, hideInternal, hideExport } = props;
+  const {
+    layout,
+    apiDescriptionUrl = '',
+    logo,
+    hideTryIt,
+    hideSchemas,
+    hideInternal,
+    hideExport,
+    tryItCorsProxy,
```
mnaumanali94, line 98: @pamgoodrich can we document this in the elements config here please: https://meta.stoplight.io/docs/elements/ZG9jOjMyNjU4OTY1-getting-started-with-elements-in-react#configuration

PR #1719, opened 2021-08-25
chore: optimize stacked layout

mpodlasin:
Up for discussion/testing.

I've noticed that stacked layout was getting quite slow recently, so I made an optimisation attempt.

If we can validate that it doesn't have any serious regressions, I would encourage to merge it. ;)


mallachari (approved):
Works way better. I don't see any regressions or drawbacks with it.

PR #1720, opened 2021-08-25
feat: validate required parameters in try it

### LINKED ISSUE #1566: Provide a warning when users use TryIt without required parameters.
We currently don't validate at all if users provided all the required parameters to the request when using TryIt.

This can cause a confusion when seeing an error message like that (in this case from our mock API):

Depending on the API, those messages can be very cryptic and users will find themselves confused what is wrong.

Thus I propose, in addition to displaying an error response, to display a warning message (for example above the error), saying that not all required parameters were provided and this might have caused an error response.

mpodlasin:
Closes #1566 

Screenshot 2021-08-25 at 10 43 54
Screenshot 2021-08-25 at 11 08 22

### DIFF
--- a/packages/elements-core/src/components/TryIt/Parameters/OperationParameters.tsx
+++ b/packages/elements-core/src/components/TryIt/Parameters/OperationParameters.tsx
@@ -8,9 +8,15 @@ interface OperationParametersProps<P extends keyof any = string> {
   parameters: readonly ParameterSpec[];
   values: Record<P, string>;
   onChangeValue: (parameterName: P, newValue: string) => void;
+  validate?: boolean;
 }
 
-export const OperationParameters: React.FC<OperationParametersProps> = ({ parameters, values, onChangeValue }) => {
+export const OperationParameters: React.FC<OperationParametersProps> = ({
+  parameters,
+  values,
+  onChangeValue,
+  validate,
+}) => {
   return (
     <Panel defaultIsOpen>
       <Panel.Titlebar>Parameters</Panel.Titlebar>
@@ -21,6 +27,7 @@ export const OperationParameters: React.FC<OperationParametersProps> = ({ parame
             parameter={parameter}
             value={values[parameter.name]}
             onChange={(value: string | number) => onChangeValue(parameter.name, String(value))}
+            validate={validate}
           />
         ))}
       </Panel.Content>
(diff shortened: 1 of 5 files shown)


mallachari:
It works great. One thing that I've noticed is that the request is sent even though the parameters are missing. We might prevent sending it in that case.

https://user-images.githubusercontent.com/14196079/130767810-251d940e-bcb6-4ba8-9522-ab23714dc8ca.mov

mallachari:
Or do we want the request to be sent and just notify user about missing parameters?

mpodlasin:
> Or do we want the request to be sent and just notify user about missing parameters?

Yeah, I left it on purpose. @philsturgeon was talking about this at some point and he convinced me.

The reason is that docs might say that a parameter is obligatory, while in reality is not. In that case our TryIt would block you from sending a request, even though the API would return an answer.

mallachari:
> > Or do we want the request to be sent and just notify user about missing parameters?
> 
> Yeah, I left it on purpose. @philsturgeon was talking about this at some point and he convinced me.
> 
> The reason is that docs might say that a parameter is obligatory, while in reality is not. In that case our TryIt would block you from sending a request, even though the API would return an answer.

Yeah, that makes sense. Then there's nothing else to do here

mallachari (approved):

mnaumanali94:
This is awesome!

I personally think we should not allow the user to send the request if a required parameter is missing. That's an invalid OAS that the creator should fix, and this could be a forcing function. CC: @philsturgeon

mpodlasin:
> This is awesome!
> 
> I personally think we should not allow the user to send the request if a required parameter is missing. That's an invalid OAS that the creator should fix, and this could be a forcing function. CC: @philsturgeon

Maybe I'll merge as is? 

If we later decide that we indeed want to block requests, it will be easy to add in a separate PR.

mnaumanali94:
Lets block requests in this PR as we'd like to shout about this feature once released. I feel like users would assume validation means requests will be blocked.

philsturgeon:
Yeah I don’t remember stating opinions on this topic. I think I said we should not block for invalid values (that’s not a string, that’s not an email address!), but consider required to be different to that. Good call @mnaumanali94.

mpodlasin:
@mallachari can you make a re-review with not fetching when required params are not provided? 🙏

mallachari (approved):
Still looks good

mnaumanali94 (approved):

PR #1585, opened 2021-07-09
chore: option to set credentials policy in fetch request

haseebeqx:

Issue #1548, opened 2021-07-05
Cookies are not included in the request


haseebeqx:
I wanted to try an api that involves cookies. Currently no cookies were sent while trying the api. 

Is there an option to disable this default behaviour and send cookies like a regular web request.

## Expected behavior

There should be an option to sent cookies if required.

philsturgeon:
This is sadly not possible due to the way browsers handle cookies. You can disable Try It with hideTryIt+true if its not useful for your API.

haseebeqx:
@philsturgeon I was able to fix  this by changing https://github.com/stoplightio/elements/blob/42b2118aa4ef286b3fdaf0858ac425475008d1ff/packages/elements-core/src/components/TryIt/build-request.ts#L73
 to
```js
credentials: 'same-origin' 
```

If you want I can contribute to this repo by adding an addidional option field `TryItCredentialPolicy` with default value of `omit` and user can set `include` or `same-origin` .

Reopen this issue if you are interested.

philsturgeon:
Very interested! See if you can pull request something and I'll take a look.

mnaumanali94:
Need to review and merge this pull request.

wmhilton:
@mallachari @mmiask @mpodlasin @paulatulis Hey team! Can you please add your [planning poker estimates with ZenHub](https://app.zenhub.com/workspaces/-Team-Pierogi-Platoon-610871b8d6e5310013071b72/issues/stoplightio/elements/1548?planning-poker)

PR #1729, opened 2021-08-26
feat: tryit credentials policy

mallachari:
Addresses https://github.com/stoplightio/elements/issues/1548 
Based on @haseebeqx PR at https://github.com/stoplightio/elements/pull/1585

Adds support for setting credential policy in TryIt by adding the optional `tryItCredentialsPolicy` property.

@haseebeqx your PR is great, thanks for contribution. I've moved it here so it's not based on the fork and did some small changes:
- added similar property for `elements-dev-portal`
- removed redundant `tryItCredentialsPolicy` from `StackedLayout` - we're using TryIt separately there
- setting default value to `omit` in `build-request`
- added basic tests

### DIFF
--- a/packages/elements-core/src/components/Docs/Docs.tsx
+++ b/packages/elements-core/src/components/Docs/Docs.tsx
@@ -43,6 +43,14 @@ interface BaseDocsProps {
   exportProps?: ExportButtonProps;
 
   /**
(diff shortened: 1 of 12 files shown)


mallachari:
Yes, exactly, it uses cookies directly. Credentials won't mean anything in this case.

PR #1732, opened 2021-08-30
fix: check x-examples for model multiple examples

domagojk:
For https://github.com/stoplightio/platform-internal/issues/7698

x-examples are now also checked for Model examples dropdown.

It can be tested by copy/pasting an object with `x-examples` in storybook (in `nodeData`)


paulatulis (approved):
Works great for me!

mnaumanali94 (approved):

Issue #1619, opened 2021-07-20
Allow server URLs to be copied


mnaumanali94:
There's no easy way to copy the Server URLs in elements.

Let's add an option for users to copy Server/Mock server URLs in the API overview which shows up on hover.

Also fix the HTML in this one to allow highlighting just the URL.

PR #1737, opened 2021-08-31
feat: server url copy button

mpodlasin:
Related to #1619 

Screenshot 2021-08-31 at 15 40 13


paulatulis (commented):
I was just interacting with the preview, and this works like a charm, but I would never know to hover over in the first place. Is it overkill to add an icon or some other visual cue?

mpodlasin:
> I was just interacting with the preview, and this works like a charm, but I would never know to hover over in the first place. Is it overkill to add an icon or some other visual cue?

Yeah, I implemented what was required in the issue. However I agree with the doubts.

@mnaumanali94 thoughts?

paulatulis (approved):
Approving per our conversation in standup, which was that we can add visual enhancements later, but the functionality should be merged, and this is functional!

PR #1738, opened 2021-09-01
fix: support examples with and without value prop

domagojk:
For https://github.com/stoplightio/platform-internal/issues/7698

`x-examples` can be defined with and without `value` and both should be supported:

```json
{
  "title": "User",
  "type": "object",
  "description": "",
  "x-examples": {
    "Alice Smith": {
      "id": 142,
      "firstName": "Alice",
      "lastName": "Smith",
      "email": "alice.smith@gmail.com",
      "dateOfBirth": "1997-10-31",
      "emailVerified": true,
      "signUpDate": "2019-08-24"
    },
    "example-2": {
      "value": {
        "id": 0,
        "firstName": "n",
        "lastName": "string",
        "email": "user@example.com",
        "dateOfBirth": "1997-10-31",
        "emailVerified": true,
        "createDate": "2019-08-24"
      }
    },
  },
  "properties": {
    "id": {
      "type": "integer",
      "description": "Unique identifier for the given user."
    },
    "firstName": {
      "type": "string"
    },
    "lastName": {
      "type": "string"
    },
    "email": {
      "type": "string",
      "format": "email"
    },
    "dateOfBirth": {
      "type": "string",
      "format": "date",
      "example": "1997-10-31"
    },
    "emailVerified": {
      "type": "boolean",
      "description": "Set to true if the user's email has been verified."
    },
    "createDate": {
      "type": "string",
      "format": "date",
      "description": "The date that the user was created."
    }
  }
}
```

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.2.5",
+  "version": "7.2.6",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**",
--- a/packages/elements-core/src/utils/exampleGeneration.ts
+++ b/packages/elements-core/src/utils/exampleGeneration.ts
@@ -64,10 +64,11 @@ export const generateExamplesFromJsonSchema = (schema: JSONSchema7): Example[] =
     });
   } else if (isPlainObject(schema?.['x-examples'])) {
     for (const [label, example] of Object.entries(schema['x-examples'])) {
-      if (isPlainObject(example) && example.hasOwnProperty('value')) {
+      if (isPlainObject(example)) {
+        const val = example.hasOwnProperty('value') && Object.keys(example).length === 1 ? example.value : example;
         examples.push({
           label,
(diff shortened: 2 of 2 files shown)


packages/elements-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.2.5",
+  "version": "7.2.6",
```
mmiask, line 3: Please consider creating separate PR for a release. I think that Mateusz pushed a change in the meantime.

mmiask (commented):
Looks good! I left only a one comment regarding releasing Elements

mmiask (approved):

packages/elements-core/package.json
domagojk: I think we can do it for a quick fixes like this. The branch is up to date and only a single (internal) package is changed, so it seems like a ok thing to do 🤷‍♂️

mnaumanali94:
Doesn't HTTP spec normalize examples?

domagojk:
@mnaumanali94 no since we use this function on json schema objects (not oas spec which is transformed with http-spec)

PR #1749, opened 2021-09-07
feat (dev-portal): add search component

paulatulis:
Part 1 of #[7715](https://app.zenhub.com/workspaces/-team-pierogi-platoon-610871b8d6e5310013071b72/issues/stoplightio/platform-internal/7715)

Gets dev portal's Search component to the point where it functions in an example application. Once this is approved, I'll write corresponding documentation and get that reviewed as well :) 

WIP Docs: https://github.com/stoplightio/elements/compare/ps-docs-search


examples/react-cra/src/components/Navigation.tsx
```diff
@@ -1,11 +1,19 @@
...
+import { Search } from './Search';
+
 export const Navigation = () => {
+  const queryClient = new QueryClient();
+
   return (
     <nav className="topnav">
       <NavLink to="/stoplight-project">Stoplight Project</NavLink>
       <NavLink to="/zoom-api">Zoom API</NavLink>
+      <QueryClientProvider client={queryClient}>
```
paulatulis, line 14: My goal is to not have to include this here because I have already included it in the component file [here](https://github.com/stoplightio/elements/pull/1749/files#diff-8eac8a4161d783c42a3155b47969f0ac84a1878213c4638df85c873fc0f46594R130), but I continue to get error that no QueryClient is set if I remove it.

wmhilton: I have a strong suspicion that it's because it doesn't know what URL to do the query from. I think it's because StoplightProject takes in the platform URL and has an extra `<DevPortalProvider platformUrl={platformUrl}>` and I suspect we need to do the same with the search component.

paulatulis: So I did also try that and it did fix this issue. I can give it another shot, though!

wmhilton: HMMM! Fascinating. OK. 🤔

mnaumanali94:
How did it work in our dev portal if it needed changes? O.o Apologies for my ignorance

paulatulis:
> How did it work in our dev portal if it needed changes? O.o Apologies for my ignorance

I actually am not sure it does work in our dev portal. [This](https://github.com/stoplightio/dev-portal/blob/make-it-work/components/Search.tsx) is on a branch called make-it-work and I don't see it anywhere in the [dev portal application](https://stoplight-dev-portal.vercel.app/).

The provider that we use for `StoplightProject` doesn't work the same magic for the `Search` component since it relies on `useGetNodes`. 

No need to apologize! I was under the same impression 😄

wmhilton:
@mnaumanali94 we do not use this component at all in Ninja. We use a completely different implementation (`packages/ninja/src/components/DocsSearch.tsx`) the Search component in elements-dev-portal appears to be an old, unmaintained, half-working thing.

mpodlasin:
Screenshot 2021-09-08 at 12 48 01

I've made it work. It only needed a MosaicProvider, because modals in Mosaic require it to work.

This however should be changed and MosaicProvider should be put in DevPortalProvider so that user's don't have to know anything about mosaic.

After this, I imagine everything should work. The only requirement for Search to work should be DevPortalProvider, nothing else. Sorry if this wasn't clear.


Screenshot 2021-09-08 at 12 50 05

paulatulis:
Thanks so much for the assistance @mpodlasin! I will look into the provider swap today.

packages/elements-dev-portal/src/components/DevPortalProvider/index.tsx
```diff
@@ -13,7 +14,11 @@ const PlatformProvider: React.FC<DevPortalProviderProps> = ({
   platformAuthToken,
   children,
 }) => {
-  return <PlatformContext.Provider value={{ platformUrl, platformAuthToken }}>{children}</PlatformContext.Provider>;
+  return (
+    <PlatformContext.Provider value={{ platformUrl, platformAuthToken }}>
+      <MosaicProvider>{children}</MosaicProvider>
```
paulatulis, line 19: Is this allowed to happen? Is there a better/more elegant way to get this in here so it's not exposed to the end user using dev portal?

examples/react-cra/src/components/Search.tsx
```diff
@@ -0,0 +1,40 @@
...
+    workspaceId,
+  });
+
+  const handleClose = () => {
+    setOpen(false);
+    setSearch('');
+  };
+
+  const handleClick = (data: NodeSearchResult) => {
+    console.log('clicked', data);
```
mpodlasin, line 24: It wouldn't be that much more work to actually make it send you to a proper article url.

packages/elements-dev-portal/src/components/DevPortalProvider/index.tsx
mpodlasin: I think this is 💯 proper approach. My only worry would be conflicts in platform.

In platform MosaicProvider and DevPortalProvider are used in separation, which is probably ok, because mosaic is used explicitly there (or at least will be).

So what happens when there are two MosaicProviders in the code? Does it crash? Cause bugs? Or is it safe? 🤔 Those questions would have to be answered before we can make that change.

paulatulis: Ahh okay that makes sense. Hmm I could try yalcing into platform to see what happens, but maybe even that wouldn't show the full story.

examples/react-cra/src/components/Search.tsx
wmhilton: yes, but even then I'm not sure exactly how to make it work. It looks like internally the API component uses `react-router`... I'm not sure how to hook the API component and this callback together.

packages/elements-dev-portal/src/components/DevPortalProvider/index.tsx
paulatulis: Update: I yalced into ninja and interacted with the currently existing dev-portal components. I also added this Search component to test. Nothing misbehaved that I observed, though, I'm still not sure this is a completely reliable way to confirm that having two MosaicProviders is alright. Happy to give anything else a shot!

paulatulis: Something else I noticed, is that when I try to implement the dev-portal `Search` component in ninja to make sure it doesn't get messed up with all the different providers, is that it's built to take strings of both `projectIds` and `workspaceId`. In ninja, I'm getting back `projectIds` and `workspaceIds` that are numbers and look nothing like those that we plug into the example. Even if I hard code that data, I don't receive any search results in platform, but I do in elements storybook and elements examples. We already have a search component that works and I am not trying to replace it, but if other people try to implement this in their environments, I'm worried it's not going to work for them, so I'd like to figure out why that's happening.

paulatulis: I'm having more thoughts and questions about this whole search thing. Doesn't someone need to be able to grab their encoded `workspaceId` from their workspace to be able to plug it into the component since it uses `useGetNodes`? That means we need to have somewhere in the UI in platform that exposes this.

packages/elements-dev-portal/src/handlers/getNodes.ts
```diff
@@ -2,31 +2,41 @@ import { NodeSearchResult } from '../types';
...
   platformUrl?: string;
   platformAuthToken?: string;
 }): Promise<NodeSearchResult[]> => {
   const queryParams = [];
+  let fetchedWorkspaceId = workspaceId || '';
 
-  if (projectIds && projectIds.length) {
-    queryParams.push(...projectIds.map((projectId, index) => `project_ids[${index}]=${projectId}`));
+  if (!workspaceId) {
+    const response = await fetch(`${platformUrl}/api/v1/projects/${projectIds[0]}`, {
```
mpodlasin, line 21: Hmm, I am on the fence here.

On one hand it will likely work quite well, and it is a fast fix.

On the other, the cleaner solution would be to actually update elements-service, to not require workspaceId when searching nodes... 🤔 

@wmhilton would appreciate your opinion here.

paulatulis: Yeah, agree! I borrowed this from `getNodeContent` because it was quicker/easier for me to handle here. But, I am open to approaching from the service side. If that's preferred, I may need some additional help getting acquainted over there.

examples/react-cra/src/components/Search.tsx
paulatulis: I need some help with this conceptually. I was originally thinking to just use `window.location.href` and set it to the node, but will need to then get the workspace slug from the encoded ID that comes back in `data.` I have been searching to see if I could find where something like that is already happening anywhere but not sure it is. This seems like not what Will was thinking...

packages/elements-dev-portal/src/handlers/getNodes.ts
paulatulis: I kind of think that editing the service should be a different issue because the scope is getting quite large. Feel free to push back on me!

examples/react-cra/src/components/Search.tsx
paulatulis: I have a naive approach I committed in 4771e98. I have a feeling this is the least efficient way to do it, but I am out of ideas. Would love to get some guidance if possible.

packages/elements-dev-portal/src/handlers/getNodes.ts
wmhilton: Hmmmmmm. I agree with @mpodlasin that it would be cleaner to just fix up the API so we didn't need the workspace id. And while `/v1/projects/{project_id}/nodes` would be an obvious choice (seeing how the other paths are named)

image

That would mean changing the hook so you can only pass one `project_id`? And IDK how to modify the endpoint to handle multiple ids. Would it be `/v1/projects/{comma_separated_ids}/nodes`? What other complicated things would turn up?

So... I'm inclined to just shrug and say "sure" for now? Weird to have to make two HTTP requests but whatever. Let's figure that out in a new issue I guess.

wmhilton (approved):
Love it! Much cleaner than it was originally, and although the internals got messy, it got easier for users to use. So that's a good thing.

I think we should tackle making the search endpoint more generic in another issue, another time. Maybe `/v1/search?projects=[1,2,3]&query=todo` or something.

paulatulis:
> Love it! Much cleaner than it was originally, and although the internals got messy, it got easier for users to use. So that's a good thing.
> 
> I think we should tackle making the search endpoint more generic in another issue, another time. Maybe `/v1/search?projects=[1,2,3]&query=todo` or something.

I would really like to do that and circle back to this and make it cleaner. It's very messy! But to avoid dragging this particular issue out longer, I'll merge this and then create a new issue to accomplish your suggestion, @wmhilton. Thanks for reviewing!

Issue #1746, opened 2021-09-06
Copying server URL doesn't work for mock URL


mpodlasin:
We implemented copying server URL in API Overview section, but forgot to do the same for mock URL.

![Screenshot 2021-09-06 at 11.33.19.png](https://images.zenhubusercontent.com/5f857aaa311cec7e20b4f258/bd17f4bc-aaae-4758-805a-007ae0f258f5)

PR #1760, opened 2021-09-10
fix: copy mock server url

mpodlasin:
Addresses https://github.com/stoplightio/elements/issues/1746

Screenshot 2021-09-10 at 15 07 23


mnaumanali94 (approved):

paulatulis (approved):
Looks good!

PR #1770, opened 2021-09-16
fix: allow large specs to load schema

paulatulis:
Addresses #[7923](https://app.zenhub.com/workspaces/-team-pierogi-platoon-610871b8d6e5310013071b72/issues/stoplightio/platform-internal/7923)

Before: page becomes unresponsive when clicking around to models with circular refs

After:

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -49,7 +49,7 @@
     "@fortawesome/react-fontawesome": "^0.1.11",
     "@stoplight/json": "^3.10.0",
     "@stoplight/json-schema-ref-parser": "^9.0.5",
-    "@stoplight/json-schema-sampler": "0.2.0",
+    "@stoplight/json-schema-sampler": "0.2.1",
     "@stoplight/json-schema-viewer": "^4.3.1",
     "@stoplight/markdown": "^3.1.1",
     "@stoplight/markdown-viewer": "^5.3.2",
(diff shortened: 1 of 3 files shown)


wmhilton:
Did we figure out whether the issue was circular refs or just "big models"? because limiting the depth to 4 seems to imply it's big models.

paulatulis:
> Did we figure out whether the issue was circular refs or just "big models"? because limiting the depth to 4 seems to imply it's big models.

@wmhilton I don't think we confirmed one way or the other. Could it not be that it's both? 

Also, if we don't limit this number, won't the example be extremely long to the point someone is scrolling forever? Is it helpful from a product standpoint to have such a long example?

Nezteb:
> ```
> ../packages/elements-core/src/utils/exampleGeneration.ts:82:5 - error TS2345: Argument of type '{ maxSampleDepth: number; }' is not assignable to parameter of type 'Options'.
>   Object literal may only specify known properties, and 'maxSampleDepth' does not exist in type 'Options'.
> 
> 82     maxSampleDepth: 4,
>        ~~~~~~~~~~~~~~~~~
> 
> 
> Found 1 error.
> ```

🤔

paulatulis:
> > ```
> > ../packages/elements-core/src/utils/exampleGeneration.ts:82:5 - error TS2345: Argument of type '{ maxSampleDepth: number; }' is not assignable to parameter of type 'Options'.
> >   Object literal may only specify known properties, and 'maxSampleDepth' does not exist in type 'Options'.
> > 
> > 82     maxSampleDepth: 4,
> >        ~~~~~~~~~~~~~~~~~
> > 
> > 
> > Found 1 error.
> > ```
> 
> 🤔

This is because `json-schema-sampler` hasn't been released yet after [this PR](https://github.com/stoplightio/json-schema-sampler/pull/20) was just approved. It still thinks this is not a prop.

wmhilton (approved):
I like it!

PR #1773, opened 2021-09-20
feat: allow pausing getNodes query by passing prop

mmiask:
A part of https://github.com/stoplightio/platform-internal/pull/8024

This allows to put the `useGetNodes` in the `idle` state when we explicitly tell it to (more about pausing queries in https://react-query.tanstack.com/guides/disabling-queries)

We could of course make the hook be in `idle` until `search` isn't falsy but this would change the current, default behavior of this hook. With this change we don't, you need to explicitely state if you want to pause the query or not. I was trying to mimick the behavior of `urql(..., { pause })` option

### DIFF
--- a/packages/elements-dev-portal/src/hooks/useGetNodes.ts
+++ b/packages/elements-dev-portal/src/hooks/useGetNodes.ts
@@ -10,16 +10,19 @@ export function useGetNodes({
   workspaceId,
   branchSlug,
   projectIds,
+  pause,
 }: {
   search: string;
   workspaceId?: string;
   branchSlug?: string;
   projectIds?: string[];
+  pause?: boolean;
 }) {
   const { platformUrl, platformAuthToken } = React.useContext(PlatformContext);
   const [debounceSearch] = useDebounce(search, 500);
   return useQuery(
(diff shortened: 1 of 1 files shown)


packages/elements-dev-portal/src/hooks/useGetNodes.ts
```diff
@@ -10,16 +10,19 @@ export function useGetNodes({
...
   branchSlug?: string;
   projectIds?: string[];
+  pause?: boolean;
 }) {
   const { platformUrl, platformAuthToken } = React.useContext(PlatformContext);
   const [debounceSearch] = useDebounce(search, 500);
   return useQuery(
     ['workspaceNodes', workspaceId, branchSlug, projectIds, debounceSearch, platformUrl, platformAuthToken],
     () => getNodes({ workspaceId, branchSlug, projectIds, search: debounceSearch, platformUrl, platformAuthToken }),
+    { enabled: pause ? false : true },
```
wmhilton, line 26: I think `!pause` is clearer here but maybe that's just me.

mmiask: 👍 I will do that

wmhilton (approved):

wmhilton:
oh oops that's not what I meant 🤦🏻‍♂️

PR #1774, opened 2021-09-20
fix: remove unnecessary logic

wmhilton:


paulatulis (approved):

mmiask (approved):

packages/elements-dev-portal/src/hooks/useGetNodes.ts
```diff
@@ -23,6 +23,6 @@ export function useGetNodes({
   return useQuery(
     ['workspaceNodes', workspaceId, branchSlug, projectIds, debounceSearch, platformUrl, platformAuthToken],
     () => getNodes({ workspaceId, branchSlug, projectIds, search: debounceSearch, platformUrl, platformAuthToken }),
-    { enabled: !pause ? true : false },
```
Nezteb, line 26: I'd prefer `pause ? false : true` :kappa:

(Kidding)

Nezteb (approved):

packages/elements-dev-portal/src/hooks/useGetNodes.ts
wmhilton: well that's what it was _originally_... https://github.com/stoplightio/elements/pull/1773#discussion_r712394496

I'm fine with changing it back to that if it's actually easier for y'all to read. Maybe I'm just too old and wizened. @paulatulis @mmiask does `? false : true` it make it clearer that the returned value is a boolean or you like that there's no implicit type casting to remember or...?

Nezteb: (I was 100% kidding and think `!pause` is pretty "idiomatic"!)

wmhilton: FYI, I frequently would tell Jakub R his code was too terse and hard to read and he should make it “dumber” so I could understand it at a glance, so I’m totally open to receiving the same feedback about my code.

PR #1775, opened 2021-09-20
chore: release elements-dev-portal 1.4.0

mmiask:
Release:
- `elements-dev-portal`: 1.4.0

fix: remove unnecessary logic (#1774)
feat: allow pausing getNodes query by passing prop (#1773)


Nezteb (approved):
I'm approving this because all of the checks pass. I don't know if there is any manual testing I'm supposed to do.

mmiask:
@Nezteb by now, everything should be tested within PRs introducing changes, but a few sanity checks are always welcomed :) especially those related to what will be published

PR #1781, opened 2021-09-23
fix: default values for optional parameters

mallachari:
Addresses: https://github.com/stoplightio/platform-internal/issues/7664

Removes default value from optional enum parameters.
Adds `required` property when mapping from JsonSchema to OpenApi parameter schema.

Simple API document for testing: [api](https://github.com/stoplightio/platform-internal/files/7016097/5547.zip)


domagojk (commented):
Code looks good, and it works, but once you select an option there is no way to unselect it. Would it make sense to include blank option if the field is not required? 🤔

mallachari:
> Code looks good, and it works, but once you select an option there is no way to unselect it. Would it make sense to include blank option if the field is not required? 🤔

That's valid point. I'll add it

domagojk (approved):
👍

Issue #1757, opened 2021-09-10
Enabling mocking adds an empty Prefer header


philsturgeon:
## Describe the bug

When I enable mocking in try it using StoplightProject, I see an empty Prefer: header show up even though no config has been set for mocking. 

## To Reproduce

1. Given this OpenAPI document: https://meta.stoplight.io/docs/sample-specs/b3A6OTM1-random-sticker
2. Enable Mocking in Try It
3. See error `--header 'Prefer:'`

## Expected behavior

If no preferences set lets keep that header out of there. 

**Screenshots**

<img width="491" alt="Screen Shot 2021-09-10 at 10 29 38 AM" src="https://user-images.githubusercontent.com/67381/132833131-99187c19-566c-4f1e-9af2-be313f589501.png">

**Environment (remove any that are not applicable):**
 - Library version: Sept 10th 2021
 - Platform: web

PR #1782, opened 2021-09-24
fix: do not include prefer header with no mock data

mallachari:
Addresses https://github.com/stoplightio/elements/issues/1757

Prevents adding `Prefer` header when no mocking options are selected

Screenshot 2021-09-24 at 11 31 49


mmiask (approved):
Simple, clean - looks good!

Issue #1750, opened 2021-09-07
Missing array enum item definitions in query parameters


pelzhaus:
## Describe the bug

Enums items of arrays in query parameters are not rendered correctly. This is related to https://github.com/stoplightio/elements/issues/1665 but now only appearing in query parameters.

I believe this is the right repo to post in, but if not let me know where I can add this!

## To Reproduce

Given this OpenAPI path parameters:
```json
        "parameters": [
          {
            "description": "An array query parameter",
            "in": "query",
            "name": "enum_array",
            "required": false,
            "schema": {
              "description": "An array query parameter",
              "items": {
                "enum": [
                "foo",
                "bar"
                ],
                "type": "string"
              },
              "type": "array"
            }
          }
	    ]
```
In the old-UI Stoplight, this is rendered as:
![image](https://user-images.githubusercontent.com/12112738/132413828-8c1d7cc7-b83b-4aad-a955-16779d8b474f.png)

In the new-UI Stoplight, this is rendered as:
![image](https://user-images.githubusercontent.com/12112738/132414091-b9131db4-7e77-481a-b5dd-2f7304072f5e.png)

## Expected behavior

I would expect the rendering to look just like body array enums, with `array[string]` description and allowed values:
![image](https://user-images.githubusercontent.com/12112738/132416315-1f052d70-7449-4a14-baa6-fc2e80d3b902.png)

## Additional context

**Screenshots**
See above

**Environment (remove any that are not applicable):**
 - Library version: ???
 - OS: macOS catalina 10.15.6
 - Browser: Chrome 92.0.4515.159
 - Platform: web

rawbee:
+1

awendland:
Having full type information displayed for our query parameters in endpoint docs is important to ensure our customers know how to interface with the API correctly. We're blocked on upgrading to the new UI by this. Thanks for the progress made on #1665 already!

PR #1783, opened 2021-09-24
Fix: Missing array enum item definitions in query parameters

mmiask:
Resolves: #1750

Enables `Elements` to display enum definitions places in `schema/items/enum`. Until now only `schema/enum` was supported. We are now also displaying array subtype for request parameters.

I've also added unit tests, and updated some existing ones.

## Testing

Testing schema for OAS3 and OAS3.1 (put it in `paths/parameters` in API document):

```json
{
          "description": "An array query parameter",
          "in": "query",
          "name": "enum_array",
          "required": false,
          "schema": {
            "description": "An array query parameter",
            "items": {
              "enum": [
                "foo",
                "bar"
              ],
              "type": "string"
            },
            "type": "array"
          }
        }


```

### **Before**

Screen Shot 2021-09-24 at 4 53 44 PM

### **After**

Screen Shot 2021-09-24 at 4 54 51 PM

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
@@ -76,7 +76,7 @@ export const Parameter: React.FunctionComponent<IParameterProps> = ({ parameter,
   const schemaExamples = parameter.schema?.examples;
   const schemaExamplesArray = Array.isArray(schemaExamples) ? schemaExamples : [];
 
-  let validations = omitBy(
+  const validations = omitBy(
     {
       ...omit(parameter, ['name', 'required', 'deprecated', 'description', 'schema', 'style', 'examples']),
       ...omit(get(parameter, 'schema'), ['description', 'type', 'deprecated']),
--- a/packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
@@ -55,7 +55,9 @@ export const Parameter: React.FunctionComponent<IParameterProps> = ({ parameter,
   // TODO (CL): This can be removed when http operations are fixed https://github.com/stoplightio/http-spec/issues/26
   const description = get(parameter, 'description') || get(parameter, 'schema.description');
 
-  const type = get(parameter, 'schema.type', 'unknown');
+  const rootType = get(parameter, 'schema.type', 'unknown');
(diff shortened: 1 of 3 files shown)


paulatulis:
I am likely not testing this out right, but I pasted the above example verbatim under `paths/parameters` in [simpleApiWithoutDescription](https://github.com/stoplightio/elements/blob/main/packages/elements/src/__fixtures__/api-descriptions/simpleApiWithoutDescription.ts#L33), and I see the following: 
Screen Shot 2021-09-27 at 1 45 58 PM

Any suggestions for me in how I can reproduce the fix?

packages/elements-core/src/components/Docs/HttpOperation/Parameters.tsx
```diff
@@ -74,10 +76,11 @@ export const Parameter: React.FunctionComponent<IParameterProps> = ({ parameter,
   const schemaExamples = parameter.schema?.examples;
   const schemaExamplesArray = Array.isArray(schemaExamples) ? schemaExamples : [];
 
-  const validations = omitBy(
+  let validations = omitBy(
```
domagojk, line 79: hm, why `const` -> `let` change?

mmiask: No need, a leftover from some experiments. Changing it back 👍

mmiask:
> I am likely not testing this out right, but I pasted the above example verbatim under `paths/parameters` in [simpleApiWithoutDescription](https://github.com/stoplightio/elements/blob/main/packages/elements/src/__fixtures__/api-descriptions/simpleApiWithoutDescription.ts#L33), and I see the following:
> Screen Shot 2021-09-27 at 1 45 58 PM
> 
> Any suggestions for me in how I can reproduce the fix?

@paulatulis Huh, you're right. I've been testing this using `dev-portal/StoplightProject` (and it works there), but it looks like we don't pass the correct data to `API` component. I'm not sure though if I can do this quickly within this PR, since we handle data differently for `API` and `StoplightProject` and it might require some more digging 🤔 

This is what is passed to the component:
Screen Shot 2021-09-28 at 12 05 27 PM

If you folks have some ideas, I'm open for suggestions 🙏

mpodlasin:
Tested on OAS2 and OAS3 and everything works great.

Also, upgraded test suite looks very nice.

Good work! 💪

mpodlasin (approved):

PR #1789, opened 2021-09-28
fix: apiDescriptionDocument input in storybook

mpodlasin:
I've noticed that some fields in storybook are not displayed properly, which makes testing and debugging a bit more difficult (at least for me):

Before:

Screenshot 2021-09-28 at 11 32 53

After:

Screenshot 2021-09-28 at 11 31 02

### DIFF
--- a/packages/elements/src/containers/API.stories.tsx
+++ b/packages/elements/src/containers/API.stories.tsx
@@ -42,13 +42,13 @@ APIWithJSONProvidedDirectly.storyName = 'Direct JSON Input (Zoom)';
 
 export const APIWithoutDescription = Template.bind({});
 APIWithoutDescription.args = {
-  apiDescriptionDocument: simpleApiWithoutDescription,
+  apiDescriptionDocument: JSON.stringify(simpleApiWithoutDescription, undefined, 2),
 };
 APIWithoutDescription.storyName = 'API Without Description';
 
 export const APIWithInternalOperations = Template.bind({});
 APIWithInternalOperations.args = {
-  apiDescriptionDocument: simpleApiWithInternalOperations,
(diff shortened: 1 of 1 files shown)


mallachari (approved):

mmiask (approved):
Good catch!

packages/elements/src/containers/API.stories.tsx
```diff
@@ -42,13 +42,13 @@ APIWithJSONProvidedDirectly.storyName = 'Direct JSON Input (Zoom)';
 
 export const APIWithoutDescription = Template.bind({});
 APIWithoutDescription.args = {
-  apiDescriptionDocument: simpleApiWithoutDescription,
+  apiDescriptionDocument: JSON.stringify(simpleApiWithoutDescription, undefined, 2),
```
domagojk, line 45: I don't know why, but I always use `null` for this... but it's the same 🤷‍♂️ 

```suggestion
  apiDescriptionDocument: JSON.stringify(simpleApiWithoutDescription, null, 2),
```

domagojk (approved):

packages/elements/src/containers/API.stories.tsx
mpodlasin: I know, team void crew doesn't like `undefined` 😂

mpodlasin: In the other place in the file there is also `null` used, so I am changing to `null`, for consistency's sake. You win!

domagojk: hehe, no I was actually a silent spy regarding `undefined` and secretly changed `void 0` wherever they wouldnt notice 
🤫  🥷

PR #1800, opened 2021-10-04
chore: remove yalcing section

mpodlasin:
Completely removes "yalcing to platform" section.

My rationale is that `CONTRIBUTING.md` is now mostly for **external** contributors. We don't want to confuse them with `platform-internal` or `ninja` or anything like that.

Additionally `platform-internal` has a whole section on yalcing, so it's unnecessary to have this section here anyways.


Nezteb (approved):

mpodlasin:
Ah, whatever, it's not that important 😂

Issue #1805, opened 2021-10-15
x-internal operations are still visible on documentation


rolivares:
## User story.

As an API designer role person I want to mark certain operations as internally visible in order to hide those operations from public display on documentation

## Is your feature request related to a problem?

Compatibility problem related to documentation consistency

## Describe the solution you'd like

When I'm using stoplight as publishing platform is posible to me acomplish this using the extended property "x-internal", so it would nice that Webcomponent would to have the same behavior

chohmann:
Fix released in 7.3.8

PR #1830, opened 2021-10-28
fix: don't render groups that have no items

chohmann:
Addresses #1805 

If a group does not contain any items (possible if all the items are x-internal and hideInternal flag is true) do not render the group.

### DIFF
--- a/packages/elements/src/components/API/utils.ts
+++ b/packages/elements/src/components/API/utils.ts
@@ -97,21 +97,24 @@ export const computeAPITree = (serviceNode: ServiceNode, config: ComputeAPITreeC
     });
 
     groups.forEach(group => {
-      tree.push({
-        title: group.title,
-        items: group.items.flatMap(operationNode => {
-          if (mergedConfig.hideInternal && operationNode.data.internal) {
-            return [];
-          }
-          return {
-            id: operationNode.uri,
-            slug: operationNode.uri,
-            title: operationNode.name,
-            type: operationNode.type,
-            meta: operationNode.data.method,
-          };
-        }),
+      const items = group.items.flatMap(operationNode => {
+        if (mergedConfig.hideInternal && operationNode.data.internal) {
+          return [];
+        }
+        return {
+          id: operationNode.uri,
+          slug: operationNode.uri,
+          title: operationNode.name,
+          type: operationNode.type,
+          meta: operationNode.data.method,
+        };
       });
+      if (items.length > 0) {
+        tree.push({
+          title: group.title,
+          items,
+        });
+      }
     });
   }
 
--- a/packages/elements/src/components/API/__tests__/utils.test.ts
+++ b/packages/elements/src/components/API/__tests__/utils.test.ts
@@ -559,4 +559,58 @@ describe('computeAPITree', () => {
       },
     ]);
   });
+
+  it('excludes groups with no items', () => {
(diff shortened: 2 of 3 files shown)


packages/elements/src/components/API/utils.ts
```diff
@@ -97,21 +97,24 @@ export const computeAPITree = (serviceNode: ServiceNode, config: ComputeAPITreeC
...
-            id: operationNode.uri,
-            slug: operationNode.uri,
-            title: operationNode.name,
-            type: operationNode.type,
-            meta: operationNode.data.method,
-          };
-        }),
+      const items = group.items.flatMap(operationNode => {
+        if (mergedConfig.hideInternal && operationNode.data.internal) {
+          return [];
```
jgreen44, line 102: I’m curious why an empty array is returned if the if condition is satisfied, else returns an object. Two different data types? I honestly don’t know this code well enough, but that’s something that stuck out to me.

chohmann: There is a notion of marking items in your specs (i.e operations, models, etc.) as "Internal". If something is marked as internal then only members of your workspace will be able to see those items. So this is basically filtering out those internal items if the `hideInternal` flag is set (which it will be if being viewed by someone unauthenticated or not a member of the workspace). At least that's my understanding. This is my first time in this code base 😄 

As for returning an empty array, [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) flattens one level. So that empty array gets flattened into the outer array and essentially adds no items to the array.

Also, GitHub diff is weird, but basically just pulled the existing items definition out of the object so that I could check the length before pushing it onto the tree. So I didn't add or change that logic at all.

Nezteb: @jgreen44 `flatMap` essentially ignores empty arrays in return values: https://www.typescriptlang.org/play?target=99#code/MYewdgzgLgBANgS2jAvDA2gKBjA3ggEwC4YBGAXwBps9CSAmKm-YmAZiZxZIBZPbWAVn7cYANhF0YAdkmsAHHJIBOJWWGYAugG5MmUJFhgApgHcAMklhpE0AHQAzOAEMoAWWcAHABTeoACwQwAHMSUTAAVwBbACNjACdyAEpUAD48GgQHP0CQu0IYAFIYelQUNAAGFNwaHHjjKAj4sAwdGAB6dpgnVw9PGHqokAA3YwgYYyjPKABPeCsIGnJM7ICg4PyCIvYyyuragYamltxybQ6unvcvGAIQMZgwEFhBkeMJqdmYEBiAK2NgFBFjhljgsjl1pttoJdjAqhkcHUjs08GsQpRbghhkgEDE4MYAEIzABiWOMJCg8QixjOSz0SMaKNwdhZaOCGKeJgA8g4ACr+YwAQRibwpVJpmGSun04AgIHxdjgIGC3hMFisSW0QA

Some links:
* https://www.digitalocean.com/community/tutorials/list-processing-with-map-filter-and-reduce
* https://www.digitalocean.com/community/tutorials/js-flat-flatmap

jgreen44: Thanks @chohmann and @Nezteb : flatMap is new to me.

Nezteb (approved):

PR #1867, opened 2021-11-16
fix (elements-dev-portal): anchor tags stripped from uri

jgreen44:
Addresses #8606

Anchor tags were not appended to links if those links were pointing to another document. Links are now appended, but scrolling to the anchor position is not working. This seems to be a known bug when using react-router-dom:
https://github.com/remix-run/react-router/issues/394#issuecomment-220221604

To test:

Option 1:
1. Checkout branch
2. Create two articles. In one article, create a top header and a bottom header located somewhere else down the page.
2. In your second article, link to the top header and the bottom header.
3. You will notice the anchor tags are appended to the outbound links, but the scrolling to the anchor position is not working

https://user-images.githubusercontent.com/52048026/141879149-60a7b490-56f6-417c-a4bc-2d75f7090e9f.mov

Option 2:
1. Checkout branch:
2. CD into packages/elements-dev-portal and run `yarn storybook`
3. Select NodeContent/Create Todo
4. Use the following information in the Controls section to the right:

```
nodeSlug: ZG9jOjg3NDE4NA-article-two
projectId: cHJqOjExNzIxMg
platformUrl: https://dojo-stg-v2.stoplight.io
```

You will notice the links have the anchor tags, but the links are not clickable

https://user-images.githubusercontent.com/52048026/141879186-1e2103d2-a152-482e-a91c-91ff8cc99c4e.mov

### DIFF
--- a/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
+++ b/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
@@ -127,6 +127,7 @@ const LinkComponent: CustomComponentMapping['a'] = ({ children, href }) => {
     );
 
     if (edge) {
+      console.log('Edge Found ', edge);
       return <Link to={`${edge.slug}${hash && `#${hash}`}`}>{children}</Link>;
     }
   }
--- a/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
+++ b/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
@@ -127,11 +127,7 @@ const LinkComponent: CustomComponentMapping['a'] = ({ children, href }) => {
     );
 
     if (edge) {
-      return (
-        <Link to={edge.slug} hash={hash}>
-          {children}
-        </Link>
-      );
+      return <Link to={`${edge.slug}${hash && `#${hash}`}`}>{children}</Link>;
     }
   }
 
--- a/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
+++ b/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
@@ -127,7 +127,6 @@ const LinkComponent: CustomComponentMapping['a'] = ({ children, href }) => {
     );
 
     if (edge) {
-      console.log('Edge Found ', edge);
       return <Link to={`${edge.slug}${hash && `#${hash}`}`}>{children}</Link>;
     }
   }
--- a/packages/elements-dev-portal/src/containers/StoplightProject.tsx
+++ b/packages/elements-dev-portal/src/containers/StoplightProject.tsx
@@ -130,9 +130,6 @@ const StoplightProjectImpl: React.FC<StoplightProjectProps> = ({
     elem = (
       <NodeContent
         node={node}
-        // Link={({ children, ...props }) => {
-        //   return <a href={props.to}>{children}</a>;
-        // }}
         Link={ReactRouterMarkdownLink!}
         hideTryIt={hideTryIt}
         hideMocking={hideMocking}
--- a/packages/elements-dev-portal/src/containers/StoplightProject.tsx
+++ b/packages/elements-dev-portal/src/containers/StoplightProject.tsx
@@ -8,6 +8,7 @@ import {
   withQueryClientProvider,
   withStyles,
 } from '@stoplight/elements-core';
+import { ReactRouterMarkdownLink } from '@stoplight/elements-core/components/MarkdownViewer/CustomComponents/ReactRouterLink';
 import { flow } from 'lodash';
 import * as React from 'react';
 import { Link, Redirect, Route, useHistory, useParams } from 'react-router-dom';
@@ -129,9 +130,10 @@ const StoplightProjectImpl: React.FC<StoplightProjectProps> = ({
     elem = (
       <NodeContent
         node={node}
-        Link={({ children, ...props }) => {
-          return <a href={props.to}>{children}</a>;
-        }}
+        // Link={({ children, ...props }) => {
+        //   return <a href={props.to}>{children}</a>;
+        // }}
+        Link={ReactRouterMarkdownLink!}
         hideTryIt={hideTryIt}
         hideMocking={hideMocking}
         hideExport={hideExport}
--- a/packages/elements-dev-portal/src/containers/StoplightProject.tsx
+++ b/packages/elements-dev-portal/src/containers/StoplightProject.tsx
@@ -1,5 +1,6 @@
 import {
   findFirstNode,
+  ReactRouterMarkdownLink,
   RoutingProps,
   SidebarLayout,
   useRouter,
@@ -8,7 +9,6 @@ import {
   withQueryClientProvider,
   withStyles,
 } from '@stoplight/elements-core';
-import { ReactRouterMarkdownLink } from '@stoplight/elements-core/components/MarkdownViewer/CustomComponents/ReactRouterLink';
 import { flow } from 'lodash';
 import * as React from 'react';
 import { Link, Redirect, Route, useHistory, useParams } from 'react-router-dom';
--- a/packages/elements-dev-portal/src/containers/StoplightProject.tsx
+++ b/packages/elements-dev-portal/src/containers/StoplightProject.tsx
@@ -129,7 +129,9 @@ const StoplightProjectImpl: React.FC<StoplightProjectProps> = ({
     elem = (
       <NodeContent
         node={node}
-        Link={Link}
+        Link={({ children, ...props }) => {
+          return <a href={props.to}>{children}</a>;
+        }}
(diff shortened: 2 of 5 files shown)


packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
```diff
@@ -127,11 +127,7 @@ const LinkComponent: CustomComponentMapping['a'] = ({ children, href }) => {
     );
 
     if (edge) {
-      return (
-        <Link to={edge.slug} hash={hash}>
```
Nezteb, line 131: To someone more knowledgable than me: `react-router-dom` doesn't support `hash` as a top-level prop, it needs to be part of the `to` prop object: https://v5.reactrouter.com/web/api/Link/to-object

So putting everything in the `to` prop as a string might be "incorrect'? Should we fix the actual prop type here? https://github.com/stoplightio/elements/blob/6f3ae8fa1aa73466fee10dca589ef49f8a58ce2c/packages/elements-core/src/components/MosaicTableOfContents/types.ts#L10-L15

packages/elements-dev-portal/src/containers/StoplightProject.tsx
```diff
@@ -129,7 +129,9 @@ const StoplightProjectImpl: React.FC<StoplightProjectProps> = ({
     elem = (
       <NodeContent
         node={node}
-        Link={Link}
+        Link={({ children, ...props }) => {
+          return <a href={props.to}>{children}</a>;
+        }}
```
Nezteb, line 134: Neither the original code nor this `<a>` tag works correctly with scrolling. (@jgreen44 so maybe we'll revert this specific code and go back to the `react-router-dom` component.

From https://knowbody.github.io/react-router-docs/api/Link.html:
> Note: React Router currently does not manage scroll position, and will not scroll to the element corresponding to the hash. Scroll position management utilities are available in the scroll-behavior library.

The hashes are in the URL bar in Jason's PR when running locally now, but they don't scroll when clicked. We assume this is because the browser tries to resolve the anchor before the content is actually loaded, but I don't know how to fix that. Do we need to do something special to accomplish this? https://v5.reactrouter.com/web/guides/scroll-restoration/generic-solution

packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
mallachari: Not sure why the hash property was used on top level, looks like indeed it doesn't do anything there. If so we could update that type to support `to` object, though string is fine too.

packages/elements-dev-portal/src/containers/StoplightProject.tsx
mallachari: Yeah, it seems the problem is unrelated se we might get back to router link. But the scrolling itself is tricky. Could we window.scroll it to the header?

packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
mallachari: I think that prop is misleading, no need for it there in such case

jgreen44: Removed hash from custom type

packages/elements-dev-portal/src/containers/StoplightProject.tsx
Nezteb: I just messaged Jason this but figured I'd put it here for others to sanity-check me:

- Jason and I found this SO thread yesterday: https://stackoverflow.com/a/45777157
- It mentions a third party dep `react-router-hash-link`, and we were both like “we don’t want to import another dep just for this”.
- I checked, and it looks like we already pull that into elements: https://github.com/stoplightio/elements/blob/0069c4648f109fee8485e94aa999a66699bcdac1/packages/elements-core/package.json#L74
- We also apparently use it here: https://github.com/stoplightio/elements/blob/0069c4648f109fee8485e94aa999a66699bcdac1/packages/elements-core/src/components/MarkdownViewer/CustomComponents/ReactRouterLink.tsx#L3
- So I see we have this component `ReactRouterMarkdownLink ` which uses the HashLink thing from `react-router-hash-link`, which supposedly has some magic sauce that takes care of scrolling.
- So… I wonder if we can use that `ReactRouterMarkdownLink` component instead of the vanilla `Link` from `react-router-dom` here: https://github.com/stoplightio/elements/blob/0069c4648f109fee8485e94aa999a66699bcdac1/packages/elements-dev-portal/src/containers/StoplightProject.tsx#L132

mallachari: Perfect, that's it! I didn't even remember that we had it there. We should be able to use that. I've just quickly checked if that helps and looks like scrolling to anchors works fine.

jgreen44: @mallachari : Pushed latest commit with @Nezteb 's suggestion. I cannot yalc into Platform without errors. If someone could look at this code to see if it'll work.

mallachari: Sure, I'll take a look

jgreen44: @mallachari : Were you actually able to get the scrolling to the anchors working?

jgreen44:
@Nezteb , as @mallachari pointed out, the issue may not be with needing to use `react-router-hash-link`. The H1 header and H4-6 are not creating anchor links. The video shown is from STG and obviously doesn't have any of the commits from this PR.

https://user-images.githubusercontent.com/52048026/142477817-964920e8-bf4a-430b-9f05-642976483f99.mov

jgreen44:
https://user-images.githubusercontent.com/52048026/142478793-f4b912e3-be01-4a2c-bb6c-f977c2b408dd.mov

jgreen44:
@Nezteb @mallachari : Can you kindly review this thread and let me know if this PR is complete?
https://stoplight-internal.slack.com/archives/C028M2ACATY/p1637262268276600

packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
```diff
@@ -127,11 +127,8 @@ const LinkComponent: CustomComponentMapping['a'] = ({ children, href }) => {
...
-          {children}
-        </Link>
-      );
+      console.log('Edge Found ', edge);
```
mallachari, line 130: Let's not commit that. Though this is my favourite way of debugging 😉

packages/elements-dev-portal/src/containers/StoplightProject.tsx
```diff
@@ -129,7 +130,10 @@ const StoplightProjectImpl: React.FC<StoplightProjectProps> = ({
     elem = (
       <NodeContent
         node={node}
-        Link={Link}
+        // Link={({ children, ...props }) => {
+        //   return <a href={props.to}>{children}</a>;
+        // }}
```
mallachari, line 135: I think that wasn't supposed to be here too

jgreen44: removed :)

packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
jgreen44: removed :)

mallachari (approved):

Nezteb (approved):

packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
marbemac: Consumers of elements do things like integrate it into a nextjs app. NextJS has it's own router, which is why we exposed these props - so that could hook elements into most router implementations. For anybody interested, here is a proof of concept elements nextjs integration we were working on until we had to stop to work on an urgent migration project: https://github.com/stoplightio/dev-portal/blob/plugin/plugin/components/NodeLink.tsx#L9

As you can see, NextJS routing expects the hash to be passed in as a prop, which we can no longer do if it's not exposed on the `CustomLinkComponent` (see typing error below):

Screen Shot 2021-11-23 at 2 47 41 PM

packages/elements-dev-portal/src/containers/StoplightProject.tsx
```diff
@@ -129,7 +130,7 @@ const StoplightProjectImpl: React.FC<StoplightProjectProps> = ({
     elem = (
       <NodeContent
         node={node}
-        Link={Link}
+        Link={ReactRouterMarkdownLink!}
```
lottamus, line 133: The reason you had to force the typing here with `!` is because `CustomLinkComponent` is expecting `to` but `ReactRouterMarkdownLink` is expecting `href`

image


image

PR #1868, opened 2021-11-17
fix: reuse mosaic, react-query context if possible

marbemac:
If the application that elements is embedded into is already using mosaic or react-query, re-use those contexts rather than creating new ones.

Particularly applicable to platform - as it stands now we are creating excess contexts, and not respecting the options set on the parent contexts.


lottamus (approved):
Looks great! This will make it easier for folks to use Mosiac and React Query outside of the DevPortalProvider context ❤️ 

Still might be good to add a test that explicitly tests this case still renders a single MosiacProvider and QueryClientProvider?

```jsx
<MosaicProvider>
 <DevPortalProvider>...</DevPortalProvider>
</MosaicProvider>
```

paulatulis (approved):
Agree with Chris on the test, but looks good to me!

marbemac:
Done and done!

PR #1879, opened 2021-11-23
chore(e2e): wait for api to respond

lottamus:
Reduces flakey tests by waiting for API calls to return

https://docs.cypress.io/api/commands/intercept#Waiting-on-a-request


cypress/integration/api.ts
```diff
@@ -82,6 +82,10 @@ describe('API component', () => {
 });
 
 function loadZoomApiPage() {
+  cy.intercept('https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/openapi.yaml').as(
```
P0lip, line 85: eh, we could probably spawn a local server using `nock` or something.
We really don't need to rely on GitHub in this particular case, unless we're testing the integration with GH.

P0lip (approved):

cypress/integration/api.ts
lottamus: Good idea, this should do it 35dbd88c6a260637a877a080e12b4b19e7acba3c

mpodlasin (approved):

PR #1880, opened 2021-11-23
fix(dev-portal): add fetching state to search

bzmw:
Currently when you type into the search box while the http request is being completed you see "No results found".

This PR uses the react-query `isFetching` boolean to show "Fetching results..." while the query is in progress.

### DIFF
--- a/examples/react-cra/src/components/Search.tsx
+++ b/examples/react-cra/src/components/Search.tsx
@@ -9,7 +9,7 @@ export type SearchProps = {
 export const Search = ({ projectIds }: SearchProps) => {
   const [search, setSearch] = React.useState('');
   const [open, setOpen] = React.useState(false);
-  const { data } = useGetNodes({
+  const { data, isFetching } = useGetNodes({
     search,
     projectIds,
   });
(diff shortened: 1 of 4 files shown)


mpodlasin:
Merging main will probably help with failing tests. We recently did an update to make them less flakey

mpodlasin:
Thank you for the PR!

If I am being honest, I would prefer a different (granted, more complex) solution.

It would be better to show some kind of loading icon on the input itself, while keeping the results from the previous request.

This way we would avoid so much "jumping" in the results section. 🤔 

I am - ironically - finding it both distracting *and* unclear that a state changed with "text only" messaging.


https://user-images.githubusercontent.com/7916523/143565002-fd0371a7-1023-4533-96a2-93a1b5369d33.mov


What do you think?

bzmw:
Main merged.

Yeah I envisioned a better loading spinner as well but I simply didn't have time for it so I threw this together.

mpodlasin:
@mnaumanali94 what do you think? (check out the behaviour in the video I've posted in the previous comment)

bzmw:
Closing in favour of #1894

PR #1881, opened 2021-11-23
chore: use fontawesome library

lottamus:


lottamus:
Closing in favor of keeping fontawesome imports

PR #1883, opened 2021-11-24
fix(build): mjs/nextjs 12/swc compatibility

marbemac:
Adjustments such that the build output can be used in mjs environments such as NextJS v12.

- chore(demo): fix linode spec link
- fix: circular import
- fix: do not use `_.chain`, mjs compat and it breaks tree-shaking
- fix: root FA/lodash imports for mjs compat (also added an eslint rule to enforce this pattern)

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -60,7 +60,7 @@
     "@stoplight/yaml": "^4.2.2",
     "classnames": "^2.2.6",
     "httpsnippet": "^1.24.0",
-    "jotai": "1.2.2",
+    "jotai": "1.3.9",
     "json-schema": "^0.3.0",
     "lodash": "^4.17.19",
     "nanoid": "^3.1.20",
@@ -75,7 +75,7 @@
     "xml-formatter": "^2.4.0"
   },
   "devDependencies": {
-    "@stoplight/scripts": "9.1.0",
+    "@stoplight/scripts": "9.2.0",
     "@testing-library/dom": "^7.26.5",
     "@testing-library/jest-dom": "^5.11.5",
     "@testing-library/react": "^11.1.1",
--- a/packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
+++ b/packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
@@ -1,7 +1,7 @@
 import { safeStringify } from '@stoplight/json';
 import { IHttpParam, INodeExample, INodeExternalExample } from '@stoplight/types';
 import { JSONSchema7Definition, JSONSchema7Type } from 'json-schema';
-import _, { isObject, map } from 'lodash';
+import { isObject, keyBy, map, mapValues } from 'lodash';
 
 export type ParameterSpec = Pick<IHttpParam, 'name' | 'examples' | 'schema' | 'required'>;
 const booleanOptions = [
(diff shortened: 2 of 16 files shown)


packages/elements-core/package.json
```diff
@@ -60,7 +60,7 @@
     "@stoplight/yaml": "^4.2.2",
     "classnames": "^2.2.6",
     "httpsnippet": "^1.24.0",
-    "jotai": "1.2.2",
+    "jotai": "1.3.9",
```
marbemac, line 63: Had to update to `1.3.x` for esm/mjs compat.

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
```diff
@@ -81,10 +81,8 @@ const getInitialValueForParameter = (parameter: ParameterSpec) => {
 };
 
 export const initialParameterValues: (params: readonly ParameterSpec[]) => Record<string, string> = params => {
-  return _.chain(params)
```
marbemac, line 84: Chain cannot be tree-shaken, and thus the rollup lodash plugin that re-writes lodash imports for bundle size + esm compat does not work on it. We should avoid using chain.

packages/elements-dev-portal/package.json
```diff
@@ -86,5 +86,10 @@
   },
   "exports": {
     "./styles.min.css": "./styles.min.css"
+  },
+  "rollup": {
+    "bundleDeps": [
+      "use-debounce"
```
marbemac, line 92: Bundling this tiny lib because it has a messed up `esm` build.

mpodlasin (approved):
Looks good.

Do those changes lead to some immediate bundle size gains? 🤔

marbemac:
> Do those changes lead to some immediate bundle size gains?

Hmm, I actually did not think to check the bundle size - which changes stick out to you as maybe could increase bundle size?

mpodlasin:
> > Do those changes lead to some immediate bundle size gains?
> 
> Hmm, I actually did not think to check the bundle size - which changes stick out to you as maybe could increase bundle size?

I have no clue. 🤷 

Those Lodash changes? Thinking mostly about web-components bundles, where we bundle *everything*.

mpodlasin:
Oh, we misunderstood each other. By "gains" I meant... benefits. So smaller sizes 😁 

But I checked just for curiosity and.... hmm........

`main`

Screenshot 2021-11-24 at 07 02 19

`this branch`

Screenshot 2021-11-24 at 07 04 26

marbemac:
> Oh, we misunderstood each other. By "gains" I meant... benefits. So smaller sizes 😁

Ah lol, oops 😅  

I don't think this will increase final bundle size - if anything I would expect it to be more tree-shakeable. For example here we were importing the entire lodash library: https://github.com/stoplightio/elements/pull/1883/files#diff-854f88eb3b9df2be6409dc03abfcb1820bb6412aa2fffec2fe57787cd78293e1L4.

Additionally, `@stoplight/scripts` bundle command now uses a rollup plugin to transform lodash imports into deep imports (for tree shaking) with a `.js` extension (for next 12 / mjs compatibility)

Screen Shot 2021-11-24 at 1 17 18 AM

---

Now as to those built file sizes for elements package 🤔... nice catch.

It's actually because `@stoplight/scripts` is now properly bundling deps that are not defined in package.json `dependencies`. Looks like the `elements` packages imports from `@stoplight/yaml`, but does not list `@stoplight/yaml` in its package.json dependencies. I just committed a fix, check it out now.

Should be marginal diff in size of the built files - this PR might actually result in a built file that is a tiny bit larger, because each lodash import is separated into its own line. But the final JS bundle after the consumer has run webpack/next/whatever should be smaller.

mpodlasin:
Yeah, I've seen later that `web-components.min.js` is actually slightly smaller on this branch, so it seems to work nicely.

Accept is already there. ✅

PR #1884, opened 2021-11-24
chore(demo): fix GH star button

marbemac:
Just fixing the github star button on the demo page.

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -15,6 +15,7 @@
     "history": "^5.0.0",
     "react": "16.14.0",
     "react-dom": "16.14.0",
+    "react-github-btn": "1.2.1",
     "react-router-dom": "6.0.0-beta.0"
   },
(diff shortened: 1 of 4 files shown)


demo/package.json
```diff
@@ -15,6 +15,7 @@
     "history": "^5.0.0",
     "react": "16.14.0",
     "react-dom": "16.14.0",
+    "react-github-btn": "1.2.1",
```
mpodlasin, line 18: Now I know there is everything to be found on npm. 😂

mpodlasin (approved):
Tested. Looks sweet!

demo/package.json
marbemac: lol yeah, for real. TBH I didn't have time to look into why the basic script tag method was working in local but not in PR demo env or demo site... but this works so 🤷‍♂️

PR #1885, opened 2021-11-24
fix(core|dev-portal): server rendering fixes

wdalmijn:
### Description
In order to get stoplight to start rendering properly in SSR/prerender there are currently some issues in the elements repository. In a nutshell, some incorrect window defined checking, and the static router is missing which makes it harder to prerender content.

### Fixes done
- Fixed _typeof window === 'undefined'_ checking in `useLocationHash`
- Added `StaticRouter` to the `useRouter` logic
- Added `RouterType` in the global types for `elements-core`

### Note
I'll probably create another PR soon to allow for easier prefetching of data during SSR and can probably provide an example how to get SSR working using Next.JS

### Mosaic Framework issue
I found another issue in the mosaic package that I'm not able to fix myself as I guess you didn't open-source this package. Basically there's an instanceof checking doing this:
```
} else if (props.targetRef.current instanceof Node) {
```

However, `Node` only exists in the browser, so when run in node this will crash.

I created a manual (dirty) patch for the dist files which I'll paste here:
```
diff --git a/node_modules/@stoplight/mosaic/core.esm.js b/node_modules/@stoplight/mosaic/core.esm.js
index 62e8cad..cc0ecfa 100644
--- a/node_modules/@stoplight/mosaic/core.esm.js
+++ b/node_modules/@stoplight/mosaic/core.esm.js
@@ -8851,7 +8851,7 @@ function useOverlayPosition(_a) {
     if (!props.placement.includes(placement)) {
       // direction is flipped, disable main axis offset
       disableOffset = true;
-    } else if (props.targetRef.current instanceof Node) {
+    } else if (typeof window !== 'undefined' && props.targetRef.current instanceof Node) {
       const boundaryElement = props.boundaryElement || (typeof document !== 'undefined' ? document.body : null);
       const boundaryDimensions = getContainerDimensions(boundaryElement);
       const triggerDimensions = getContainerDimensions(props.targetRef.current);
diff --git a/node_modules/@stoplight/mosaic/core.umd.js b/node_modules/@stoplight/mosaic/core.umd.js
index 9be5b93..6f21d39 100644
--- a/node_modules/@stoplight/mosaic/core.umd.js
+++ b/node_modules/@stoplight/mosaic/core.umd.js
@@ -8868,7 +8868,7 @@
 	    if (!props.placement.includes(placement)) {
 	      // direction is flipped, disable main axis offset
 	      disableOffset = true;
-	    } else if (props.targetRef.current instanceof Node) {
+	    } else if (typeof window !== 'undefined' && props.targetRef.current instanceof Node) {
 	      const boundaryElement = props.boundaryElement || (typeof document !== 'undefined' ? document.body : null);
 	      const boundaryDimensions = getContainerDimensions(boundaryElement);
 	      const triggerDimensions = getContainerDimensions(props.targetRef.current);
```

However, this issue should be solved by fixing the webpack/babel config, where we make sure the target is also set to node.

Would be great if you guys could create a fix for this on your end, as I'm stuck using patches at the moment.


packages/elements-core/src/types.ts
```diff
@@ -116,6 +116,8 @@ export type Item = {
   uri: string;
 };
 
+export type RouterType = 'history' | 'memory' | 'hash' | 'static';
+
 export interface RoutingProps {
   /**
    * Only applies when using `history`-based routing. (See the `router` prop.) Specifies the base path under which
```
mpodlasin, line 123: Hey, this is a great PR!

Would you mind updating this doc description with `static` option? 🙏

mpodlasin:
Re: mosaic, it is indeed still internal.

I'll make an issue there 🙏

mpodlasin:
https://v5.reactrouter.com/core/api/StaticRouter

Hmm, now that I think about it however, this PR lacks something I believe.

StaticRouter allows you to control the route via `location` prop. But this PR doesn't allow us to pass this prop in any way to StoplightProject component? Unless I am missing something. 🤔

mpodlasin:
Re: using dev-portal with Next.js, I believe the approach we are currently taking in the company is to use those smaller components provided by dev-portal separately - `TableOfContents`, `NodeContent` etc. Each component comes with a separate data-fetching function, which - from my limited understanding of next.js - makes it easier to use there.

For example here is a storybook example of `NodeContent`. You can see there that data fetching is separate from the component itself.

https://github.com/stoplightio/elements/blob/main/packages/elements-dev-portal/src/components/NodeContent/NodeContent.stories.tsx

packages/elements-core/src/types.ts
wdalmijn: Sure.. happy to update the docs!

wdalmijn:
Re: using dev-portal with Next.JS and SSR.. I will need to make a lot of changes, with probably a bit of a more custom implementation of the router. Indeed.. locally I already have something set up like this:
```
import { useRouter as useNextRouter } from 'next/router'
// Next import is basically an override which features the code in this PR.
import { useRouter as useReactRouter } from 'vendors/@stoplight/elements-dev-portal/elements-core/hooks/useRouter'

....

const StoplightProjectRouter = ({ platformUrl, basePath = '/', router, ...props }: StoplightProjectProps) => {
  const { Router, routerProps } = useReactRouter(router ?? 'history', basePath)
  const { asPath } = useNextRouter()

  if (router === 'static') {
    routerProps.location = asPath
  }

  return (
    <DevPortalProvider platformUrl={platformUrl}>
      <Router {...routerProps} key={basePath}>
        <Route path="/branches/:branchSlug/:nodeSlug" exact>
          <StoplightProjectImpl {...props} />
        </Route>

        <Route path="/:nodeSlug" exact>
          <StoplightProjectImpl {...props} />
        </Route>

        <Route path="/" exact>
          <StoplightProjectImpl {...props} />
        </Route>
      </Router>
    </DevPortalProvider>
  )
}
```

However this is very specific for Next.JS, so respecting separations of concerns we should not have this in this router, it would be better to provide the currentpath via a param. Do you want me to update the definition of the router here so it can be provided a `currentPath` property which gets injected into the static router as a location? This probably means we'll have to drill this prop down though, and also add this to the `StoplightProjectProps`. Happy to take any suggestions on naming btw ;-). We should keep `basename` separate though, in my app I'm using this to build multiple projects in one website, which can be accessed by using the navigation bar.

Another issue I'm fixing in terms of pre-fetching data is that I'll have to remove the `withQueryClientProvider` HOCs around all of the components so I can prefetch everything in either `getStaticProps` or `getServerSideProps`. Basically I want to be able to wrap my own `<QueryClient>` from RQ around the entire app to have this control and also be able to provide the dehydrated state generated during server render.

Basically the components affected by this are:
- Search (in `elements/packages/elements-dev-portal/src/components/Search/Search.tsx`) [optional, but added so this can also be debugged in DevTools]
- DevPortalProvider (in `elements/packages/elements-dev-portal/src/components/DevPortalProvider/index.tsx`)
- StoplightProject (in `elements/packages/elements-dev-portal/src/containers/StoplightProject.tsx`)

I was wondering if there are any architectural constraints why we have multiple QueryClientProviders provided in the app. I guess this was mainly done to make everything very modular and isolated right?

I guess for these components I was going to create a PR where we also export the `Impl` versions of this, so there's more granular control for everyone on which HOCs we want to wrap around these components. This should allow me to get rid of the `withQueryClientProvider` HOC and just wrap a single `QueryClientProvider` around my entire Next.JS app in the _app.tsx file, something along the lines of this:
```
export default function MyApp({ Component, pageProps }: AppProps) {
  const [queryClient] = useState(
    () =>
      new QueryClient()
  );

  return (
    <QueryClientProvider client={queryClient}>
      <Hydrate state={pageProps.dehydratedState}>
        <Component {...pageProps} />
      </Hydrate>
    </QueryClientProvider>
  )
}
```

This should make it a lot easier to prefetch everything to a centralized QueryClientProvider, which when provided with the DevTools allows us to have a better understanding of which state our RQ queries are in.

mpodlasin:
I think the `StaticRouter` change (including adding a prop to control it) isn't very controversial, so you can continue your work here. If we support all other kinds of router I don't see any reason to not support this one. Especially since I wanted to do a PoC of improving our Gatsby support and this definitely goes in line with this goal.

Regarding the name of the prop I don't have strong feelings here, so you can propose something in the PR - `currentPath` sounds good to me. Alternatively we could do something more emphasising that it is only a prop working with static router - `staticRouterPath`? 

But - perhaps surprisingly - the next change you are proposing, exposing "naked" components *feels* a bit uneasy to me. It enlarges API surface for what is a very specific goal. I would need a bit more time to think about that, your use case etc. Not saying we will not do it, but I would like to have more time to consider this change and its implications. We are currently focusing on other goals in the company so we might not be very responsive. Could you give me at least a week to sleep on it? 🤔 

I actually will need to investigate why we are wrapping so many components in the QueryClientProvider, as you pointed out, because it is as mysterious for me as it is to you - I didn't create that part of the code 😅

mpodlasin:
BTW recently there was a change made to inherit options from a provider higher in the tree https://github.com/stoplightio/elements/pull/1868

I am just pulling this with very little context, but perhaps this would allow you to achieve your goals? 🤔 

Just a fast thought though, might be a complete miss.

mpodlasin:
Another piece of the puzzle is this - https://github.com/stoplightio/dev-portal

There we are using dev-portal components with next.js, but I think this project does also a bit more, so might not be clear/simple enough to bring you answers you seek.

wdalmijn:
@mpodlasin I think that might work actually. I'll have a try and pull the latest upstream, thanks!

I've updated the PR according to what we discussed, could you have another review?

mpodlasin:
Awesome. I'll try to look at this again today, max tomorrow 🙏

mpodlasin:
Hmmmm. The code looks good now, and yet when I am testing in storybook it does't behave as I would expect - I don't see the proper subpage of docs rendered. Did you manage to make it work in storybook?

Screenshot 2021-12-02 at 12 53 30
Screenshot 2021-12-02 at 12 54 47

mpodlasin:
Not to mention - you've adjusted one storybook story, but we have many more.

Both `elements` and `elements-dev-portal` have 2 storybooks - 1 for React and one for web-components.

All 4 should have proper controls and work properly with those controls (unless we find some specific reason why they don't).

wdalmijn:
@mpodlasin I didn't properly test the storybook. However as the static router only renders a static route a single time and doesn't support navigation I'm not sure how much added value this would have as it's designed to mainly be used for server rendering or tests. Maybe better to remove it from the Storybook altogether?

mpodlasin:
Nah, nah - for me it would be amazing if it worked in storybook.

You are correct, it locks navigation. But if I set `router="static"` and `staticRouterPath="/some/path"` in storybook, than it's reasonable to expect to see the `/some/path` page rendered, isn't it?

If it's gone from storybook, how can I test that it works properly? 🤔

wdalmijn:
@mpodlasin Well I would normally just write unit tests and test the outcome using RTL but it seems this project doesn't have many of those.

I can have a look but it would also be nice if you can point me a bit to what else needs to be updated as I don't have in-depth knowledge of this entire project and it would be bit quicker than going back and forth on these PR discussions ;).

mpodlasin:
Yeah, I get it. I'll take a look at it today and try to make it gel nicely with storybook. 👍 

For some reason it behaves weirdly. Storybook in `elements` doesn't react to `staticRouterPath`, but renders the main page at least.

On the other hand storybook in `elements-dev-portal` just renders a blank page with `router="static"`.

So it's weird. I'll take a look when I am free.

mpodlasin:
Ah ok, I think I already know. It's just storybook being crappy, I believe.

In this case, yeah I would be for writing a test.

BTW StoplightProject tests are already very close to what we want - they are using memory router.

Mocks are already setup etc., it looks fairly solid:

https://github.com/stoplightio/elements/blob/main/packages/elements-dev-portal/src/containers/StoplightProject.spec.tsx#L47

So it shouldn't be too much of a hustle to add a second test with a static router.

I can do that, but much later in the day so it's up to you if you want to try. :)

wdalmijn:
@mpodlasin I tried debugging Storybook, but first issue is that it doesn't seem to encode the params used in the stories properly, so all these slashes are getting removed. So otherwise I'll remove it from the storybook examples and I'll just create a static router test to make sure that the static router works as intended and the implementation is tested :)

I'll let you know when I've updated my branch with the test.

wdalmijn:
@mpodlasin Added the test as you suggested, cleaned up the storybook parameter to prevent further confusion.

mpodlasin (approved):
Amazing work @wdalmijn !

It was a pleasure working with you.

I'll try to make mosaic changes happen ASAP.

Good luck with your project!

mpodlasin:
@wdalmijn btw do you need this released ASAP? 🤔

PR #1893, opened 2021-11-29
chore(core): do not log example generation failures

P0lip:
This `console.error` call heavily pollutes Sentry.
image
We don't really need to call `console.error`, as in most cases it's an error on the user end. Moreover, the message is displayed to the end-user, therefore they know that an error occurred without looking into the console.


mmiask (approved):
Maybe we could do `console.warn()`/`console.log()`/`console.info()`? Basically whatever that does not pollute Sentry. I'm fine with whichever, taking into account that the error message is already displayed for users 👍

PR #1897, opened 2021-12-01
feat: re-worked content flow

marbemac:
Closes https://github.com/stoplightio/platform-internal/issues/9232. Result of collab between Plaid & Nauman.

Also yalc'd and manually verified inside of platform.

## Integrate JSV v4.4

Full release notes over here: https://github.com/stoplightio/json-schema-viewer/releases/tag/v4.4.0

Summary:

<details>
  <summary>Path crumbs bar that makes it easier to understand where a given property "lives" within the context of it's schema.</summary>
  
</details>

<details>
  <summary>Schema tree UI guidelines: These guidelines help visualize how deep one is in large schemas, and connect child properties at the same level together visually.</summary>
  
</details>

<details>
  <summary>Show more for long property descriptions.</summary>
  
</details>

## Highlight HTTP method & URL on operation pages

Make http operation method + url front and center at the top of operation pages.

<details>
  <summary>Example:</summary>
  Screen Shot 2021-12-02 at 12 19 09 AM
</details>

## Improve readability of HTTP operation pages

- Reduce the number of indented levels of UI across the board.
- Improve the consistency of spacing between elements.

An example:

<details>
  <summary>Before:</summary>
      <a href="https://elements-demo.stoplight.io/?spec=https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/openapi.yaml#/operations/accounts">https://elements-demo.stoplight.io/?spec=https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/openapi.yaml#/operations/accounts</a>

Screen Shot 2021-12-02 at 12 20 24 AM
</details>

<details>
  <summary>After:</summary>
      <a href="https://deploy-preview-1897--stoplight-elements-demo.netlify.app/?spec=https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/openapi.yaml#/operations/accounts">https://deploy-preview-1897--stoplight-elements-demo.netlify.app/?spec=https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/openapi.yaml#/operations/accounts</a>

Screen Shot 2021-12-02 at 12 20 53 AM
</details>

## Improved operation page markdown description <-> reference docs consistency

Adjustments such that an operation's reference docs (request, responses, etc) feel more consistent with longer markdown descriptions. 

An example:

<details>
  <summary>Before:</summary>
[...]

PR #1900, opened 2021-12-03
refactor: move servers dropdown + smarter persist server between re-renders

marbemac:
- TryIt: Move servers dropdown next to send
- TryIt: Remove top path/method bar when not in embedded context, since we now display that information front and center at top of operation page
- TryIt: add Mock Server as an option to servers dropdown to control whether it is enabled or not, rather than the separate enabled button in mock settings dropdown (the mock server is just another server)
- TryIt: be smarter about preserving the chosen server when changing between APIs in platform, if the URLs match up (addresses https://github.com/stoplightio/platform-internal/issues/8935)

And when a mock url is passed in:


examples/static-html/stoplight-project/index.html
```diff
@@ -8,8 +8,8 @@
     <meta name="description" content="Example site using Stoplight Elements to display docs" />
     <meta name="author" content="Stoplight" />
 
-    <script src="https://unpkg.com/@stoplight/elements-dev-portal/web-components.min.js"></script>
-    <link rel="stylesheet" href="https://unpkg.com/@stoplight/elements-dev-portal/styles.min.css" />
+    <script src="/modules/@stoplight/elements-dev-portal/web-components.min.js"></script>
+    <link rel="stylesheet" href="/modules/@stoplight/elements-dev-portal/styles.min.css" />
```
marbemac, line 12: @stoplightio/pierogi-platoon I think this problem has been here for quite a while 😅.

Basically the e2e static-html tests were operating on the published elements version rather than the built version from the PR in question etc.

mpodlasin: It's a feature, not a bug. ;)

Back in the day Phil really, really *really* wanted each folder in `examples` to be as standalone as possible, so that you could just copy it to a different directory and treat is as a starter for your project.

Therefore each package uses prod stuff, but then when we run the tests, we "swap" (in various ways) prod stuff for local stuff.

Here is a cypress intercept for unpkg stuff for example:

https://github.com/stoplightio/elements/blob/main/cypress/support/commands.js#L5

mpodlasin: angular, cra and Gatsby swap stuff in package.json on the other hand: https://github.com/stoplightio/elements/blob/main/use-local-elements.sh

marbemac: Ahh, I see. 🤔 hmm - are we sure the cypress thing works? When I spun e2e up locally for the static html example, it seemed to be pulling from prod unpkg (which was causing the tests to fail).

mnaumanali94:
It doesn't persist the server when the page is loaded for the first time 🤔

examples/static-html/stoplight-project/index.html
mpodlasin: It might not work, of course 😂 

I would honestly be happy if we just sticked to using local stuff in examples.

Having this juggling between prod stuff and local stuff made things more complex for very little in exchange.

So I think I would be okay with changes from this PR.

marbemac:
> It doesn't persist the server when the page is loaded for the first time

Didn't do local storage persist or anything, just smarter when a different API is rendered (that bug).

marbemac:
Ohh, wtf hmm - nice catch. I'll take a look

marbemac:
Should be all set!

packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
```diff
@@ -64,39 +58,37 @@ export const NodeContent = ({
   tryItCorsProxy,
 }: NodeContentProps) => {
   return (
-    <PersistenceContextProvider>
```
marbemac, line 67: @stoplightio/pierogi-platoon @mpodlasin is removing `PersistenceContextProvider` from here unsafe in any way that you can think of?

This is the root cause of https://github.com/stoplightio/elements/pull/1900#issuecomment-987265556. Basically we already wrap the entire dev portal in a `PersistenceContextProvider`, but then we're wrapping each page in it's own `PersistenceContextProvider` again here. This causes atom values accessed by children to essentially reset between pages (unless the atom is persisted somewhere other than memory, such as local storage - this is why the issue doesn't present itself for the code sample selection state).

Nezteb: How did you go about figuring out that the double context provider was causing the state to reset between pages? (Just curious for future reference)

Nezteb:
Not a code review question per se, but a general question: when someone makes a PR that addresses multiple things (four from your original PR description, only one of which I see a ticket/issue for) it makes it more difficult (or at least makes it take longer) to review.

Granted, your code is always well-commented with good tests that give me a good idea of what's going on. In general I'd like for all of us to fight for smaller PRs whenever possible. 😄

I'm not saying this PR should be split up, because I know how much work that is. I'm more just asking the question for the future!

packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
Nezteb: IMO as long as we can guarantee that any other use of `NodeContent` is wrapped with `PersistenceContextProvider`(Sourcegraph would be useful to figure that out), we're good.

That being said, I _feel_ like this would count as a breaking change *iff* it's normal for non-Stoplight repos to use this component (it's `export`'d, so who knows?) without `StoplightProject`. Any consumers of `NodeContent` would have problems.

marbemac: > How did you go about figuring out that the double context provider was causing the state to reset between pages?

Since something was up w the `chosenServer` value when navigating between previously unmounted/loaded operation pages, I started by logging that value out via good old console log. I noticed that it was resetting to its default value, and from there I had a hunch that the state library must be somehow resetting or pulling from different contexts (hunch was in part a result of running into these sorts of issues in the past w state libraries, I actually don't know much about Jotai specifically - this and the recent JSV PR is first time I've used it). Last step was to look-up how Jotai manages context, and then search codebase for that jotai provider to check it it might be double rendered... anddd, whala.

> IMO as long as we can guarantee that any other use of...

A good thought - with how things seem to be organized atm, we cannot guarantee this and there is indeed a use case to render `NodeContent` (rather than `StoplightProject`, which is basically a pre-built/easy-mode combination of `NodeContent`, `TableOfContents`, etc).

However, `DevPortalProvider` always needs to be in the tree, so that could be a good spot - @mpodlasin any particular reason the persistence provider is included via the `StoplightProject` component rather than the `DevPortalProvider`? For that matter, `StoplightProject` renders a `DevPortalProvider`, and both of them include the mosaic context and react-query context (so we're doubling up on these unnecessarily). Any reason persistence, mosaic, and react-query context can't be consolidated to `DevPortalProvider`?

marbemac:
> In general I'd like for all of us to fight for smaller PRs whenever possible. 😄
> 
> I'm not saying this PR should be split up, because I know how much work that is. I'm more just asking the question for the future!

I generally don't shy away from this size PR, for better or worse. However, I do also have plenty of small PRs. Might not be a popular opinion but I generally don't feel strongly about the PRs should be as small as possible and only do a single thing philosophy... plenty of value to managing PR complexity / risk / size, and being mindful of your reviewer's time, but there is a meaningful cost to the small PRs at all costs approach.

Also important to consider the folks involved - the experience levels of the implementor / reviewers, and should keep the likely reviewer's PR size tolerance in mind (you prefer smaller PRs, which is totally fine!).

Finally, the larger the PR, the more effort the implementor should put towards automated + manual testing, and PR description + preemptive comments calling out notable areas on the diff for reviewers. I knew https://github.com/stoplightio/elements/pull/1897 pushed the boundary re number of changes in a single PR (diff wasn't huge, but it was a beefy PR nonetheless), hence the mega PR description + a LOT of manual testing across the board - even built and yalc'd the PR into platform to test everything there.

---

Everybody has an opinion on this one, but at the very least we can all work to understand and factor in each other's PR size tolerances over time. 

My two cents 😃.

mnaumanali94:
Looking good generally. It doesn't persist for mock servers across APIs though. IMO, it should that too.

packages/elements-core/src/components/TryIt/Mocking/useMockingOptions.ts
```diff
@@ -2,5 +2,5 @@ import { atom, useAtom } from 'jotai';
 
 import { MockingOptions } from './mocking-utils';
 
-const persistedMockingOptionsAtom = atom<MockingOptions>({ isEnabled: false });
+const persistedMockingOptionsAtom = atom<MockingOptions>({});
```
mpodlasin, line 5: Very tiny thing, but it bothers me.

I would remove `persisted` prefix from here.

I believe that in other parts of the codebase we refer to things as being "persisted" if they are saved somewhere between sessions (like local storage).

For atoms we just use `somethingSomethingAtom` convention, so `mockingOptionsAtom` is a perfect name IMHO.

packages/elements-core/src/components/TryIt/Servers/ServersDropdown.tsx
```diff
@@ -0,0 +1,50 @@
...
+      title: 'Servers',
+      value: chosenServer?.url || '',
+      onChange: url => {
+        const server = servers.find(server => server.url === url);
+        setChosenServer(server);
+      },
+      children: [
+        ...servers.map((server, i) => ({
+          id: server.url,
+          title: server.name || server.description || `Server ${i + 1}`,
```
mpodlasin, line 28: There is a similar code https://github.com/stoplightio/elements/blob/main/packages/elements-core/src/utils/http-spec/IServer.ts#L11

Maybe could be unified somehow, to keep the names consistent in the whole app?

packages/elements-core/src/components/TryIt/TryIt.spec.tsx
```diff
@@ -646,10 +650,10 @@ describe('TryIt', () => {
...
       await waitFor(() => expect(fetchMock).toHaveBeenCalledTimes(1));
 
-      // disable mocking and send
-      userEvent.click(mockingButton);
-      enableItem = await screen.getByRole('menuitemcheckbox', { name: /enabled/i });
-      userEvent.click(enableItem);
+      // select regular server and send
+      userEvent.click(serversButton);
+      let server1 = screen.getByRole('menuitemradio', { name: /live server/i });
+      act(() => userEvent.click(server1));
```
mpodlasin, line 656: what is the rationale for `act` here? userEven wraps actions in `act` anyways, as far as I understand, so this should not be needed. 🤔

packages/elements-core/src/components/TryIt/TryIt.tsx
```diff
@@ -98,20 +98,37 @@ export const TryIt: React.FC<TryItProps> = ({
...
 
-  const servers = React.useMemo(
-    () => getServersToDisplay(httpOperation.servers || defaultServers),
-    [httpOperation.servers],
-  );
+  const servers = React.useMemo(() => {
+    const toDisplay = getServersToDisplay(httpOperation.servers || defaultServers);
+
+    if (mockUrl) {
+      toDisplay.push({
```
mpodlasin, line 105: Actually `getServersToDisplay` could be enhanced to return (optionally) mock server as well.

Then `ServerInfo` component could be refactored to utilise it as well. 

This would help keep all the names and options consistent throughout the codebase.

packages/elements-core/src/components/TryIt/TryIt.tsx
```diff
@@ -98,20 +98,37 @@ export const TryIt: React.FC<TryItProps> = ({
...
   const hasRequiredButEmptyParameters = allParameters.some(
     parameter => parameter.required && !parameterValuesWithDefaults[parameter.name],
   );
 
   React.useEffect(() => {
-    setChosenServer(firstServer);
-  }, [firstServer, setChosenServer]);
+    const currentUrl = chosenServer?.url;
+
+    // simple attempt to preserve / sync up active server if the URLs are the same between re-renders / navigation
```
mpodlasin, line 124: If you have to put in the comment that it's simple, then it is not simple :trollface:

packages/elements-core/src/components/TryIt/TryIt.tsx
```diff
@@ -197,81 +214,94 @@ export const TryIt: React.FC<TryItProps> = ({
...
-        </Panel.Titlebar>
+  const isOnlySendButton =
+    !httpOperation.security?.length && !allParameters.length && !formDataState.isFormDataBody && !mediaTypeContent;
 
+  const tryItPanelContents = (
+    <>
+      {httpOperation.security?.length ? (
         <TryItAuth
           onChange={setOperationAuthValue}
           operationSecurityScheme={httpOperation.security ?? []}
```
mpodlasin, line 225: Kind of awkward that we are checking if `httpOperation.security` is defined (and even not empty!), but then we have to provide a fallback anyway.

packages/elements-core/src/components/TryIt/TryIt.tsx
```diff
@@ -197,81 +214,94 @@ export const TryIt: React.FC<TryItProps> = ({
...
+          examples={mediaTypeContent.examples ?? []}
+          requestBody={textRequestBody}
+          onChange={setTextRequestBody}
+        />
+      ) : null}
+
+      <Panel.Content className="SendButtonHolder" mt={4} pt={!isOnlySendButton && !embeddedInMd ? 0 : undefined}>
+        <HStack alignItems="center" spacing={2}>
+          <Button appearance="primary" loading={loading} disabled={loading} onPress={handleSendRequest} size="sm">
+            Send API Request
```
mpodlasin, line 256: Screenshot 2021-12-07 at 12 46 54

Why we are always this name longer and longer? 😂 

We already have problem with responsiveness btw, and this just makes the problem even more apparent.

packages/elements-core/src/components/TryIt/TryIt.tsx
```diff
@@ -197,81 +214,94 @@ export const TryIt: React.FC<TryItProps> = ({
...
+        <Panel.Titlebar bg="canvas-300">
+          <Box fontWeight="bold" color={!isDark ? HttpMethodColors[httpOperation.method] : undefined}>
+            {httpOperation.method.toUpperCase()}
+          </Box>
+          <Box fontWeight="medium" ml={2} textOverflow="truncate" overflowX="hidden">
+            {`${chosenServer?.url || ''}${httpOperation.path}`}
+          </Box>
+        </Panel.Titlebar>
+
+        {tryItPanelContents}
```
mpodlasin, line 291: Hmm, since we are wrapping the content here anyways, I wonder how feasible it would be to literally define a `EmbeddableTryIt` component, that would simply use `TryIt`.

Would reduce the ifology and make `TryIt` itself more streamlined? (TryIt is already borderline out of control complex IMHO and I don't want to aggravate the situation).

packages/elements-core/src/components/TryIt/state.ts
```diff
@@ -0,0 +1,5 @@
+import { IServer } from '@stoplight/types';
```
mpodlasin, line 1: I don't know if like naming this file `state`.

It kind of suggests that this is the whole state that TryIt has, meanwhile I believe it already uses more atoms (?)

I actually like the idea of having a dedicated file where you could glance at all the state that component has (and can share with other components). But this would be a larger work and as it is right now, I believe this file name simply misinforms the reader.

I would just stick to some `chosenServer` or `chosenServerAtom` name 🤷

packages/elements-core/src/components/TryIt/Servers/ServersDropdown.tsx
mpodlasin: Looking at it second time, I find it strange that we use servers from `getServersToDisplay`, which already covers naming issues, but then we do this defensive programming for titles again. 🤔

packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
mpodlasin: I looked a bit on development history of dev-portal at it seems it just happened this way, for no particular reason.

I think moving it up the tree to `DevPortalProvider` is absolutely the most logical solution 👍

packages/elements-core/src/components/TryIt/Mocking/useMockingOptions.ts
marbemac: Completely agree! It confused me as well, I just didn't add these so wasn't sure re making name change. Will do though.

packages/elements-core/src/components/TryIt/Servers/ServersDropdown.tsx
marbemac: Can prob get rid of the third fallback `Server ${i + 1}` becaus as you pointed out this logic is already handled in IServer, will do. The rest of it though is relevant, because we're basically mapping server props to MenuItem props.

packages/elements-core/src/components/TryIt/TryIt.spec.tsx
marbemac: Likely leftover, I'll check and remove if possible.

packages/elements-core/src/components/TryIt/TryIt.tsx
marbemac: Nice one - done

packages/elements-core/src/components/TryIt/TryIt.tsx
marbemac: Don't need the fallback anymore - will remove

packages/elements-core/src/components/TryIt/TryIt.tsx
marbemac: Yeahh, responsive is a whole separate can of worms - any suggestions for right now or do we improve later? It's OK ATM until about 1450px screen width.

packages/elements-core/src/components/TryIt/TryIt.tsx
marbemac: Yeah, I had to reeeallly resist refactoring the entire TryIt component 😊. How about I re-arrange to pull relevant bits out into a `TryItRequest` component (complement to the `TryItResponse` we already have), you good with that?

packages/elements-core/src/components/TryIt/state.ts
marbemac: I don't mind either way - could collocate it with it's most relevant component (`ServerDropdown`) and ditch this file altogether? Or just re-name the file to `chosenServer` as you originally suggested - which do you prefer?

packages/elements-core/src/components/TryIt/TryIt.tsx
marbemac: Eh - it's a fairly large change (I still think would be good to do) and we'd have to re-think the state since `RequestSamples` needs access to the state from `TryItRequest`. Prob best in separate PR.

packages/elements-core/src/components/TryIt/state.ts
marbemac: Renamed to `chosenServer`

mpodlasin (approved):
🚀

Issue #1886, opened 2021-11-26
(Try to) update angular example to 13


mpodlasin:
<img width="869" alt="Screenshot 2021-11-26 at 10 49 34" src="https://user-images.githubusercontent.com/7916523/143561397-e1f54449-b04b-4eb4-ae27-f3f5f84e9419.png">

We are getting lots of noise in the repo from angular update attempts.

Would be probably good to make a single PR attempting to update the example to v13.

PR #1907, opened 2021-12-06
chore: make angular dependabot updates more rare

mpodlasin:
Relates to #1886 

Currently dependant spams us with angular 12 -> 13 updates on a weekly basis.

This PR makes them more rare.

To be reverted once #1886 is done.


paulatulis (approved):
👍

PR #1913, opened 2021-12-06
feat(article - on this page): make on this page sticky when scrolling

Nezteb:
Addresses https://github.com/stoplightio/platform-internal/issues/9049

https://user-images.githubusercontent.com/3588798/144941540-93d0b2e6-95cb-4405-85d3-ae2d6d57ef55.mov

Note: The popover's button overlapping with the title is already an issue on the `main` branch. I tried to address it but couldn't figure it out; if anyone has any ideas, let me know.

### DIFF
--- a/cypress/integration/api.ts
+++ b/cypress/integration/api.ts
@@ -48,7 +48,7 @@ describe('API component', () => {
 
     it('invokes TryIt request (receives network error)', () => {
       cy.visit('/zoom-api/operations/userCreate');
-      cy.findByRole('button', { name: /Send Request/i }).click();
+      cy.findByRole('button', { name: /Send API Request/i }).click();
       cy.findByRole('button', { name: 'Error' }).should('exist');
     });
   });
--- a/cypress/integration/stoplight-project.ts
+++ b/cypress/integration/stoplight-project.ts
@@ -50,7 +50,7 @@ describe('Stoplight component', () => {
 
     it('invokes TryIt request', () => {
       loadCreateTodoPage();
-      cy.findByRole('button', { name: /send request/i }).click();
+      cy.findByRole('button', { name: /send api request/i }).click();
 
       // Temporarily changing response code as the requested api is unavailable
       cy.findByText('500 Internal Server Error').should('exist');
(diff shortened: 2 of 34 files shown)


mpodlasin:
https://user-images.githubusercontent.com/7916523/145021020-1fb21df6-609b-4d15-96a8-7b48cba2d9e8.mov


I don't know what I am doing wrong but... it simply doesn't work for me. At all. Neither in storybook nor in examples...

And I seem to have the proper changes on my computer:


Screenshot 2021-12-07 at 12 28 18

mmiask (commented):
This does not work for me neither in Storybook nor examples. Button does not stick to the top of the page when scrolling

domagojk (commented):
It works in react-cra, but not in elements-dev-portal 😕 

From what I see, the reason is that its parent has overflow properties. When I removed  `overflowX` and `overflowY` In `SidebarLayout` it started working for me in storybook example:

Screenshot 2021-12-07 at 12 24 50

But, I guess this is here for a reason so probably not safe to just delete it.

As for the issue itself, i think they are using react-cra so it should work for them, so maybe fixing dev-portal is not needed right now? I really dont know 😬 

Hope @mpodlasin has more context?

mnaumanali94:
Not fixing it in dev portal means it wouldn't be fixed in platform, right?

domagojk:
@mnaumanali94 yes, I think it would not work in platform

Nezteb:
image

I was only testing in `react-cra` because I couldn't find any actual articles with the "on this page" ToC within storybook. So I'll need y'all to point me to what you were using for testing specifically. 😭 

Co to jest kurwa!

mnaumanali94:
Hmm. So this would work for qualtrics but leave out a fix for most of our users which I wouldn't prefer. Unless it's very hard to fix in dev portal, we should that here too.

domagojk:
😄 
>  So I'll need y'all to point me to what you were using for testing specifically.

I tried `yarn storybook` in `elements-dev-portal`

Nezteb:
It should work now!

https://user-images.githubusercontent.com/3588798/145100607-1f8c981b-a691-453f-a8b3-7bc22b914de1.mov

wmhilton (approved):
IT WORKS!

Much sticky!

image

marbemac:
The approach here doesn't quite work - TOC (sidebar) and content no longer scroll separately, as shown in the videos above, and the elements demo site no longer scrolls at all (https://github.com/stoplightio/elements/pull/1913#issuecomment-987390060).

Nezteb:
Crap. I'll try to come up with a fix.

Nezteb:
@marbemac Crap, I'll try to come up with a fix. It actually does work in `elements-dev-portal` I think, because in Will's screenshot above they scrolled down on the page and the left sidebar is still visible. The fact that my fix only works in some places is confusing to me. If you have any ideas, I'm all ears (or eyes in this case)!

PR #1924, opened 2021-12-13
fix(perf): bump mosaic to fix a rendering perf issue

marbemac:
To test:

Navigate around the zoom example in prod elements demo, note rendering performance issues: https://elements-demo.stoplight.io/

Do the same thing, but in the elements demo env associated with this PR - rendering perf issues should be gone. Example: https://deploy-preview-1924--stoplight-elements-demo.netlify.app/?spec=https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/openapi.yaml#/


mpodlasin (approved):
🐢

PR #1926, opened 2021-12-15
 fix(elements-core): removed hideExamples prop

jgreen44:
Removal of hideExamples prop allows for the examples in studio to be shown.

To test:
1. Create an API style guide in studio. 
2. In properties, type a string
3. In JSV, your example should be shown

**NOTE:** I can provide a branch to the yalc'd `elements-core` into platform if needed

image

image

image


paulatulis:
More just curious: is there a reason that prop was added to begin with?

jgreen44:
@paulatulis : Possibly this could shed some light on your question:
https://github.com/stoplightio/platform-internal/issues/8896#issuecomment-991169068

mpodlasin (approved):
👍

PR #1938, opened 2021-12-20
feat: support for optional parameters

domagojk:
Addresses: stoplightio/platform-internal#9109

I'm not sure how this should look (there is no design for it) and I've never really worked on `TryIt` component so I don't know if I'm missing something. But I played with it and it works, so I think it's at least good enough for a review from "elements people" (@mpodlasin @mallachari ) 🙂 

There is more work to be done in UI, but as I said, I'm not sure if I should even do it like this (@mnaumanali94 ?):

Screenshot 2021-12-20 at 16 41 41

### DIFF
--- a/packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
+++ b/packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
@@ -58,7 +58,7 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({ specification, value
               key={parameter.name}
               parameter={parameter}
               value={typeof value === 'string' ? value : undefined}
-              onChange={(value: string | number) =>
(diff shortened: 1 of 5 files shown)


mallachari:
That changes the design a bit. Perhaps it might be enough to just not add a param when it is optional and empty? 
I guess it's good question to @mnaumanali94

mallachari:
I've created [separate PR](https://github.com/stoplightio/elements/pull/1939) with filtering empty optional headers. 
@mnaumanali94 let's decide which one to use. I would prefer the one with filtering those params out as it doesn't add any ux features (they were mentioned by the creator but that doesn't mean we need to add them).

mallachari:
Closing as we decided to go with https://github.com/stoplightio/elements/pull/1939

PR #1939, opened 2021-12-22
fix(try-it): empty optional headers

mallachari:
Addresses: https://github.com/stoplightio/platform-internal/issues/9109

Filters out optional headers with no value provided when building request

### DIFF
--- a/packages/elements-core/src/__fixtures__/operations/put-todos.ts
+++ b/packages/elements-core/src/__fixtures__/operations/put-todos.ts
@@ -420,6 +420,13 @@ export const httpOperation: IHttpOperation = {
           },
         ],
       },
+      {
+        name: 'optional_header',
+        schema: {
+          type: 'string',
+        },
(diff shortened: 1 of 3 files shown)


paulatulis (commented):
Going to leave this to @mnaumanali94 to approve because to someone who does not know what they are doing (me the first time I looked at this), I don't see a way to explicitly indicate I don't want any value passed, but after reading the issue and the code, not typing anything is technically a clear indication you don't want the value passed. :D I guess we should assume our users understand that subtle difference. In that case, LGTM!

mallachari:
And we are already doing that for query params. It's just headers that are sent even when empty. So I just did for headers what we already have for query params. Besides, mosaic checkboxes don't look good In try it :wink:

mallachari:
@mnaumanali94 already confirmed that this approach is fine so we can go with it

paulatulis (approved):
Sounds good!

Issue #1857, opened 2021-11-12
Elements component crashed when we provide strings in examples


venkatgarlapati:
## Describe the bug
Elements component crashed when we provide examples with string values

## To Reproduce
We can reproduce this issue using [spec file](https://gist.githubusercontent.com/VenkatGarlapati/bcac966cdfa6f65acebea9f4c0849343/raw/1cee2e4ffbf520daed6e8c8d1526aae453300999/Testing.v1.yaml)
1. Open demo [application](https://elements-demo.stoplight.io/?spec=https://gist.githubusercontent.com/VenkatGarlapati/bcac966cdfa6f65acebea9f4c0849343/raw/1cee2e4ffbf520daed6e8c8d1526aae453300999/Testing.v1.yaml#/operations/get-users-userid)
2. Click on `Your GET endpoint` 
3. In parameters section `formulas` try to change `singleItem` to `multipleItems`
4. See error in console and UI

**Screenshots**

<details>
  <summary>Click to view before error!</summary>

![image](https://user-images.githubusercontent.com/28777323/141426314-438dccfb-515e-4864-ba81-102c3748f9b7.png)
</details>

<details>
  <summary>Click to view error UI and Console!</summary>

In `multipleitems` parameters, given the string value(`"ANN_R"`) and it is creating error. If i remove string it is working fine
```
 multipleItems:
          summary: Multiple Data Items
          value:
            - 'P_PRICE(0,-2,D)'
            - FG_COMPANY_NAME
            - 'FF_SEGMENT_RPT_LABELS("ANN_R",0,-1AY,BUS,SEG)'
            - 'FF_SEGMENT_RPT_DATA(ANN_R,0,-1AY,SALES,BUS,SEG)'
```

![image](https://user-images.githubusercontent.com/28777323/141426409-c6c9486c-5e9b-4fe9-b31a-17f1f9f138c2.png)
</details>

PR #1965, opened 2022-01-10
fix(try-it): escape quotes in parameter examples

mallachari:
Addresses https://github.com/stoplightio/elements/issues/1857

The reason for elements breaking was a quote in one of the examples: `FF_SEGMENT_RPT_LABELS("ANN_R",0,-1AY,BUS,SEG)` which crashes in mosaic Select due to failure in executing query selector for `data-key="{value_with_quote}"` property.

Escaping quotes (just like in JSV) makes it work. The downside is that it changes the original value. 
Another solution could be escaping it only for the Select component and resetting the original value after selecting but that might be a bit over-engineered way.


mpodlasin (approved):
I am ok with that solution.

PR #1971, opened 2022-01-13
fix: encode uri components before calling platform

daniel-white:
This fixes places where the platform URL components were only concatenated and not encoded. I separated out the encoded variables in order to maintain readability of the interpolated string


Nezteb:
My only request is to add one or more tests for this change. 🙂

daniel-white:
> My only request is to add one or more tests for this change. 🙂

on it.

mmiask (approved):
This looks great @daniel-white - I'm glad you fixed all of the other possible vulnerabilities besides the branch name/slug. Also, I appreciate the thorough testing ❤️

PR #1998, opened 2022-01-27
chore: bump http-spec

domagojk:
For: https://github.com/stoplightio/platform-internal/issues/9530

Includes this http-spec fix: https://github.com/stoplightio/http-spec/pull/170
And https://github.com/stoplightio/http-spec/pull/166 which was picked up by this update


mmiask (approved):
I used the API document attached in the related issue and it works great now. Awesome stuff 🚀

PR #2012, opened 2022-02-02
fix: bump mosaic and export core css

lottamus:
- Bumps to the latest Mosiac
- Removes some unused css
- Separates core css and mosaic css. This is useful if you're already importing mosaic's styles so you won't end up duplicating them (platform)
- Fix storybook names


mnaumanali94 (approved):
Looks good!

PR #2015, opened 2022-02-04
fix(core): gracefully handle invalid URLs in server objects

P0lip:
Fixes https://github.com/stoplightio/platform-internal/issues/9800

According to [RFC3986#3.2.2](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2), which OAS spec adheres to, `[` and `]` are not valid chars to use as a host.

I'm uncertain how to handle https://github.com/stoplightio/elements/blob/9f9ae13782edde922f4299a3db817a545095aa55/packages/elements-core/src/components/TryIt/build-request.ts#L44L45 from the UX standpoint.
If the URL is invalid, we'll end up using the current origin of the active frame elements is embedded in, which might be a potentially jarring experience. 
Either way, since this is likely uncommon to happen, we may be good to leave it as is.

### DIFF
--- a/packages/elements-core/src/utils/http-spec/IServer.ts
+++ b/packages/elements-core/src/utils/http-spec/IServer.ts
@@ -3,9 +3,15 @@ import URI from 'urijs';
 
 import { isProperUrl } from '../guards';
 
+type ServerWithOptionalUrl = Omit<IServer, 'url'> & { url: string | null } & { description: string };
+
+function isValidServer(server: ServerWithOptionalUrl): server is ServerWithOptionalUrl & { url: string } {
+  return server.url !== null && isProperUrl(server.url);
+}
+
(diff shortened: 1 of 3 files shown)


packages/elements-core/src/utils/http-spec/IServer.ts
```diff
@@ -26,15 +32,21 @@ export const getServersToDisplay = (originalServers: IServer[], mockUrl?: string
   return servers;
 };
 
-export const getServerUrlWithDefaultValues = (server: IServer): string => {
+export const getServerUrlWithDefaultValues = (server: IServer): string | null => {
```
marbemac, line 35: Making it nullable will prob have downstream consequences right? Not end of world, but we should be aware as we update in elements and elsewhere - hopefully the typings catch it.

marbemac (approved):
This would make for a good default/recommended spectral rule as well ;)

paulatulis (approved):
🥇

PR #2020, opened 2022-02-09
chore(docs): Add `apiDescriptionDocument` documentation

### LINKED ISSUE #2014: Add support for Web Component to load from a variable
## User story.

As a developer, I can use Elements with an OpenAPI spec uploaded client-side to the browser, so I do not need to publish my OpenAPI spec publicly.

The ability to specify an alternative to the `apiDescriptionUrl` which allows rendering from i.e. location(s) on disk, or uploaded to the browser.

## Additional context

Right now, this is still possible, using the web component and an HTTP server running locally in the directory the API spec, and any other file-based `$ref`s can be found in.

jamietanna:
As noted in #2014, it's unclear how to use this with the static Web
Component, although it does appear to be documented in the React
documentation.

Closes #2014.


pamgoodrich (approved):
Looks great. Thanks for contributing!

mnaumanali94 (approved):

PR #2019, opened 2022-02-09
feat(elements-dev-portal): expose cache keys

P0lip:
Needed by https://github.com/stoplightio/platform-internal/pull/10056

PR #2021, opened 2022-02-09
chore(release): elements-dev-portal@1.6.13

P0lip:


mmiask (approved):
Looks good! Even though this releases a feature - https://github.com/stoplightio/elements/pull/2019 I think it's ok to release a patch, since the feature is mostly for internal usage in Stoplight

PR #2029, opened 2022-02-10
fix(elements-core): allow trailing slash use for endpoints

jgreen44:
Addresses: [#9966](https://github.com/stoplightio/platform-internal/issues/9966)

**Summary**

Endpoints that had trailing slashes (ex. /user/) were returning a response of 422 (sometimes). This was because the trailing slash was being stripped out, ultimately pointing to either a non-existent path, or in an edge-case, to the wrong path (ex. /user). This edge-case was returning a response of 200. Allowing the use of a trailing slash now points to two paths if necessary (/user and /user/).

**Steps to Test**

1. Create API project in local or prod with both use-cases:
a. With endpoint /user
b. With endpoint /user/
2. Run storybook in elements-dev-portal
3. Copy nodeSlug, projectId from newly created API project (step 1)
4. Paste values into Storybook and use https://stoplight.io (if step 1 is prod) or https://stoplight-local.com (if step 1 is local) for platformUrl
5. Select endpoint /user
6. In the TryIt component, select Mock Server, click Send API Request (res should equal 200)
7. Select endpoint /user/
8. In the TryIt component, select Mock Server, click Send API Request (res should equal 200)
NOTE: Step 8 is the edge-case. The 200 response is because it's requesting from /user, not /user/

9. Delete endpoint /user
10. Select endpoint /user/
11. In the TryIt component, select Mock Server, click Send API Request (res should equal 200)
NOTE: Response used to return 422

### DIFF
--- a/packages/elements-core/src/components/TryIt/TryIt.tsx
+++ b/packages/elements-core/src/components/TryIt/TryIt.tsx
@@ -79,6 +79,7 @@ export const TryIt: React.FC<TryItProps> = ({
   tryItCredentialsPolicy,
   corsProxy,
 }) => {
+  TryIt.displayName = 'TryIt';
   const isDark = useThemeIsDark();
 
   const [response, setResponse] = React.useState<ResponseState | ErrorState | undefined>();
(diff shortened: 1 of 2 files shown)


wmhilton:
It works! But we should make sure that the URL that's passed into the Sample Code generator also gets the trailing slash. Right now it's still got the trailing slash stripped in all the example codes:

image

jgreen44:
Nice catch! I'll submit a fix.

wmhilton:
> Nice catch! I'll submit a fix.

oooh ooh! I think it's line 221 in buildHarRequest. Same change needed there basically

jgreen44:
@wmhilton : I introduced a bug, but somehow with an invalid URL, a 200 response was returned. Confused as to why it didn't fail. I'm fixing this bug that tests caught:
image

wmhilton (approved):
LGTM!

PR #2066, opened 2022-03-14
fix(elements-dev-portal): useGetNodes does not accept branch

P0lip:
Integrated here https://github.com/stoplightio/platform-internal/pull/10535

### DIFF
--- a/packages/elements-dev-portal/src/hooks/useGetNodes.ts
+++ b/packages/elements-dev-portal/src/hooks/useGetNodes.ts
@@ -10,22 +10,24 @@ export function useGetNodes({
   search,
   workspaceId,
   projectIds,
+  branchSlug,
   pause,
 }: {
   search: string;
   workspaceId?: string;
   projectIds?: string[];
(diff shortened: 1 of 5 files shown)


packages/elements-dev-portal/src/hooks/useGetNodes.ts
```diff
@@ -10,22 +10,24 @@ export function useGetNodes({
   search,
   workspaceId,
   projectIds,
+  branchSlug,
```
marbemac, line 13: `branch`?

marbemac (approved):
Pre-approved, but had one minor suggestion re simplifying `branchSlug` arg name to just `branch`.

packages/elements-dev-portal/src/hooks/useGetNodes.ts
P0lip: yeah, that works I think. I actually started with `branch` but then went with `branchSlug` as this is how the query param was called

Issue #2076, opened 2022-03-25
Create new non-CRA elements example templates


Nezteb:
## Summary

In https://github.com/stoplightio/elements/issues/2058 we discovered that CRA 5+ does not (and will not) work with the existing elements template repos because CRA templates cannot eject and modify their webpack configs to enable Node.js polyfills.

So we had to update our documentation to pin the `npx create-react-app` command to `4.0.3`, which isn't ideal in the long run.

We're still looking into alternatives to CRA such as:
- [Create Next App with `-e`](https://nextjs.org/docs/api-reference/create-next-app#options)
- [npm init using](https://ar.al/2021/04/01/npm-init-using/)
- [Vite/degit](https://vitejs.dev/guide/#community-templates)
- etc.

There is also https://github.com/stoplightio/dev-portal/tree/plugin but DnC didn't create it and it would need more documentation in place before suggesting it to new users.

I have [a PoC](https://github.com/stoplightio/elements-template) to test some of these other options but it's broken and needs more work.

## Tasks

- [ ] Figure out which direction we want to go (Next, Vite, npm init using, etc).
- [ ] Implement new demo repos.

aerialist7:
Hi @Nezteb,
I'm looking for using `@stoplight/elements` with Vite.js. 
But I'm facing different errors while running dev server and production build:
- global is not defined
- process.env is not defined
- Prism is not defined 
- Cannot read properties of undefined (reading 'allocUnsafe')
    at node_modules/fastestsmallesttextencoderdecoder/NodeJS/EncoderAndDecoderNodeJS.min.mjs
    
In my opinion, this is due to the lack of Node.js polyfills.
I tried to add them according to [this article](https://medium.com/@ftaioli/using-node-js-builtin-modules-with-vite-6194737c2cd2), but still doesn't work.

Do you have any example using `@stoplight/elements` with Vite.js? 
Could you please help me?

P.S. I can share my Vite.js project if that helps.

Yahkob:
just FYI the directions to use `npx create-react-app@4.0.3 my-dir --template @stoplight/elements` no longer work. CRA is a headache these days and the react team is not really supporting it (i.e. look at how many issues they have with no answers 😞) I recently tried to use and something as fundamental as using a webpack proxy no longer works

leoperria:
[...]

PR #2084, opened 2022-03-31
"Elements" Web Components Support For ESM

aerialist7:
Hi Team,

we're using `@stoplight/elements` web components in [Vite/React project](https://github.com/aerialist7/stoplight-elements-sample/tree/vite-react-web-component).
But there is unpleasant error while run `build` or `dev` commands:
```
[vite:resolve] Missing "./web-components.min.js" export in "@stoplight/elements" package
error during build:
Error: Missing "./web-components.min.js" export in "@stoplight/elements" package
```
That's [caused by Vite's strict adherence to the exports field in the package.json](https://github.com/vitejs/vite/issues/6480#issuecomment-1011937533).

So here is a fix for this problem.


aerialist7:
cc @lottamus @philsturgeon @ryotrellim @heitortsergent @marcelltoth

aerialist7:
Hi @pamgoodrich @lottamus @Nezteb @marbemac!
Can anyone please review my PR?

lottamus (changes requested):
@aerialist7 thanks for the PR. Why did you decide to use the web-components build in your React app versus the React components? The web-components are primarily meant for non-React apps and are less flexible than the React components.


For https://github.com/aerialist7/stoplight-elements-sample/blob/vite-react-web-component/src/App.tsx, instead you could do the following:

```tsx
import { API } from '@stoplight/elements'
import '@stoplight/elements/styles.min.css'
import './App.css'

function App() {
  return (
    <div className="App">
      <API
        apiDescriptionUrl="https://raw.githubusercontent.com/stoplightio/Public-APIs/master/reference/zoom/openapi.yaml"
      />
    </div>
  )
}

export default App
```

aerialist7:
Hi @lottamus, thank you for respond!

Sure, I've tried to use React component at the beginning. But the app doesn't work.

Here you can see my sample
- https://github.com/aerialist7/stoplight-elements-sample/tree/vite-react-component

Run `dev` or `preview` commands and open browser console - ```Uncaught ReferenceError: require is not defined```

Then I've tried to use it as Web Component

aerialist7:
Hi @lottamus, can you please answer?

lottamus:
@aerialist7 I've tried your Vitejs example and see it's not working out of the box. I'm not as familiar with Vite and esbuild, so I'll need to investigate to see what config options are needed to support Elements. Ideally it's just supported out of the box 😞 

Initial findings:
- Seems related to `require is not defined`: https://github.com/vitejs/vite/issues/3409
- Might need to add a `commonjs` plugin targeting Elements: https://vitejs.dev/guide/dep-pre-bundling.html#the-why

aerialist7:
@lottamus, thank you! 

It would be great and we hope that it will be supported out of the box!
But now we can't use even Web Component variant without hacks. 
So could you please merge my PR and release the new version of Stoplight "Elements"?

aerialist7:
Hi @lottamus, can you please answer?

lottamus:
@aerialist7 I've been able to successfully run `npm run dev` on your branch using the following vitejs config:

```js
import { defineConfig } from "vite";
import react from "@vitejs/plugin-react";
import { NodeGlobalsPolyfillPlugin } from "@esbuild-plugins/node-globals-polyfill";
import NodeModulesPolyfills from "@esbuild-plugins/node-modules-polyfill";

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [react()],

  optimizeDeps: {
    esbuildOptions: {
      plugins: [
        NodeGlobalsPolyfillPlugin({
          buffer: true,
          process: true,
        }),
        NodeModulesPolyfills(),
      ],
    },
  },
});
```

Make sure to install the two additional dev deps `@esbuild-plugins/node-globals-polyfill` and `@esbuild-plugins/node-modules-polyfill`.

Now that `dev` is working, I have a better hunch on why the prod build `npm run build && serve -s dist` is not working, but I've not yet come to the right config. I'm still hitting this issue: https://github.com/vitejs/vite/issues/2672

aerialist7:
@lottamus I applied your configuration [here](https://github.com/aerialist7/stoplight-elements-sample/tree/vite-react-component)

Thank you, `dev` works, but with problems:
# unable to select anything in the sidebar
# warnings in the browser console (mb related to first problem)

Yep, It doesn't work for me in `prod` either.
Crashes in runtime with the following error:
```
Uncaught ReferenceError: require is not defined
    at index.4675c23e.js:768:3139
    at index.4675c23e.js:1:644
    at index.4675c23e.js:1044:71129
```

marcelltoth (approved):
I don't have write access here anymore but I believe that this change is fundamentally correct, and the fact that the React distribution does or does not work with Vite should be a separate topic. (Tracked in an issue,)

@lottamus We explicitly support using the web components distribution, then why are you so hesitant to approve a PR that makes the statement `import '@stoplight/elements/web-components.min.js'` legal under the Node ESM spec? Would you like to restrict web components usage to direct script tag usage for some reason?

Now for the bug itself, I think that should be handled separately, but in the meantime I think I have a workaround for @aerialist7: try adding it the old-school way, using a script tag in the `<head>`. That does work I believe, even when using React, why wouldn't it.

aerialist7:
@lottamus, any news? 
May be we really can continue merge this PR and continue our discussion within separate issue? (e.g. #2076)

aerialist7:
@lottamus

lottamus (approved):

aerialist7:
@lottamus, thank you!
When we can expect a new release?

aerialist7:
@lottamus

PR #2106, opened 2022-04-11
feat: use http-spec v5

P0lip:
Tbh I don't get why we store `IHttpOperation` instead of OAS documents. This feels weird to me. 
We should be generating these fixtures on the fly using the latest http-spec we have rather than having some pre-hardcoded artifacts that are prone to change anyway.


philsturgeon:
I believe the goal of using IHttpOperation instad of OAS directly is that Elements does not want to be tied to OAS directly. Just like Prism actually works with Postman Collections, this could too. Could support OASv4 someday which hopefully isn't (but could be) rather different, and we'll have to keep making it more and more and more abstract to handle the superset of functionality. Seems a little wild when its mostly OASv3 most of the time, but that's the design choice that was made.

P0lip:
> I believe the goal of using IHttpOperation instad of OAS directly is that Elements does not want to be tied to OAS directly. Just like Prism actually works with Postman Collections, this could too. Could support OASv4 someday which hopefully isn't (but could be) rather different, and we'll have to keep making it more and more and more abstract to handle the superset of functionality. Seems a little wild when its mostly OASv3 most of the time, but that's the design choice that was made.

Yeah, I got it that elements doesn't want to be tied to plain OAS, as one may support other specs at some point.
What I was thinking here was that instead of storing these plain `IHttpOperation` objects, one could just do

```ts
const document = {
  openapi: '3.1.0',
  //
};

export const httpOperation = transformOperation({ document, path: '/user', method: 'get' });
```

In any case, it's too late now.

PR #2113, opened 2022-04-14
feat(dev-portal): prep stable ids + custom paths

marbemac:
Small adjustments to prep for the upcoming stable IDs + custom paths release.

### DIFF
--- a/packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx
+++ b/packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx
@@ -230,7 +230,7 @@ const Node = React.memo<{
   onLinkClick?(): void;
 }>(({ item, depth, meta, onClick, onLinkClick = () => {} }) => {
   const activeId = React.useContext(ActiveIdContext);
-  const isActive = activeId === item.id;
+  const isActive = activeId === item.slug || activeId === item.id;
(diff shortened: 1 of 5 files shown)


packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx
```diff
@@ -257,7 +257,7 @@ const Node = React.memo<{
       className="ElementsTableOfContentsItem"
     >
       <Item
-        id={getHtmlIdFromItemId(item.id)}
+        id={getHtmlIdFromItemId(item.slug || item.id)}
```
lottamus, line 260: Can you double check this still works with the scroll logic? It looks like it will: https://github.com/stoplightio/elements/blob/2b7f277fe401c7cb4644a46e4bf5e1c1ea698025/packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx#L39

lottamus (approved):
Looks good assuming tests pass once the API is updated 👍

packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx
marbemac: I did already, works :).

PR #2148, opened 2022-05-17
fix: remove rtl

lukearmstrong:
The box containing the URL for the endpoint was behaving strangely when trying to select it to copy and paste it. It wasn't possible to select the entire URL, you could only select the "Server URL" part, or the "Path" part.

It wasn't a big deal, then I noticed it wasn't displaying my "Server URL" correctly:
image

Notice the hyphen at the start? This is what my Server URLs look like in my OpenAPI v3.1 Definition:
```
servers:
    - url: "https://europe-west2-example.cloudfunctions.net/production-"
      description: Production
    - url: "https://europe-west2-example.cloudfunctions.net/staging-"
      description: Staging
    - url: "https://europe-west2-example.cloudfunctions.net/testing-"
      description: Testing
    - url: "https://europe-west2-example.cloudfunctions.net/mock-"
      description: Mock
    - url: "http://localhost:8888/development-"
      description: Development
```

We use Google Cloud Functions and unfortunately the only way to have different environments is to deploy the same functions to different paths with different environment variables, also the path cannot contain `/` so we can't make it look like a sub-directory either. So each of the Server URLs ends in a hyphen, which for some reason if you set the html element to be `dir="rtl"` will move that hyphen to the front.

Removing `dir="rtl"` it now looks like this:
image

Removing `dir="rtl"` also now means you can select the entire URL string. It's still a bit annoying because you can't triple click the address as it is split across two elements, but at least now you can select it all.

Maybe in the future a nice feature would be to add a "Copy" button next to the URL.


P0lip:
We need that `dir="rtl"` for the string truncation to work properly.
Without `dir` set to `rtl`, we trim off the leading part of the server url instead of the trailing one.
Currently:
image

After the change:
image

Nonetheless, I do agree with you that the particular behavior you suffer from is indeed an unwanted side effect we need to address.
Let me know if we can do something about that truncating.

P0lip:
One could do
```
 {chosenServerUrl ? (
        <Box
          color="muted"
          style={{ unicodeBidi: 'bidi-override' }}
          dir="rtl"
          fontSize="lg"
          textOverflow="truncate"
          overflowX="hidden"
        >
          {chosenServerUrl.split('').reverse().join('')}
        </Box>
      ) : null}
```

to address the visual defect, but I don't really like that solution, because the text is actually reversed, so things like copy are broken. This probably has some negative impact on SEO and such, but I'm not an expert here.
image
image

P0lip:
Once again for the PR!
I went with a slightly different fix in https://github.com/stoplightio/elements/pull/2217.
It should address your issue, while preserving the existing behavior.

PR #2160, opened 2022-06-03
feat(core): support the choice of server variables

danieleds:
This PR adds support for setting the value of the server variables.

Screenshot 2022-06-03 at 18 58 28

-----


danieleds:
Hi @dotslashderek. Anything I can do to get this under review?

dotslashderek (changes requested):
Hey Daniel - apologize for the delay in reviewing this - code looks good - just had one comment and a request for some test coverage.

One thought - doesn't this functionality already exist to a certain extent?  It's not a variable that you can alter on the fly, but you can include a list of servers to use in the API spec itself, and then select the one you want to use from:

https://elements-demo.stoplight.io/#/operations/get-todos-id

If you have multiple servers defined in your spec file like in this one:

https://github.com/stoplightio/studio-demo/blob/master/reference/todos/todo.v1.yaml

You get that drop-down next to "Send API Request" so you can change the server you want to TryIt / cURL against, etc... and then all the paths change based on that selection.  

Screen Shot 2022-08-12 at 1 31 31 AM

packages/elements-core/src/utils/http-spec/IServer.ts
```diff
@@ -57,3 +61,28 @@ export const getServerUrlWithDefaultValues = (server: IServer): string | null =>
 
   return stringifiedUrl.endsWith('/') ? stringifiedUrl.slice(0, -1) : stringifiedUrl;
 };
+
+export const getServerUrlWithFilledVariables = (server: IServer, values?: { [key: string]: string }): string | null => {
```
dotslashderek, line 65: You can probably get rid of the old getServerUrlWithDefaultValues function if we add this - calling this new one and only passing in the server should be identical to calling the old one, right?  

There are only a couple other files with references to that old function that need to be updated to call this one - build_request.ts and IServer.spec.ts.  

Should update those tests anyhow to add coverage for getServerUrlWithFilledVariables.

ashubham:
@dotslashderek This functionality does not exist. For cases where there is no fixed API server. A user may have their own private installation at a custom domain, which we do not want to add to the list of `servers` in the api spec.

packages/elements-core/src/components/TryIt/Servers/Servers.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <Panel defaultIsOpen>
+      <Panel.Titlebar>Server</Panel.Titlebar>
+      <Panel.Content className="ParameterGrid">
+        {Object.entries(chosenServer.variables).map(([key, value]) => (
+          <React.Fragment key={key}>
+            <div>{key}</div>
+            <Text mx={3}>:</Text>
+            <Flex flex={1}>
+              <Input
```
dotslashderek, line 27: If the spec provides an enum of valid values for a specific variable, we should expose a dropdown select populated with those enum values, rather than an Input.

dotslashderek (changes requested):

packages/elements-core/src/utils/http-spec/IServer.ts
```diff
@@ -57,3 +61,28 @@ export const getServerUrlWithDefaultValues = (server: IServer): string | null =>
...
+  let urlString = server.url;
+
+  const variables = Object.entries(server.variables ?? {});
+  variables.forEach(([variableName, variableInfo]) => {
+    urlString = urlString.replace(`{${variableName}}`, values?.[variableName] || variableInfo.default);
```
P0lip, line 70:
```suggestion
    urlString = urlString.replaceAll(`{${variableName}}`, values?.[variableName] || variableInfo.default);
```
I believe this should be `replaceAll` in case one makes use of variable a number of times.

P0lip: `getServerUrlWithDefaultValues` has the same bug in there too - we could fix it while at it.

packages/elements-core/src/utils/http-spec/IServer.ts
P0lip, line 65: do we actually need that util at all? 
It does the same as `getServerUrlWithDefaultValues`, with the only difference being we optionally take values and apply them.
Feels like we could rename `getServerUrlWithDefaultValues` to `getServerUrlWithValues`, add `values`, read them, and that's it.

packages/elements-core/src/components/TryIt/chosenServerVariables.ts
```diff
@@ -0,0 +1,4 @@
+import { atomWithStorage } from 'jotai/utils';
+
+// track null separately from undefined so that we can tell if the server has been set (undefined indicates it has not been "processed" yet)
```
P0lip, line 3: this comment doesn't apply here

packages/elements-core/src/components/TryIt/chosenServerVariables.ts
```diff
@@ -0,0 +1,4 @@
...
+export const chosenServerVariablesAtom = atomWithStorage<{ [key: string]: string }>('TryIt_chosenServerVariables', {});
```
P0lip, line 4:
```suggestion
export const chosenServerVariablesAtom = atomWithStorage<Record<string, string>>('TryIt_chosenServerVariables', {});
```

mnaumanali94:
@danieleds Thank you for this! I'm sure elements users would appreciate this a lot. I'm thinking this information about server variables would also be useful in the API overview section just below the list of URLs. We can add info on parameters , their type e.t.c

danieleds:
Thanks for the suggestions! I can't work on this anymore I am afraid, but please feel free to update my PR accordingly :)

mnaumanali94:
@danieleds Thanks for you work on this. Server variables support has been added using #2389 Feel free to try it out and let us know what you think.

PR #2177, opened 2022-06-21
feat: use new guzzle code sample

philsturgeon:
The pecl 1 and 2 extensions are not what people are using in PHP world, and tbh I've never seen them in the wild in over a decade writing PHP. Apparently they're onto version 3 and 4 now wanyway.

Just look at any "best HTTP clients" article and you'll see no mention of the pecl extensions.

- https://www.twilio.com/blog/5-ways-to-make-http-requests-in-php
- https://ourcodeworld.com/articles/read/674/top-7-best-php-http-client-libraries

Let's use Guzzle, and if any more of these popular HTTP clients pop up on httpsnippet we can add those too.


mnaumanali94 (approved):

philsturgeon:
It's loading the curl example because the httpsnippet library has not released a new version since September 2021. 

image

Let's see if they reply to my latest offer of help. https://github.com/Kong/httpsnippet/issues/185.  otherwise we'll have to update our fork and use that (which is bloody ironic as I was asking we get rid of more of our forks). https://github.com/stoplightio/httpsnippet

P0lip (approved):

PR #2187, opened 2022-07-01
feat(elements-dev-portal): add compact option to NodeContent component

lottamus:
**elements-core**
- Adds a `layoutOptions.compact` prop to `Docs` to control when `Model` is in single vs two column layout
- Adds a copy button to `Model` examples

**elements-dev-portal**
- Adds a `compact` option to `NodeContent` to be passed down to the `Docs` component above

```ts
/**
 * Provide a number to trigger compact mode when the component is within that pixel width,
 * or a boolean to enable or diable compact mode.
 * @default false
 * @example 600
 */
compact?: number | boolean;
```

---

### DIFF
--- a/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
+++ b/packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
@@ -5,6 +5,7 @@ import {
   MockingProvider,
   ReferenceResolver,
 } from '@stoplight/elements-core';
+import { DocsProps } from '@stoplight/elements-core/components/Docs';
 import { CustomComponentMapping } from '@stoplight/markdown-viewer';
 import { dirname, resolve } from '@stoplight/path';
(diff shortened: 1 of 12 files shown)


packages/elements-dev-portal/src/components/NodeContent/NodeContent.tsx
```diff
@@ -5,69 +5,52 @@ import {
   MockingProvider,
   ReferenceResolver,
 } from '@stoplight/elements-core';
+import { DocsProps } from '@stoplight/elements-core/components/Docs';
```
marbemac, line 8: should we export this typing from root of core?

lottamus: Yeah good idea, done in b5f425b339af79a62366c12070eb8e1a9c2aea26

marbemac (approved):

PR #2191, opened 2022-07-05
fix(elements-dev-portal): some properties are optional for NodeContent

lottamus:
After #2187, the docs properties that node content we're unexpectedly made required when they should still be optional

image


lottamus:
Nvm, this is working as expected. My vscode needed a reset

image

PR #2217, opened 2022-08-12
fix(elements-core): display operation path correctly at all times

P0lip:
Aside from a fix, I also did it so that a single click selects the whole url.

https://user-images.githubusercontent.com/9273484/184356301-73afd1b5-c8ff-49d5-9691-04837e65b9d8.mp4


mnaumanali94 (approved):

dotslashderek (approved):

lukearmstrong:
@P0lip I've only just seen this linked to my other issue, this is brilliant, thanks for fixing the problem and making it easier to select the URL. Our api docs site just uses the simple script tag method pointing at the latest version, so it has updated for us and the fix works perfectly :)

P0lip:
I'm glad it worked for you! :)

PR #2218, opened 2022-08-15
chore(internal): introduce nodeHasChanged option

marbemac:
A new `nodeHasChanged` option to support the visual diff work. For now this is expected to be leveraged primarily by platform - there is not really an external use case atm. 

Unobtrusive & low risk - the newly added code effectively does nothing when the `nodeHasChanged` option is not provided. The biggest risk in this PR is simply bumping mosaic and JSV - there's always a little risk there.

Related: 

- JSV (done): https://github.com/stoplightio/json-schema-viewer/pull/204
- See screenshots for how this behaves in practice in this platform PR: https://github.com/stoplightio/platform-internal/pull/12693

Areas that need to support node change annotations:

- [x] JSV everywhere
- [x] Model title
- [x] Model description
- [ ] ~Model example panel~ skipping for the time being
- [x] Service title row
- [x] Service tags row (which includes version)
- [x] Service servers panel
- [x] Service security panel
- [x] Service additional info panel
- [x] Operation title row (which includes internal + deprecated)
- [x] Operation method + url row
- [x] Operation description
- [x] Operation request description
- [x] Operation security panel
- [x] JSV usage within operation (just make sure it's working as expected in all the spots e.g. request query/headers/etc, response headers)
- [ ] ~Operation responses tabs~ skipping for the time being
- [x] Operation response description


packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.tsx
```diff
@@ -34,28 +36,28 @@ const HttpOperationComponent = React.memo<HttpOperationProps>(
...
+        path={data.path}
+        noHeading={layoutOptions?.noHeading}
+        hasBadges={hasBadges}
+        name={prettyName}
+        isDeprecated={isDeprecated}
+        isInternal={isInternal}
+      />
     );
 
+    const descriptionChanged = nodeHasChanged?.({ nodeId: data.id, attr: 'description' });
```
marbemac, line 52: These `nodeHasChanged` calls are effectively noops for all existing users of elements (since it's expected that platform is really the only consumer of elements that leverages the new `nodeHasChanged` prop).

packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.tsx
```diff
@@ -34,28 +36,28 @@ const HttpOperationComponent = React.memo<HttpOperationProps>(
...
     const description = (
       <VStack spacing={10}>
-        {data.description && <MarkdownViewer className="HttpOperation__Description" markdown={data.description} />}
+        {data.description && (
+          <Box pos="relative">
+            <MarkdownViewer className="HttpOperation__Description" markdown={data.description} />
+            <NodeAnnotation change={descriptionChanged} />
```
marbemac, line 58: Ditto on these `NodeAnnotation` components - they just return `null` if `change` is undefined, which it will be for all existing usage - so it does nothing - these changes overall are quite low risk / unobtrusive.

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
```diff
@@ -15,45 +17,76 @@ import { ServerInfo } from './ServerInfo';
...
-          {(data.contact?.email || data.license || data.termsOfService) && (
-            <AdditionalInfo contact={data.contact} license={data.license} termsOfService={data.termsOfService} />
-          )}
-        </Box>
-      </VStack>
-      {data.description && <MarkdownViewer className="sl-my-5" markdown={data.description} />}
-    </Box>
-  );
-});
+const HttpServiceComponent = React.memo<HttpServiceProps>(
```
marbemac, line 20: Prettier decided to indent this more which caused a large diff here - in reality very little has changed.

packages/elements-core/src/components/Docs/HttpService/HttpService.tsx
```diff
@@ -15,45 +17,76 @@ import { ServerInfo } from './ServerInfo';
...
-          <VersionBadge value={data.version} />
-        </Box>
-      )}
-      {pathname && layoutOptions?.showPoweredByLink && (
-        <PoweredByLink source={data.name ?? 'no-title'} pathname={pathname} packageType="elements" layout="stacked" />
-      )}
-      <VStack spacing={6}>
-        <ServerInfo servers={data.servers ?? []} mockUrl={mocking.mockUrl} />
-        <Box>
-          {data.securitySchemes?.length && (
```
marbemac, line 43: I fixed a little bug here - one should never `&&` off of a numeric value when rendering react stuff. When the value is `0` it ends up rendering the `0`...

Screen Shot 2022-09-14 at 3 27 44 PM

packages/elements-core/src/constants.ts
```diff
@@ -47,6 +50,7 @@ export const NodeTypeIconDefs: Dictionary<IconDefinition, NodeType> = {
   table_of_contents: faQuestionCircle,
   spectral_ruleset: faQuestionCircle,
   styleguide: faQuestionCircle,
+  image: faImage,
```
marbemac, line 53: Required due to the `@stoplight/types` version bump - typings complained.

packages/elements-core/src/context/Options.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const useOptionsCtx = () => {
+  return React.useContext(ElementsOptionsContext) || DEFAULT_CONTEXT;
+};
+
+export type ProviderProps = Partial<ElementsOptionsContextProps> & {
+  children: React.ReactNode;
+};
+
+export function ElementsOptionsProvider({ children, nodeHasChanged }: ProviderProps) {
```
marbemac, line 19: Decided to add an options provider to avoid having to pass top level options through the entire component tree everywhere. Just using it for `nodeHasChanged` via `const { nodeHasChanged } = useOptionsCtx();` at the moment, but we can easily use this for other options in the future to cut down on the manual prop passing.

packages/elements-core/src/utils/ref-resolving/resolvedObject.ts
```diff
@@ -80,4 +80,5 @@ export const getOriginalObject = (resolvedObject: object): object => {
   return resolvedObject[originalObjectSymbol] || resolvedObject;
 };
 
-const isReference = (value: unknown): value is { $ref: string } => isObject(value) && typeof value['$ref'] === 'string';
+export const isReference = (value: unknown): value is { $ref: string } =>
```
P0lip, line 83: tbh one could use `hasRef` from `@stoplight/json`. It does the same

packages/elements-core/.storybook/preview.jsx
```diff
@@ -13,7 +13,9 @@ const ThemeProvider = (Story, context) => {
     subscribeTheme({ mode: theme === 'dark' ? 'dark' : 'light' });
   }, [subscribeTheme, theme]);
   return (
-    <Story {...context} />
+    <Box style={{maxWidth: 1200, }} mx='auto' py={10}>
```
P0lip, line 16:
```suggestion
    <Box style={{ maxWidth: 1200 }} mx='auto' py={10}>
```
nit, but my eyes always hurt when I see such stuff haha

packages/elements-core/src/components/Docs/HttpService/AdditionalInfo.tsx
```diff
@@ -1,16 +1,21 @@
...
+  id?: string;
   termsOfService?: string;
   contact?: IHttpService['contact'];
   license?: IHttpService['license'];
 }
 
-export const AdditionalInfo: React.FC<AdditionalInfoProps> = ({ termsOfService, contact, license }) => {
+export const AdditionalInfo: React.FC<AdditionalInfoProps> = ({ id, termsOfService, contact, license }) => {
+  const { nodeHasChanged } = useOptionsCtx();
+  const hasChanged = nodeHasChanged?.({ nodeId: id || '', attr: ['termsOfService', 'contact', 'license'] });
```
P0lip, line 17: I'm a bit on the edge with it. Let's perhaps make `id` required?
I assume the `id` we pass here comes from `IHttpService`, so it should be always there.

P0lip (commented):
Looks good overall.
I'll take another look in the morning.

packages/elements-core/.storybook/preview.jsx
marbemac: Ha - I specifically saved without auto-formatting because I didn't want to increase the diff. I'll just format the file, I agree it hurts my eyes too.

packages/elements-core/src/context/Options.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+
+import type { DocsProps } from '../components/Docs';
+
+const DEFAULT_CONTEXT: ElementsOptionsContextProps = {};
+
+export type ElementsOptionsContextProps = Pick<DocsProps, 'nodeHasChanged'>;
+
+export const ElementsOptionsContext = React.createContext<ElementsOptionsContextProps | undefined>(undefined);
```
P0lip, line 9:
```suggestion
export const ElementsOptionsContext = React.createContext<ElementsOptionsContextProps>(DEFAULT_CONTEXT);
```

packages/elements-core/src/context/Options.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const useOptionsCtx = () => {
+  return React.useContext(ElementsOptionsContext) || DEFAULT_CONTEXT;
```
P0lip, line 12:
```suggestion
  return React.useContext(ElementsOptionsContext);
```

P0lip: wouldn't that work ^?

packages/elements-core/src/components/Docs/HttpService/ServerInfo.tsx
```diff
@@ -25,25 +27,25 @@ export const ServerInfo: React.FC<ServerInfoProps> = ({ servers, mockUrl }) => {
...
+            ))}
+          </VStack>
+        </Panel.Content>
       </Panel>
     </InvertTheme>
   );
 };
 
-const ServerUrl = ({ description, url, marginBottom = true }: IServer & { marginBottom?: boolean }) => {
+const ServerUrl = ({ id, description, url }: IServer & { marginBottom?: boolean }) => {
```
P0lip, line 42: not sure if intended from the UI standpoint or not, but you lost the default value for `marginBottom`.
It was `marginBottom = true` in the prior version

P0lip (approved):
Left a few remarks, but the code looks good.

packages/elements-core/src/context/Options.tsx
marbemac: This is to cover the "just in case" the context provider was not used to wrap the area.

packages/elements-core/src/components/Docs/HttpService/ServerInfo.tsx
marbemac: Hmm i actually don't even see where it was used?

P0lip: let's remove it altogether than haha

packages/elements-core/src/context/Options.tsx
P0lip: Then, the default would still be used no?

P0lip: assuming it's set here https://github.com/stoplightio/elements/pull/2218#discussion_r971883500

marbemac: No if code calls `useOptionsCtx` and the tree has not been wrapped in the context provider, `React.useContext(ElementsOptionsContext)` will return undefined/null.

Issue #2205, opened 2022-07-26
Resolving links and urls


caitlinnew:
We are integrating with Stoplight Elements (https://meta.stoplight.io/docs/elements/ZG9jOjMyNjU4OTY0-elements) and we are having some issues with the links/anchors inside the page. The layout we are using is “sidebar”, the router is “hash”. 

## Current Behavior
The problem is that when I click on a link heading I get redirected on the correct place, but the URL goes to the default one.
The link I’m clicking on - http://127.0.0.1:8887/#gatewayresponse
The link I end up on - http://127.0.0.1:8887/#/

Would you be able to help us resolve this, since with this functionality it’ll be hard for the users to share the specific part of the documentation that is linked to?

## Steps to Reproduce
[6035Bug.zip](https://github.com/stoplightio/elements/files/9190191/6035Bug.zip)

SF 6035

pavinduLakshan:
We are having the same problem and it annoys our end users. A solution to overcome this problem is highly appreciated.

lazartsarev:
Hey, any update on this one?

mnaumanali94:
I can reproduce to it in https://elements-demo.stoplight.io/ 

Verify it works here once the fix is made.

PR #2220, opened 2022-08-19
fix(elements): keep unresolvable hashes when hash router is used

P0lip:
Fixes https://github.com/stoplightio/elements/issues/2205

When `HashRouter` is used, hashes are (obviously) used for navigation.
This poses an issue, as h2/h3/h4 headings may also contain anchors, which are supposed to navigate us to a particular fragment of the page.
Currently elements redirects you to root (`/`) if a given route has no node associated.
Due to the nature of `HashRouter`, both the heading anchors and the links adhere to the same mechanism, thus any link heading that's not associated with a subpage, will be redirected to `/`
Unfortunately, we cannot quite tell whether a user wants to navigate to a particular subpage or simply has an anchor, so what I ended up was: if we cannot associate any node with a particular hash, we redirect to `/` and keep the initial hash.
This way the navigation works correctly. That's presumably what 99% of hash router users would expect to happen.

### DIFF
--- a/packages/elements/src/components/API/APIWithSidebarLayout.tsx
+++ b/packages/elements/src/components/API/APIWithSidebarLayout.tsx
@@ -3,6 +3,7 @@ import {
   Logo,
   ParsedDocs,
   PoweredByLink,
+  RoutingProps,
   SidebarLayout,
   TableOfContents,
 } from '@stoplight/elements-core';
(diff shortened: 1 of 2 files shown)


eadamsatx (approved):

eadamsatx:
Great PR description. Logic and behavior both seem sound.

PR #2231, opened 2022-09-05
feat: object and array query parameters in try it

durasj:
# Support for query parameters of type object and array in Try It

Currently, array and objet query parameters are not properly used by the Try It component. Here is a handy table (from [swagger.io](https://swagger.io/docs/specification/serialization/
)) showing how they should be processed depending on the `style`: 

style | explode | URI template | Primitive value id = 5 | Array id = [3, 4, 5] | Object id = {"role": "admin", "firstName": "Alex"} | Covered by this PR
-- | -- | -- | -- | -- | -- | --
form * | true * | /users{?id*} | /users?id=5 | /users?id=3&id=4&id=5 | /users?role=admin&firstName=Alex | ✅
form | false | /users{?id} | /users?id=5 | /users?id=3,4,5 | /users?id=role,admin,firstName,Alex | ✅
spaceDelimited | true | /users{?id*} | n/a | /users?id=3&id=4&id=5 | n/a | ✅
spaceDelimited | false | n/a | n/a | /users?id=3%204%205 | n/a | ✅
pipeDelimited | true | /users{?id*} | n/a | /users?id=3&id=4&id=5 | n/a | ✅
pipeDelimited | false | n/a | n/a | /users?id=3\|4\|5 | n/a | ✅
deepObject | true | n/a | n/a | n/a | /users?id[role]=admin&id[firstName]=Alex | ❌

\* Default serialization method


durasj:
https://user-images.githubusercontent.com/1062898/188497244-186ed9b6-f067-4b83-99e0-ab0ec0b9849a.mp4

packages/elements-core/src/components/TryIt/build-request.ts
```diff
@@ -50,6 +50,68 @@ const getServerUrl = ({
   return serverUrl;
 };
 
+const getQueryParams = ({
```
P0lip, line 53: Would you mind adding some tests specifically for this function?
The table you added is great, but I'm afraid further contributors may potentially forget about it in case a change is needed.

LMK if you have any questions.

durasj: Sure thing! I was trying to match the code style around and noticed you don't write unit tests for individual files and functions. Having unit tests here would be lovely as there is a clear specification that the function needs to conform to.

durasj: Done. Let me please know if this is what you were expecting. I've checked coverage and it's basically 100% covered. I've noticed you don't use `test:` prefix for commits so I've used `chore:` here, but I guess you'll squash my commits anyway.

P0lip: Yeah, we'll squash the commits anyway, so it doesn't matter that much.

Thanks!

P0lip (approved):
Code looks good. Thanks!

P0lip:
There's something off with the CircleCI check not running - will look into it shortly, and then going to merge it.
Once again, thanks for the PR. Nice work.

durasj:
Thanks, @P0lip!

Issue #2260, opened 2022-10-09
Responses tab list cuts response codes on small screens


porotkin:
## Context

Unable to see a full list of response codes.

## Current Behavior

Response codes  are not fully displayed as shown in the picture below:
<img width="1718" alt="image" src="https://user-images.githubusercontent.com/34237389/194743206-6dfa7321-cd2b-4cef-93cc-18c508728e96.png">

## Expected Behavior

<img width="1724" alt="image" src="https://user-images.githubusercontent.com/34237389/194743243-7e17307c-7c80-495f-acf3-15383886f64f.png">

## Possible Workaround/Solution

Allow response tabs to wrap around the container so the information is fully displayed.

## Steps to Reproduce

1. Open "@stoplight/elements" storybook
2. Go to GitHub story
3. Navigate to `Create a blob` operation with a `git` tag
4. Open the `Controls` sidebar of storybook
5. See the cut response codes of `Responses` component

## Environment

- Version used: 7.7.2
- Environment name and version (e.g. Chrome 39, node.js 5.4): Chrome Version 105.0.5195.125, Node.js Version 16.17.1
- Operating System and version (desktop or mobile): macOS Monterey 12.6, desktop

porotkin:
#2261

PR #2261, opened 2022-10-09
fix(elements-core): display response codes in full on any screen (#2260)

porotkin:
This PR makes all the response codes in a `TabList` visible so that the user can get a full information even on a small screen.


porotkin:
Related issue #2260 
@P0lip, gladly ask you to review, hope this helps

Amjcraft:
@porotkin How about the approach instead https://github.com/stoplightio/elements/pull/2265?

porotkin:
#2265 is a better way indeed. I believe the issue can be resolved

PR #2061, opened 2022-03-10
In flux changes. Switches to the load more codeViewer built in util, …

Amjcraft:
TryIt was getting stuck with very large responses (due to CodeViewer's highlighting on the responses).  To fix this, the CodeViewer renders the raw response (without highlighting/styling) to not overload the page.  Code is only rendered raw when the response is over 10,000 lines.  
Raw Render (more than 10,000 lines):
image
Highlighted Render (less than 10,000 lines):
image

# Elements Default PR Template

In general, make sure you have: (check the boxes to acknowledge you've followed this template)

- [x] Read [`CONTRIBUTING.md`](../CONTRIBUTING.md)

#### Other Available PR Templates:

- Release: https://github.com/stoplightio/elements/compare?template=release.md
  - [x] [Read the release section of `CONTRIBUTING.md`](../CONTRIBUTING.md#releasing-elements)

PR #2266, opened 2022-10-24
fix: try it large responses

mallachari:
Addresses: https://github.com/stoplightio/platform-internal/issues/11697
Continuation of https://github.com/stoplightio/elements/pull/2061

Large TryIt responses are causing a bug that prevents response component to rerender correctly making old response content persist (despite the `response` state set with new data). 

I wasn't able to find the reason for that, even going deep into mosaic CodeViewer component - new props are passed and can be accessed correctly.

Solution to that issue (not the greatest one) is resetting the `response` state ~~after invoking request~~ before setting actual response, which causes the CodeViewer to unmount and render again once new response is set. ~~That causes slight change in the ui making old response disappear on invoking new request. That's something that could be considered as either feature or bug.~~

**Before**

https://user-images.githubusercontent.com/14196079/197600613-66609301-7188-4686-9ecd-68c57aed8ab6.mov

**After**

https://user-images.githubusercontent.com/14196079/197600687-37d0ad87-45ae-4843-b772-cfc4552bdc94.mov

### DIFF
--- a/packages/elements-core/src/components/TryIt/TryIt.tsx
+++ b/packages/elements-core/src/components/TryIt/TryIt.tsx
@@ -206,7 +206,7 @@ export const TryIt: React.FC<TryItProps> = ({
         const bodyText = type !== 'image' ? await response.text() : undefined;
         const blob = type === 'image' ? await response.blob() : undefined;
 
-        setResponse(undefined);
+        setResponse(undefined); // setting undefined to handle rendering large responses
         setResponse({
           status: response.status,
           bodyText,
--- a/packages/elements-core/src/components/TryIt/TryIt.tsx
+++ b/packages/elements-core/src/components/TryIt/TryIt.tsx
@@ -193,21 +193,21 @@ export const TryIt: React.FC<TryItProps> = ({
         credentials: tryItCredentialsPolicy,
         corsProxy,
       });
-      let response1: Response | undefined;
+      let response: Response | undefined;
       try {
-        setResponse(response);
-        response1 = await fetch(...request);
+        setResponse(undefined);
+        response = await fetch(...request);
       } catch (e: any) {
         setResponse({ error: new NetworkError(e.message) });
       }
-      if (response1) {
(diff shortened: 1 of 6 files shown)


kaylachun:
Does it work to setResponse(response) after response1 is received?

kaylachun (approved):
Looks good to me, I guess there is a slight concern for a different UI experience as the response is cleared until the new one is rendered.  But I think that will only be noticeable for larger responses like this one.

mallachari:
Right, I don't like this way of solving bugs, but couldn't find any better. Maybe we could use it only for these kind of large responses and have default behaviour for common ones.
@Amjcraft what do you think of it?

mallachari:
I've updated it and put right before setting actual response. This way it still keeps previous response so there's no ui change.

mallachari:
https://user-images.githubusercontent.com/14196079/197599202-9d03fd8a-7820-4a29-83b0-a1485e981a49.mov

packages/elements-core/src/components/TryIt/TryIt.tsx
```diff
@@ -203,10 +203,14 @@ export const TryIt: React.FC<TryItProps> = ({
...
+        const bodyText = type !== 'image' ? await response.text() : undefined;
+        const blob = type === 'image' ? await response.blob() : undefined;
+
+        setResponse(undefined); // setting undefined to handle rendering large responses
         setResponse({
           status: response.status,
-          bodyText: type !== 'image' ? await response.text() : undefined,
-          blob: type === 'image' ? await response.blob() : undefined,
+          bodyText,
+          blob,
```
Amjcraft, line 213: Can you explain to me why this helps? 

I can kind of see how moving the awaits up, and  blocking the state update until finished could help. The `setResponse(undefined)` not so much. From my understanding React should just "batch" the 2 setResponse calls together and just render once.

Amjcraft (approved):
LGTM

PR #2274, opened 2022-11-02
fix(core): use param delimiter for arrays in TryIt

P0lip:
https://github.com/stoplightio/platform-internal/issues/13653

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
   build:
     working_directory: /mnt/ramdisk/project
     docker:
-      - image: cimg/node:lts
+      - image: cimg/node:16.17
     resource_class: large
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
   build:
     working_directory: /mnt/ramdisk/project
     docker:
-      - image: cimg/node:lts
+      - image: cimg/node:16.17
```
P0lip, line 27: downgraded due to https://app.circleci.com/pipelines/github/stoplightio/elements/8184/workflows/10807cca-a884-425f-893f-4a51bc7701c8/jobs/34378.
We'll need to address it at some point

matty-v (approved):

PR #2281, opened 2022-11-08
feat: response tabs horizontal scrolling

mallachari:
Addresses https://github.com/stoplightio/platform-internal/issues/10150

Adds horizontal scrolling for http operation responses.

I'm not adding `x-overflow` property directly to mosaic because:
- it has some side effects with wrapping whitespaces 
- mosaic is supposed to handle overflowing tabs by switching them to select component (though it may require good bit of work)
- elements response tabs is the only case when we need that

https://user-images.githubusercontent.com/14196079/200484922-62e122e6-a5af-4b6c-9380-4e4ce280b941.mov

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -10,7 +10,7 @@
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
-    "@stoplight/elements": "^7.7.3",
+    "@stoplight/elements": "^7.7.5",
     "@stoplight/mosaic": "^1.33.0",
     "history": "^5.0.0",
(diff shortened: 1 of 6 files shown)


mmiask (approved):
Overall looks good. One thing I'm wondering is if it should display the entirety of the last response tab or should it be cut off. That way there would be some indication of the possibility of scrolling.

mmiask:
I'm seeing that it displays scrolling bar at all times for smaller screens. Ignore the above comment. It's good to go

Issue #2238, opened 2022-09-14
Schema size error while example rendering


teluren:
Dear colleagues,
I'm trying to generate the example preview over OpenAPI 3.0.x specification and getting an error:
>Example cannot be created for this schema
>Error: Schema size exceeded

The API response body for this example contains an array of objects with such a structure:
``` YAML
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: "#/components/schemas/Quote"
```
This Quote object is rather big - about 80 parameters, some of them with nested objects. The whole list is near the 250 parameters.
If I'm rendering only a single Quote schema object, an example generates correctly, but for the collection in the endpoint view falls. 

In smaller specifications, this array example works well. 
E.g., for this case:
``` YAML
  /user:
    get:
      summary: Get users list
      operationId: getUser
      responses:
        "200":
          description: User Created
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: "#/components/schemas/User"
```
with this body:
``` YAML
components:
  schemas:
    User:
      title: User
      type: object
      properties:
        id:
          type: integer
          description: Unique identifier for the given user.
        firstName:
          type: string
        lastName:
          type: string
        email:
          type: string
          format: email
        dateOfBirth:
          type: string
          format: date
          example: '1997-10-31'
        emailVerified:
          type: boolean
          description: Set to true if the user's email has been verified.
        createDate:
          type: string
          format: date
          description: The date that the user was created.
      required:
        - id
        - firstName
        - lastName
        - email
        - emailVerified
```
Elements generates this example:
``` JSON
[
  {
    "id": 0,
    "firstName": "string",
    "lastName": "string",
    "email": "user@example.com",
    "dateOfBirth": "1997-10-31",
    "emailVerified": true,
    "createDate": "2019-08-24"
  }
]
```

Could you, please help, how can I increase this limit/schema size?

smitabh84:
Hi Same issue with my GUI as well, the example as below 
Well, here is an example of a payload for the blueprint simple_launch API endpoint. I don't know that the payload will look like for launching a marketplace item. That's the purpose of this ticket.


Payload for blueprint simple_launch api endpoint - (json)

{
    "spec": {
        "app_profile_reference": {
            "kind": "app_profile",
            "name": "Default",
            "uuid": "64b87432-97f7-403c-8ee9-ba7cebea13ac"
        },
        "app_description": "Test of api provisioning",
[...]

PR #2283, opened 2022-11-10
fix: increase tick count for json schema sampler

Amjcraft:
#2238 

Fix for 2238, large example schemas throwing "Exceeds schema size error massage". Fix should allow for larger sized response and example schemas. 

The fix was to increase the tick count for the json schema sampler. By default the tick count is set to 1000. For elements I am now passing in a count of 6000. For context the schema the kicked off this issue has around 30,000 lines and requires around 4500 ticks. 

Impacts. Ticks are used as a count ceiling for performance and dose not increase execution time. Added by https://github.com/stoplightio/json-schema-sampler/pull/22 . It appears ticks might not be a great indicator of time/performance. For context the sampler execution timer is exponential but only averages around 8ms for the large schema mentioned above. I think we need a followup ticket to rethink/rework this mechanism in the json schema sampler (https://github.com/stoplightio/json-schema-sampler/issues/33)

I'm using SB here with the schema provided by the client to test.

https://user-images.githubusercontent.com/2607104/201147121-d83e1f88-da14-4970-ada9-cd4109c22a2b.mov

### DIFF
--- a/packages/elements-core/src/utils/exampleGeneration/exampleGeneration.ts
+++ b/packages/elements-core/src/utils/exampleGeneration/exampleGeneration.ts
@@ -16,7 +16,7 @@ export type GenerateExampleFromMediaTypeContentOptions = Sampler.Options;
 export const useGenerateExampleFromMediaTypeContent = (
   mediaTypeContent: IMediaTypeContent | undefined,
   chosenExampleIndex?: number,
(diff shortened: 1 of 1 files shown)


eadamsatx (approved):
I like your solution. I agree that ticks is a very weird measure.

dotslashderek (approved):

Issue #2310, opened 2023-01-10
Circular references stoped working after #2305


Thore1954:
## Context

After the release of version 7.7.8 my document freezes when opening an schema containing circular references.

## Current Behavior

Freezes when opening schemas containing circular references.

## Expected Behavior

Should not freeze when opening schemas containing circular references.

## Possible Workaround/Solution

Revert to version 7.7.7 until it's fixed.

## Steps to Reproduce

1. Open an schema containing circular references.
2.
3.
4.

## Environment

- Version used: 7.7.8
- Environment name and version (e.g. Chrome 39, node.js 5.4): Any browser
- Operating System and version (desktop or mobile): Any OS
- Link to your environment/workspace/project:

Thore1954:
@P0lip Would you mind looking into this?

Thore1954:
Here you go:
```yaml
openapi: 3.1.0
info:
  title: Tic Tac Toe
  description: |
    This API allows writing down marks on a Tic Tac Toe board
    and requesting the state of the board or of individual squares.
  version: 1.0.0
tags:
  - name: Gameplay
paths:
  # Whole board operations
  /board:
    get:
      summary: Get the whole board
      description: Retrieves the current state of the board and the winner.
      tags:
        - Gameplay
      operationId: get-board
      responses:
        '200':
          description: 'OK'
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/status'

  # Single square operations
  /board/{row}/{column}:
    parameters:
      - $ref: '#/components/parameters/rowParam'
      - $ref: '#/components/parameters/columnParam'
    get:
      summary: Get a single board square
      description: Retrieves the requested square.
      tags:
        - Gameplay
      operationId: get-square
      responses:
        '200':
          description: 'OK'
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/mark'
        '400':
          description: The provided parameters are incorrect
          content:
            text/html:
              schema:
                $ref: '#/components/schemas/errorMessage'
              example: 'Illegal coordinates'
    put:
      summary: Set a single board square
      description: Places a mark on the board and retrieves the whole board and the winner (if any).
      tags:
        - Gameplay
      operationId: put-square
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/mark'
      responses:
        '200':
          description: 'OK'
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/status'
        '400':
[...]

PR #2312, opened 2023-01-11
fix(elements-core): circular $refs in schemas no longer work

P0lip:
Fixes https://github.com/stoplightio/elements/issues/2310


wmhilton (approved):
makes sense

PR #2318, opened 2023-01-23
fix(elements-core): improve the server dropdown in TryIt

P0lip:
Fixes https://github.com/stoplightio/platform-internal/issues/14906


mnaumanali94 (approved):
Works like a charm

dotslashderek (approved):
Looks good!

PR #2328, opened 2023-03-01
fix: elements-examples studio-demo Todo API

mnaumanali94:


wmhilton (approved):

P0lip (changes requested):
We have more broken examples, i.e. https://github.com/stoplightio/elements/blob/39b2f3fd8b53391ad9493d42a8db947425f02014/packages/elements/src/web-components/__stories__/Api.stories.tsx#L34


I'd say we should just upload the files here if we don't have any todo

wmhilton:
@P0lip I was just about to suggest that as a followup!

wmhilton:
Yes, lets move the files we care about into the stoplightio/elements repo. That way it won't have any weird external dependency.

P0lip (approved):

P0lip:
> That could be an issue if we ever want to update it.

hmm yeah, but hopefully it doesn't change too often.

Issue #2323, opened 2023-02-10
Using stoplight elements dev portal with hash router throws a not found


mnaumanali94:
As a user using the hash router I get a 404 when using the secondary Nav on the right side. 

Note: this works fine with elements but not the dev portal

## Environment
Latest dev portal

lewis-benson-tbol:
I am also seeing the same issue when working with Stoplight Elements Dev Portal - clicking the secondary navigation bar on the right hand side is replacing the stoplight ID and tag, rather than adding a # onto the end.

PR #2329, opened 2023-03-02
fix(dev-portal): set RouterTypeContext

P0lip:
Related to https://github.com/stoplightio/elements/issues/2323
Needs elements-core@7.0.11 https://github.com/stoplightio/elements/pull/2330

This doesn't fix TOC. It only fixes headings such as the one in the screenshot below
image


wmhilton (approved):
LGTM

PR #2331, opened 2023-03-03
fix(elements-core): toc links yield 404 when using hash router

P0lip:
Fixes the issue with toc navigation in #2323

### DIFF
--- a/packages/elements-core/src/components/Docs/Article/index.tsx
+++ b/packages/elements-core/src/components/Docs/Article/index.tsx
@@ -2,16 +2,22 @@ import { IMarkdownViewerProps } from '@stoplight/markdown-viewer';
 import { Box } from '@stoplight/mosaic';
 import { withErrorBoundary } from '@stoplight/react-error-boundary';
 import * as React from 'react';
+import { useLocation } from 'react-router-dom';
 
+import { RouterTypeContext } from '../../../context/RouterType';
 import { MarkdownViewer } from '../../MarkdownViewer';
 import { DocsComponentProps } from '..';
 
 type ArticleProps = DocsComponentProps<IMarkdownViewerProps['markdown']>;
 
 const ArticleComponent = React.memo<ArticleProps>(({ data }) => {
+  const { pathname } = useLocation();
+  const routerKind = React.useContext(RouterTypeContext);
+  const basePath = routerKind === 'hash' ? `#${pathname.split('#')[0]}` : '';
+
   return (
     <Box className="sl-elements-article">
-      <MarkdownViewer className="sl-elements-article-content" markdown={data} includeToc />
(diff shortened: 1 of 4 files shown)


packages/elements-core/src/components/Docs/Article/index.tsx
```diff
@@ -2,16 +2,22 @@ import { IMarkdownViewerProps } from '@stoplight/markdown-viewer';
...
+import { RouterTypeContext } from '../../../context/RouterType';
 import { MarkdownViewer } from '../../MarkdownViewer';
 import { DocsComponentProps } from '..';
 
 type ArticleProps = DocsComponentProps<IMarkdownViewerProps['markdown']>;
 
 const ArticleComponent = React.memo<ArticleProps>(({ data }) => {
+  const { pathname } = useLocation();
+  const routerKind = React.useContext(RouterTypeContext);
+  const basePath = routerKind === 'hash' ? `#${pathname.split('#')[0]}` : '';
```
Amjcraft, line 16:
```suggestion
  const basePath = routerKind === 'hash' ? `${hash}` : '';
```

packages/elements-core/src/components/Docs/Article/index.tsx
```diff
@@ -2,16 +2,22 @@ import { IMarkdownViewerProps } from '@stoplight/markdown-viewer';
...
+import { useLocation } from 'react-router-dom';
 
+import { RouterTypeContext } from '../../../context/RouterType';
 import { MarkdownViewer } from '../../MarkdownViewer';
 import { DocsComponentProps } from '..';
 
 type ArticleProps = DocsComponentProps<IMarkdownViewerProps['markdown']>;
 
 const ArticleComponent = React.memo<ArticleProps>(({ data }) => {
+  const { pathname } = useLocation();
```
Amjcraft, line 14: pathname dose not return hash I believe. react-router docs says useLocations returns a hash prop.
https://github.com/remix-run/history/blob/main/docs/api-reference.md#location

Didn't test this out on our version though.

```suggestion
  const { hash } = useLocation();
```

packages/elements-core/src/components/Docs/Article/index.tsx
Amjcraft: Also just noticed the `#` before the pathname. Was that a typo or am I missing somthing?

P0lip: no, that's expected.
We use double hashes for hash router.

image

http://localhost:63342/elements/a.html#/cyrh2afdvwnpb-test#baz

the first part is the actual route, while the other hash is the anchor, hence the split.
We split to get the route and append the anchor.

Amjcraft: I see. Thanks for the clarification

Amjcraft (approved):
LGTM

PR #2366, opened 2023-05-25
fix(core): reset the auth in tryit to a supported scheme

daniel-white:
when switching between different operations that have distinct security schemes, the try it box would retain an invalid scheme. this simply resets it instead of trying to remember all instances of it.


P0lip (approved):
LGTM.
Don't forget to bump the version in package.json for the release to take place :)

PR #2383, opened 2023-08-07
fix(elements-core): markdown ref resolving

mallachari:
Fixes resolving refs in markdown files.

`Docs` component uses `InlineRefResolverProvider` referencing parsed document. That works fine with json/yaml files but in case of markdown it results in undefined document and resolver (since markdown file is not a plain object). We need to point to the schema document instead of whole markdown file. 

This PR adds internal `InlineRefResolverProvider` to `CodeComponent` to handle schemas within markdown files.

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.7.19",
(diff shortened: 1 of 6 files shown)


mmiask (approved):
Awesome work!

Issue #2386, opened 2023-08-11
Server Variable support in 'Try It' Section


mnaumanali94:
### **Description of Work**
Allow users to dynamically adjust server variables within the 'Try It' section of the documentation. 

### **User Story**
As a user, I want to be able to change server variables within the 'Try It' section, so that I can see and interact with adjusted server URLs and example code samples based on different variable values.

### **Requirements**

- [ ] Introduce input fields for each server variable within the 'Try It' section:
  - These fields should be located above the Parameters section.
  - Each input field should be labeled with the server variable's name.
  - The input fields should change if the server is changed
  

- [ ] Dynamic URL and code sample adjustment:
  - Modifying the value within the server variable input field should automatically update the displayed URL and the example code samples in real-time.
  - The default value for each server variable should be pre-filled in the input field when the 'Try It' section is accessed.

- [ ] Handle server variables with enums:
  - If a server variable has defined allowed values, replace its input field with a dropdown list containing the allowed values.
  - The dropdown should have the default value already selected upon accessing the 'Try It' section.

### **Documentation**
- Not needed

### **Design**
- [ ] This work needs to be reviewed by the Design Team. Add them to the PR as reviewers.

### **Test Plan**
- Ensure server variables can be modified in the 'Try It' section and that they reflect correctly in the URL and example curl command.
- Confirm the behavior of server variables with defined enums in the 'Try It' section.

- [ ] Update existing tests

### **Monitoring & Metrics**
- Not needed

mnaumanali94:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/-Tequila-Mockingbirds-61f945bb2ae66f0010ca5eca/issues/stoplightio/elements/2386?planning-poker) with Zenhub @Amjcraft @billiegoose @cjlanzo @dotslashderek @matty-v @P0lip

billiegoose:
Alright, I yalced it into platform-internal and it looks good.

I'll bump the versions and release it

PR #2389, opened 2023-08-17
feat(elements-core): server variables in TryIt

billiegoose:
#2386


packages/elements-core/src/utils/http-spec/__tests__/IServer.spec.ts
```diff
@@ -1,42 +1,20 @@
...
         url: 'https://[env].stoplight.io/v1',
       };
 
-      expect(getServerUrlWithDefaultValues(server)).toBeNull();
+      expect(getServerUrlWithVariableValues(server, {})).toBe('https://[env].stoplight.io/v1');
     });
   });
 
   describe('getServersToDisplay', () => {
-    it('should filter out server objects containing invalid URLs', () => {
```
P0lip, line 18: imho we could keep that as http-spec doesn't do any filtering

P0lip (approved):

PR #2393, opened 2023-08-29
fix: handle keys which have uncommon url paths, such as `/entries/{id}/data)`

weyert:
Fixes the issue when a OpenAPI specification file has a url such as `/entries/{id}/data)` the `findMapMatch`-function
throws a `invalid regular expresison` error.

### DIFF
--- a/packages/elements/src/utils/oas/index.ts
+++ b/packages/elements/src/utils/oas/index.ts
@@ -134,7 +134,9 @@ function computeChildNodes(
 function findMapMatch(key: string | number, map: ISourceNodeMap[]): ISourceNodeMap | void {
   if (typeof key === 'number') return;
   for (const entry of map) {
-    if (!!entry.match?.match(key) || (entry.notMatch !== void 0 && !entry.notMatch.match(key))) {
+    const escapedKey = key.replace(/([()[{*+.$^\\|?])/g, '\\$1');
+
+    if (!!entry.match?.match(escapedKey) || (entry.notMatch !== void 0 && !entry.notMatch.match(escapedKey))) {
       return entry;
     }
   }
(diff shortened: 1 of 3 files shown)


packages/elements/src/utils/oas/index.ts
```diff
@@ -134,7 +134,9 @@ function computeChildNodes(
 function findMapMatch(key: string | number, map: ISourceNodeMap[]): ISourceNodeMap | void {
   if (typeof key === 'number') return;
   for (const entry of map) {
-    if (!!entry.match?.match(key) || (entry.notMatch !== void 0 && !entry.notMatch.match(key))) {
+    const escapedKey = key.replace(/([()[{*+.$^\\|?])/g, '\\$1');
```
P0lip, line 137: The capturing group is a bit redundant in this case

```suggestion
    const escapedKey = key.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&')
```

weyert: Ah makes sense. Tested your suggestion locally and it still appears to behave as expected (not throwing). 

Only I am not 100% sure that this change can somehow cause regression but if I understand the code correctly the `entry.match()` and `entry.notMatch()` function are checking if some requirements defined are met

weyert:
Added test case

P0lip (commented):
Could you also bump version in package.json? This way the change will get automatically released. Example commit: https://github.com/stoplightio/elements/commit/b769cf50261f0bfa1ee7f50eaec5ad411a8d50e8

P0lip (approved):

weyert:
> Could you also bump version in package.json? This way the change will get automatically released. Example commit: [b769cf5](https://github.com/stoplightio/elements/commit/b769cf50261f0bfa1ee7f50eaec5ad411a8d50e8)

Done 👍

weyert:
Can I also bring to your attention my issue I raised for your `json-schema-viewer`-package? I have a branch locally to add vendor extensions and wanted to discuss my change before raising the PR: https://github.com/stoplightio/json-schema-viewer/issues/236

P0lip:
That's more of a question for @mnaumanali94. For me it looks fine, you'd only need to filter out some properties we already consume in some way like `x-internal` or `x-deprecated`, and most likely `x-stoplight` which doesn't contain anything relevant.

PR #2398, opened 2023-08-31
feat: responsive http operation

mallachari:
Adding responsiveness to Docs `HttpOperation`:
- Adds `compact` property and usage of `useIsCompact` hook to `HttpOperation`
- Moves `TryIt` panel to left column for compact layout
- Adds modal instead of tab responses for compact layout

https://github.com/stoplightio/elements/assets/14196079/cf20cac7-07b0-4158-9213-9a56c4737cea

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.9.0",
+  "version": "7.9.1",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**",
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.9.1",
+  "version": "7.10.0",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**",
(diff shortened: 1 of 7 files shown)


kaylachun (approved):
Awesome, this looks and works great!

paulatulis (approved):
Looks great, Jakub!

packages/elements-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "7.9.0",
+  "version": "7.9.1",
```
mmiask, line 3: I'm not strongly opinionated about this but I believe this is a `MINOR` change rather than `PATCH`. It was never stated that `Elements` design is responsive, and while this might not be a typical feature, it still improves the usage.

mmiask (approved):
Tentatively approving. I only left a comment regarding the versioning for release, but leaving it up to you to decide. Other than that it looks great 👍

packages/elements-core/package.json
mallachari: I agree and disagree in the same time, but also don't have strong opinion on that. Maybe we could have it as minor and all other responsive related releases treat as patches.

mallachari: I've make it a minor version change for elements-core and elements. Elements-dev-portal has minor version bumped already by @kaylachun with search component.

PR #2404, opened 2023-09-11
fix(elements-core): clearing server var input should reset to default

P0lip:


packages/elements-core/src/components/TryIt/Servers/ServerVariables.tsx
```diff
@@ -20,7 +20,10 @@ export const ServerVariables: React.FC<ServerVariablesProps> = ({ variables, val
             key={variable.name}
             variable={variable}
             value={values[variable.name]}
-            onChange={(value: string | number) => onChangeValue(variable.name, String(value))}
+            onChange={(value: string | number) => {
+              const actualValue = String(value);
+              onChangeValue(variable.enum || actualValue !== '' ? 'set' : 'unset', variable.name, actualValue);
```
P0lip, line 25: this is to support an empty value in enum in case one had such

billiegoose (approved):

Issue #2394, opened 2023-08-30
Show AND authentication on API overview and Endpoint level


mnaumanali94:
### **Description of Work**
Add rendering for "AND" authentication within API overview and Operation.

### **Requirements**

- [ ] Render multiple auth on the API overview level. It shows up as Security Name 1 **AND** Security Name 2 e.t.c. The AND is in bold. The description then shows the name of each security in bold followed by its corresponding description
![image](https://github.com/stoplightio/elements/assets/17899454/bced5fd3-3614-4cb3-8624-9ebf0d730630)
- [ ] Render multiple auth on the endpoint level. It shows up as Security Name 1 **AND** Security Name 2 e.t.c. The AND is in bold. The description then shows the name of each security in bold followed by its corresponding description
![image](https://github.com/stoplightio/elements/assets/17899454/d6f736a4-2751-4624-bd29-d56697877cac)

mnaumanali94:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/-Tequila-Mockingbirds-61f945bb2ae66f0010ca5eca/issues/stoplightio/elements/2394?planning-poker) with Zenhub @billiegoose @cjlanzo @dotslashderek @matty-v @P0lip

PR #2408, opened 2023-09-20
feat(docs): support AND security schemes

dotslashderek:
Part of https://github.com/stoplightio/elements/issues/2394

Just the AND support in Docs for HttpService and HttpOperation, plus test updates.

Also added some data-test attributes as @cjlanzo suggested to make e2e testing easier.

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpOperation/Request.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/Request.tsx
@@ -2,11 +2,15 @@ import { Box, NodeAnnotation, VStack } from '@stoplight/mosaic';
 import { HttpSecurityScheme, IHttpOperation } from '@stoplight/types';
 import { useAtom } from 'jotai';
 import { atomWithStorage } from 'jotai/utils';
-import { flatten } from 'lodash';
 import * as React from 'react';
 
 import { useOptionsCtx } from '../../../context/Options';
-import { getReadableSecurityName, shouldIncludeKey } from '../../../utils/oas/security';
+import {
+  getReadableSecurityName,
+  getReadableSecurityNames,
+  shouldAddKey,
+  shouldIncludeKey,
+} from '../../../utils/oas/security';
 import { getDefaultDescription } from '../../../utils/securitySchemes';
 import { MarkdownViewer } from '../../MarkdownViewer';
 import { SectionSubtitle, SectionTitle, SubSectionPanel } from '../Sections';
@@ -35,7 +39,7 @@ export const Request: React.FunctionComponent<IRequestProps> = ({
   if (!request || typeof request !== 'object') return null;
 
   const bodyIsEmpty = isBodyEmpty(body);
-  const securitySchemes = flatten(security);
+  const securitySchemes = security ?? [];
   const hasRequestData = Boolean(
     securitySchemes.length ||
       pathParams.length ||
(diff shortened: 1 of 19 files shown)


packages/elements-core/src/components/Docs/HttpOperation/Request.tsx
```diff
@@ -88,36 +92,49 @@ Request.displayName = 'HttpOperation.Request';
...
+  /* Note: not sure what to do about the nodeId below and how important it was to be tied to the scheme id like that... */
   return (
     <VStack spacing={3}>
       {schemes.map((scheme, i) => (
-        <Box pos="relative" key={i}>
-          <SecurityPanel scheme={scheme} includeKey={shouldIncludeKey(schemes, scheme.type)} />
-          <NodeAnnotation change={nodeHasChanged?.({ nodeId: scheme.id })} />
+        <Box pos="relative" key={i} p={0} data-test="security-row">
+          <SecurityPanel schemes={scheme} includeKey={shouldAddKey(scheme, schemes)} />
+          <NodeAnnotation change={nodeHasChanged?.({ nodeId: scheme.map(node => node.id).join('') })} />
```
P0lip, line 137: That will break Proposals. We'd need to pass an id of http_security_group node here.

dotslashderek: Moved it over to match individual schemes again instead of the security group - I *think* this will retain current functionality - take a look :)

packages/elements-core/src/components/Docs/HttpOperation/Request.tsx
```diff
@@ -88,36 +92,49 @@ Request.displayName = 'HttpOperation.Request';
...
 
-const SecuritySchemes = ({ schemes }: { schemes: HttpSecurityScheme[] }) => {
-  const { nodeHasChanged } = useOptionsCtx();
-
+const SecuritySchemes = ({ schemes }: { schemes: HttpSecurityScheme[][] }) => {
   if (!schemes.length) {
     return null;
   }
-
+  /* Note: not sure what to do about the nodeId below and how important it was to be tied to the scheme id like that... */
```
P0lip, line 132:
```suggestion
```
not relevant anymore :D

P0lip (commented):
The code and the implementation for HttpOperation looks good to me :+1:

What about HttpService? It's currently unchanged in comparison to main

packages/elements-core/src/components/Docs/HttpOperation/Request.tsx
dotslashderek: 🤣

dotslashderek:
@P0lip  - let me get that comment cleaned up :) 

And also 🤔 ... I'm going to add the optional in to this PR for HttpOperation.  That'll complete the work for endpoints.

Then I'll ask you to do another review and plan to close this PR at that point - and take all the HttpService work in a separate PR.

Reason being that this one + Optional can be done and merged directly into Elements and released now as is since they don't require any changes to http-spec and won't need to wait on the official release of stoplight/types.

Right now, since HttpService is showing the securitySchemes array instead of security, not having the AND support doesn't really matter that much?  Or, at least, it won't be apparent that Operation supports it and Service doesn't since Service as currently configured doesn't display the security property at all.

P0lip:
> And also 🤔 ... I'm going to add the optional in to this PR for HttpOperation. That'll complete the work for endpoints.

That works, but a separate PR is fine too. :+1:
Like this already adds value, so we can ship that and add optional a bit later.

> Right now, since HttpService is showing the securitySchemes array instead of security, not having the AND support doesn't really matter that much? Or, at least, it won't be apparent that Operation supports it and Service doesn't since Service as currently configured doesn't display the security property at all.

Yeah, but we intended to switch that view to `security` as the existing behavior is just not what users expected, but I'd leave the final call to @mnaumanali94 on that as it's a solely product decision :D

P0lip (approved):

PR #2413, opened 2023-09-25
feat(elements-core): http-service AND and optional auth support

dotslashderek:
Closes https://github.com/stoplightio/elements/issues/2394


dotslashderek:
@P0lip had to make a few changes to test fixtures related to types / http-spec bump that aren't directly related to this work - tag changes, extensions, a couple more - take a look and make sure everything looks sane please 😄

packages/elements-core/src/components/Docs/HttpService/SecuritySchemes.tsx
```diff
@@ -27,44 +27,47 @@ export const SecuritySchemes: React.FC<SecuritySchemesProps> = ({
...
-          <SecurityScheme
-            key={i}
-            scheme={scheme}
-            defaultIsOpen={defaultScheme ? scheme.key === defaultScheme : i === 0}
-            isCollapsible={schemes.length > 1}
-            showSchemeKey={shouldIncludeKey(schemes, scheme.type)}
-          />
-        ))}
+        {includeOptional && <OptionalMessage />}
+        {secSchemes
```
P0lip, line 31: what about `NodeAnnotation` for the top-level security panel? I suppose right now you won't see which security changed after you expand the panel.
We'd probably need to put it next to the Panel

packages/elements-core/src/components/Docs/Security/PanelContent.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  const collection = schemes.length > 1;
+
+  return (
+    <Box>
+      {schemes.map(scheme => (
+        <Box key={scheme.key} p={2} m={2} border>
+          {collection && (
+            <MarkdownViewer
```
P0lip, line 24: I forgot to mention it in in the other PR but using markdown viewer to render a plain header is an overkill.
IIRC we have `Heading` or something like that, or we could just use `Text`. There's isn't much happening here anyway

P0lip: Don't feel obliged to make that change - I don't think it's a new code so it's alright to keep it as is.

packages/elements-core/src/__fixtures__/services/with-url-variables.ts
```diff
@@ -47,37 +47,39 @@ export const httpServiceWithUrlVariables: IHttpService = {
...
-          refreshUrl: 'https://clientCredentials.refresh-url.com',
-          scopes: { 'scope:clientCredentials': 'clientCredentials scope description' },
-        },
-        authorizationCode: {
-          authorizationUrl: 'https://authorizationCode.authorization-url.com',
-          tokenUrl: 'https://authorizationCode.token-url.com',
-          refreshUrl: 'https://authorizationCode.refresh-url.com',
-          scopes: { 'scope:authorizationCode': 'authorizationCode scope description' },
+  securitySchemes: [],
+  security: [
```
P0lip, line 51: I'd keep it in `securitySchemes` as it was.
What we have right now is not quite valid - `flows` would never get set if `securitySchemes` didn't have that `oauth_scheme` defined.

P0lip (commented):
Looks good overall :+1:

packages/elements-core/src/components/Docs/HttpService/SecuritySchemes.tsx
dotslashderek: Hmmm... is this the http-security-group node then?  I still don't know how I'm supposed to recover that node id and make it match to Billie's work in sl-doc - it looked like she was using the parent node id 🤔 would that be the ID for the overall service - will I have access to that in the specs we're representing?

P0lip: > is this the http-security-group node then? 

right, that'd be it.
I think you'll need to reconstruct it, kind of like we do it here https://github.com/stoplightio/json-schema-viewer/blob/431fc71c26c8621141d6ed6333350cc70eafcfac/src/hash.ts#L18
the `parentId` in this case would be the id of the entire document, so we should have it in theory.
In the case of operation, this would be the id of the operation which should also be there.

dotslashderek: Ok - take a look.  There is something Billie was doing in stoplight-doc - I'm not sure why?  It's the whole if withURI use props.uri except... that prop will never be there?  For a group I mean, since it's an array.  🤔 

Anyhow, just did what I expect her stoplight-doc to do, which is fall back on the parentId and position in the array each time.  Take a look, let me know what you think :) - thanks!

packages/elements-core/src/components/Docs/HttpService/SecuritySchemes.tsx
```diff
@@ -1,44 +1,53 @@
...
+  parentId,
 }) => {
   const includeOptional = secSchemes.length > 1 && secSchemes.some(schemes => schemes.length === 0);
+  const { nodeHasChanged } = useOptionsCtx();
 
   return (
     <Panel rounded isCollapsible={defaultCollapsed} data-test="security-row">
       <Panel.Titlebar bg="canvas-300">
-        <Box as="span" role="heading">
+        <Box as="span" role="heading" key={'1'}>
```
P0lip: what is this key for?

P0lip (approved):

packages/elements-core/src/components/Docs/HttpService/SecuritySchemes.tsx
P0lip: na, the URI won't be there, can ignore that bit

packages/elements-core/src/components/Docs/HttpService/SecuritySchemes.tsx
dotslashderek: I had already whacked it lol :)

PR #2421, opened 2023-10-03
Fix/tryit/initialize with suggested sec values

dotslashderek:
As requested, we'll use the suggested values in the authorization headers / query params inside try it request samples if those values are empty or undefined strings.

(Change from ?? to || is deliberate to replace empty strings as well with suggested defaults).

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -10,7 +10,7 @@
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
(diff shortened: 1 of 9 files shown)


P0lip (approved):

ellyxc:
any particular reason we are sending input placeholder value as Auth header instead of not sending it when user not putting any value?

PR #2425, opened 2023-10-05
fix: mobile export button

mallachari:
Hides the export button for a compact layout in http service and modal components.

https://github.com/stoplightio/elements/assets/14196079/aadbcc68-3a2f-41be-bec3-1da42e50e80a


dotslashderek (approved):
Looks good 🚀

Issue #2399, opened 2023-09-05
Server variables in incorrect order


siimsoni:
## Context
The template for endpoint URL, shown below the endpoint name, is garbled.

## Current Behavior
The order of server variables is reversed, and the opening brace for the first variable has become another closing brace for the first variable.

## Expected Behavior
The endpoint URL template should be rendered correctly.

## Possible Workaround/Solution
For some reason, part of the server variables is set to RTL. The problem disappears for me as a LTR language user when setting it to LTR.

```
<div class="sl-flex sl-overflow-x-hidden sl-text-lg sl-select-all">
  <div dir="rtl" class="sl-overflow-x-hidden sl-truncate sl-text-muted">
    /{a}/{b<span dir="ltr" style="unicode-bidi: bidi-override;">}</span>
  </div>
  <div class="sl-flex-1 sl-font-semibold">/{clientid}</div
</div>
```

## Steps to Reproduce
```
{
    "openapi": "3.1.0",
    "info": {
        "title": "Schema",
        "version": "",
        "description": "Stoplight Elements test\n",
        "license": {
            "name": "Test",
            "url": "http:\/\/test"
        }
    },
    "servers": [
        {
            "url": "\/{a}\/{b}",
            "variables": {
                "a": {
                    "default": "default_a"
                },
                "b": {
                    "default": "default_b"
                }
            }
        }
    ],
    "paths": {
        "\/{param}": {
            "get": {
                "summary": "feed\/1",
                "description": "Best feed",
                "responses": {
                    "200": {
                        "description": "OK",
                        "content": {
                            "application\/json": {
                                "schema": {
                                    "type": "object"
                                }
                            }
                        }
                    }
                },
                "parameters": [
                    {
                        "name": "param",
                        "in": "path",
                        "required": true,
                        "schema": {
                            "type": "number",
                            "minimum": 0
                        }
                    }
                ]
            }
        }
    }
}
```

## Environment
```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8"/>
    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
    <title>Fishnet Feeds Specification</title>
    <!-- needed for adaptive design -->
    <script
        src="https://unpkg.com/@stoplight/elements@7.10.0/web-components.min.js"
[...]

PR #2432, opened 2023-10-18
fix(elements-core): display end of url server vars correctly

dotslashderek:
Fixes https://github.com/stoplightio/elements/issues/2399


packages/elements-core/src/components/Docs/HttpOperation/HttpOperation.tsx
```diff
@@ -126,20 +125,14 @@ function MethodPath({ method, path }: MethodPathProps) {
...
-
-        {trailing !== null && (
-          <Box as="span" dir="ltr" style={{ unicodeBidi: 'bidi-override' }}>
-            {trailing}
-          </Box>
-        )}
+        <Box as="span" dir="ltr" style={{ unicodeBidi: 'bidi-override' }}>
+          {chosenServerUrl}
+        </Box>
       </Box>
```
dotslashderek, line 135: @P0lip I'm not quite sure we strictly speaking needed to do the potential split / re-stich here, if I'm understanding the initial issue.  I do see that if you don't pull out the trailing non-alpha stuff it gets reshuffled to the front if you use rtl ordering.

But I *think* just wrapping the entire url inside a ltr inside the outer rtl Box accomplishes what you were initially fixing here:  https://github.com/stoplightio/elements/pull/2217 - ie, I still get the desired behavior of ellipsis at the front of the url, not the back - and it still maintains urls like 'https:/something.com--' - and also is still selectable.

Anyhow, take a look, let me know what I'm missing and broke 🤣

P0lip (approved):
Works fine!

PR #2434, opened 2023-10-26
feat: form body oneOf selector

EdVinyard:

### DIFF
--- a/packages/elements-core/src/__fixtures__/operations/urlencoded-post-oneof.ts
+++ b/packages/elements-core/src/__fixtures__/operations/urlencoded-post-oneof.ts
@@ -1,0 +1,80 @@
+import { IHttpOperation } from '@stoplight/types';
+
+export const httpOperation: IHttpOperation = {
+  id: '?http-operation-id?',
+  iid: 'POST_todos',
+  method: 'post',
+  path: '/todos',
+  summary: 'Create Todo',
+  responses: [
+    {
+      id: '?http-response-200?',
+      code: '200',
+    },
+  ],
+  servers: [
+    {
+      id: '?http-server-todos.stoplight.io?',
+      url: 'https://todos.stoplight.io',
+    },
+  ],
+  request: {
+    body: {
+      id: '?http-request-body?',
+      contents: [
+        {
+          id: '493afac014fa8',
+          mediaType: 'application/x-www-form-urlencoded',
+          examples: [
+            {
+              id: '83adf9aba208e',
+              key: 'Request New Token',
+              value: {
+                username: 'Neovest provided username',
+                password: 'corresponding password',
+              },
+            },
+            {
+              id: '9dcf9cf7d39ea',
+              key: 'Refresh Token',
+              value: {
+                refresh_token: 'some string',
+              },
+            },
+          ],
+          encodings: [],
+          schema: {
+            oneOf: [
+              {
+                title: 'Authenticate',
+                description: 'supply username and password for refresh and access tokens',
+                type: 'object',
+                properties: {
+                  username: { type: 'string' },
+                  password: { type: 'string' },
+                },
+              },
+              {
+                title: 'Access Token Refresh',
+                description: 'supply a refresh token for new refresh and access tokens',
+                type: 'object',
+                properties: {
+                  refresh_token: { type: 'string' },
+                },
+              },
+            ],
+            'x-stoplight': {
+              id: 'fy4lgsesqt84g',
+            },
+            'x-internal': false,
+            description: '',
+            title: 'Request Body Schemas',
+            $schema: 'http://json-schema.org/draft-07/schema#',
+          },
+        },
+      ],
+    },
+  },
+};
+
+export default httpOperation;
--- a/packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
+++ b/packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
@@ -1,6 +1,7 @@
 import { safeStringify } from '@stoplight/json';
-import { Panel } from '@stoplight/mosaic';
+import { Button, Menu, MenuItems, Panel } from '@stoplight/mosaic';
 import { IMediaTypeContent } from '@stoplight/types';
+import { JSONSchema7, JSONSchema7Definition } from 'json-schema';
 import { omit } from 'lodash';
 import * as React from 'react';
 
@@ -24,9 +25,15 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({
   onChangeParameterAllow,
   isAllowedEmptyValues,
 }) => {
-  const schema = specification.schema;
-  const parameters = schema?.properties;
-  const required = schema?.required;
+  const isOneOf: boolean =
+    specification.schema?.properties === undefined &&
+    specification.schema?.oneOf !== undefined &&
+    specification.schema.oneOf.length > 0;
+  const [schema, setSchema] = React.useState(initialSchema(specification));
+  const parameters: JSONSchema7['properties'] = schema?.properties;
+  const required: string[] = schema?.required ?? [];
+
+  console.log({ schema, parameters, required });
 
   React.useEffect(() => {
     if (parameters === undefined) {
@@ -40,7 +47,11 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({
 
   return (
     <Panel defaultIsOpen>
-      <Panel.Titlebar>Body</Panel.Titlebar>
+      <Panel.Titlebar
+        rightComponent={isOneOf && <OneOfMenu subSchemas={specification?.schema?.oneOf ?? []} onChange={setSchema} />}
+      >
+        Body
+      </Panel.Titlebar>
       <Panel.Content className="sl-overflow-y-auto ParameterGrid OperationParametersContent">
         {mapSchemaPropertiesToParameters(parameters, required).map(parameter => {
           const supportsFileUpload = parameterSupportsFileUpload(parameter);
@@ -82,3 +93,65 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({
     </Panel>
   );
 };
+
+function initialSchema(content: IMediaTypeContent<false>): JSONSchema7 {
+  let wholeSchema = content.schema;
+  let parameters: JSONSchema7['properties'] = wholeSchema?.properties;
+  const isOneOf: boolean = parameters === undefined && wholeSchema?.oneOf !== undefined && wholeSchema.oneOf.length > 0;
+  if (isOneOf) {
+    // TODO: Need to eliminate this (spurious) type-check warning.
+    const firstOneOfItem: JSONSchema7Definition = wholeSchema?.oneOf?.[0];
+    if (typeof firstOneOfItem !== 'boolean') {
+      return firstOneOfItem;
+    }
+  }
+
+  return wholeSchema ?? {};
+}
+
+interface OneOfMenuProps {
+  subSchemas: JSONSchema7Definition[];
+  onChange: (selectedSubSchema: JSONSchema7) => void;
+}
+
+function OneOfMenu({ subSchemas, onChange }: OneOfMenuProps) {
+  const onSubSchemaSelect = React.useCallback(onChange, [onChange]);
+
+  const menuItems = React.useMemo(
+    () =>
+      subSchemas.map((subSchema, index) => {
+        const label = menuLabel(subSchema, index);
+        return {
+          id: `request-subschema-${label}`,
+          title: label,
+          onPress: () => onSubSchemaSelect(typeof subSchema === 'boolean' ? {} : subSchema),
+        };
+      }),
+    [subSchemas, onSubSchemaSelect],
+  );
+
+  return (
+    <Menu
+      aria-label="Examples"
+      items={menuItems}
+      renderTrigger={({ isOpen }) => (
+        <Button appearance="minimal" size="sm" iconRight={['fas', 'sort']} active={isOpen}>
+          Variants
+        </Button>
+      )}
+    />
+  );
+}
+
+function menuLabel(schema: JSONSchema7Definition, index: number): string {
+  if (typeof schema === 'boolean') {
+    return `${index.toString()} boolean`;
+  }
+
+  // TODO: What if `title` and/or `description` are very long strings?
+  return (
+    schema?.title ??
+    schema?.description ??
+    `${index.toString()} - ${Object.getOwnPropertyNames(schema.properties).length} properties`
+  );
+}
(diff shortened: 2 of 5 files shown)


packages/elements-core/src/__fixtures__/operations/urlencoded-post-oneof.ts
```diff
@@ -0,0 +1,80 @@
...
+              id: '9dcf9cf7d39ea',
+              key: 'Refresh Token',
+              value: {
+                refresh_token: 'some string',
+              },
+            },
+          ],
+          encodings: [],
+          schema: {
+            oneOf: [
```
EdVinyard, line 47: This line is the important difference between this example and `packages/elements-core/src/__fixtures__/operations/urlencoded-post.ts`, the use of `oneOf` at the top level of the schema.

yarn.lock
```diff
@@ -7212,9 +7212,9 @@ caniuse-api@^3.0.0:
     lodash.uniq "^4.5.0"
 
 caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001317:
-  version "1.0.30001327"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001327.tgz#c1546d7d7bb66506f0ccdad6a7d07fc6d668c858"
-  integrity sha512-1/Cg4jlD9qjZzhbzkzEaAC2JHsP0WrOc8Rd/3a3LuajGzGWR/hD7TVyvq99VqmTy99eVh8Zkmdq213OgvgXx7w==
+  version "1.0.30001554"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001554.tgz"
+  integrity sha512-A2E3U//MBwbJVzebddm1YfNp7Nud5Ip+IPn4BozBmn4KqVX7AvluoIDFWjsv5OkGnKUXQVmMSoMKLa3ScCblcQ==
```
EdVinyard, line 7217: I was prompted to update this by something on the command line.  Should I keep it?

daniel-white: i think caniuse is ok, but the change to npm is a puzzle

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
```diff
@@ -116,3 +119,48 @@ export function mapSchemaPropertiesToParameters(
     ...(required?.includes(name) && { required: true }),
   }));
 }
+
+export function toParameterSpec(jsonTreeNode: RegularNode): ParameterSpec {
+  const isBoolean = jsonTreeNode.primaryType === 'boolean';
+  // TODO: Why does ParameterSpec assume JSONSchema7 and ignore 6 and 4?
+  const schema = !isBoolean ? (jsonTreeNode.fragment as JSONSchema7) : undefined;
```
EdVinyard, line 126: I don't like this cast, but this didn't feel like the right time to rework the JSON Schema 7 assumptions in this library.

yarn.lock
```diff
@@ -7217,9 +7217,9 @@ caniuse-api@^3.0.0:
     lodash.uniq "^4.5.0"
 
 caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001317:
-  version "1.0.30001327"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001327.tgz#c1546d7d7bb66506f0ccdad6a7d07fc6d668c858"
-  integrity sha512-1/Cg4jlD9qjZzhbzkzEaAC2JHsP0WrOc8Rd/3a3LuajGzGWR/hD7TVyvq99VqmTy99eVh8Zkmdq213OgvgXx7w==
+  version "1.0.30001561"
+  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001561.tgz#752f21f56f96f1b1a52e97aae98c57c562d5d9da"
+  integrity sha512-NTt0DNoKe958Q0BE0j0c1V9jbUzhBxHIEJy7asmGrpE0yG63KTV7PLHPnK2E1O9RsQrQ081I3NLuXGS6zht3cw==
```
EdVinyard, line 7222: I'm not sure why, @daniel-white , but I think the accidental switch to `npmjs.org` here may have come from my use of Node 18 to build this locally before.  After switching back to 16, I think that fixed it.  (Or something did!)

packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
```diff
@@ -24,9 +25,12 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({
   onChangeParameterAllow,
   isAllowedEmptyValues,
 }) => {
-  const schema = specification.schema;
-  const parameters = schema?.properties;
-  const required = schema?.required;
+  // TODO: What about allOf, anyOf?
```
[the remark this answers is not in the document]
EdVinyard: Because this is now handled by `json-schema-viewer`, I don't _think_ there's any further work to do.

yarn.lock
EdVinyard: I'm not sure why, @daniel-white , but I think the accidental switch to `npmjs.org` here may have come from my use of Node 18 to build this locally before.  After switching back to 16, I think that fixed it.  (Or something did!)

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
EdVinyard, line 126: I don't like this cast, but this doesn't seem like the right time to try and remove the JSON Schema 7 assumption.

packages/elements-core/src/components/TryIt/TryIt.tsx
```diff
@@ -259,7 +259,7 @@ export const TryIt: React.FC<TryItProps> = ({
 
       {formDataState.isFormDataBody ? (
         <FormDataBody
-          specification={formDataState.bodySpecification}
+          mediaTypeContent={formDataState.bodySpecification}
```
daniel-white, line 262: is `FormDataBody` a publicly exported type? if so this is a breaking change...

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
```diff
@@ -116,3 +119,48 @@ export function mapSchemaPropertiesToParameters(
...
+  const examples =
+    !isBoolean && jsonTreeNode.fragment.examples && jsonTreeNode.fragment.examples[0]
+      ? [{ key: 'example', value: jsonTreeNode.fragment.examples[0] }]
+      : undefined;
+
+  return {
+    name: last(jsonTreeNode.path) ?? '<<UNKNOWN>>',
```
daniel-white, line 133: do we do this magic unknown elsewhere?

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
```diff
@@ -36,8 +37,10 @@ export function exampleOptions(parameter: ParameterSpec) {
     : null;
 }
 
-export function parameterSupportsFileUpload(parameter: Pick<ParameterSpec, 'schema'>) {
+export function parameterSupportsFileUpload(parameter: Pick<ParameterSpec, 'schema'> | undefined) {
   return (
+    parameter &&
```
daniel-white, line 42: can we  just use the `?.` for each of these after parameter?

packages/elements-core/src/components/TryIt/Parameters/__tests__/parameter-utils.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    const children = root.children ?? [];
+
+    // Act
+    const unknownChildren = children.filter(c => isRequired(c) === undefined);
+    const requiredChildren = children
+      .filter(c => isRequired(c) === true)
+      .map(c => last(c.path))
+      .sort();
+    const optionalChildren = children
+      .filter(c => isRequired(c) === false)
```
daniel-white, line 51:
```suggestion
      .filter(c => !isRequired(c))
```

daniel-white (approved):

packages/elements-core/src/components/TryIt/Body/__tests__/FormDataBody.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+                refresh_token: {
+                  type: 'string',
+                },
+              },
+            },
+          ],
+          // 'x-stoplight': {
+          //   id: 'fy4lgsesqt84g',
+          // },
+          // 'x-internal': false,
```
matthewmurphy, line 45: Can this be removed?
```suggestion
```

packages/elements-core/src/components/TryIt/TryIt.tsx
mallachari: We export only `TryIt` and `TryItWithRequestSamples` so that should be safe

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
EdVinyard: I don't know if we do this elsewhere, but I consider this to mostly be paranoia... and appeasing the type system, since `last()` returns `T | undefined`.  If the actual JSON path array was empty, which shouldn't occur in practice, this wouldn't really be broken but it would be strangely labeled in the UI.  FWIW, there's a similar short-circuit case in the `isRequired()` function we wrote together.

I'll make this easier to see by pulling the value out to a named variable with a comment to explain the null coalesce, and see what you think.

(As an aside, is it possible to meaningfully express "a non-empty array" in TypeScript?  I couldn't figure out how.)

packages/elements-core/src/components/TryIt/TryIt.tsx
EdVinyard: un-renamed in 934903fa

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
EdVinyard: Take a look at 934903fa and let me know what you think.

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
EdVinyard: fixed in 934903fa

packages/elements-core/src/components/TryIt/Body/__tests__/FormDataBody.test.tsx
EdVinyard: Thanks for catching that.  I'll get rid of 'em.

EdVinyard: I removed both in ac00aab1.

packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
```diff
@@ -24,61 +25,120 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({
...
-      console.warn(`Invalid schema in form data spec: ${safeStringify(schema)}`);
-    }
-  }, [parameters, schema]);
+  const { selectedChoice, choices, setSelectedChoice } = useChoices(schema);
 
-  if (parameters === undefined) {
-    return null;
-  }
+  const formFieldRows = visibleChildren(selectedChoice.type);
+  console.log({ schemaNode: schema, formFieldRows, choices });
```
mallachari, line 38:
```suggestion
```

packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
```diff
@@ -24,61 +25,120 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({
...
+
+  /** called when the selection changes */
+  onChange: (choice: Choice) => void;
+}
+
+/**
+ * When the top level schema is `oneOf`, a drop-down menu that allows the user
+ * to select among the sub-schemas; otherwise `null`.
+ */
+export function OneOfMenu({ choices: subSchemas, choice: Choice, onChange }: OneOfMenuProps) {
```
mallachari, line 113: Looks like `Choice` is not used in this component. We could potentially use it for checking selected value in `menuItems`, which might be useful. There's `isChecked` property that we could add.

packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
```diff
@@ -24,61 +25,120 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({
...
+  const onSubSchemaSelect = React.useCallback(onChange, [onChange]);
+
+  const menuItems = React.useMemo(
+    () =>
+      subSchemas.map((subSchema, index) => {
```
mallachari, line 118: Similarly `index` is not used

packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
```diff
@@ -24,61 +25,120 @@ export const FormDataBody: React.FC<FormDataBodyProps> = ({
...
+    return null;
+  }
+
+  return (
+    <Menu
+      aria-label="Variants"
+      items={menuItems}
+      renderTrigger={({ isOpen }) => (
+        <Button appearance="minimal" size="sm" iconRight={['fas', 'sort']} active={isOpen} data-testid="oneof-menu">
+          Variants
```
mallachari, line 139: This is something to consider. We could display the selected name here (if it exists). We do that for Auth schemes

matthewmurphy (approved):

mallachari (approved):
It all looks and works great.
Good job with handling this trap ticket!

packages/elements-core/src/components/TryIt/Parameters/parameter-utils.ts
daniel-white: its fine. definitely could use `throw` expressions

EdVinyard: I considered throwing, but we can still produce an otherwise reasonable `ParameterSpec` instance, we just don't know its name/path.  Without some rules around what is/isn't a valid `ParameterSpec` (oh, how I miss constructors full of preconditions) I'm probably veering away from Postel's law, but trying to produce _something_.

packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
EdVinyard:
```suggestion
```

EdVinyard:
```suggestion
```

packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
EdVinyard: I went ahead and un-renamed that.  Thank you for investigating for me, though.

packages/elements-core/src/components/TryIt/Body/FormDataBody.tsx
EdVinyard: Thanks, that's a good idea.  I made that change in 0954f90b.

PR #2438, opened 2023-11-07
fix: 2412 process env not defined

chohmann:
Bumps mosaic to latest version to fix issue with process.env not being defined.


mallachari (approved):

mallachari:
Approved but looks like there's some node version problem

PR #2440, opened 2023-11-09
feat: export oas utils

mallachari:
Exports `transformOasToServiceNode` and `useExportDocumentProps` functions to let users transform oas documents and use internal elements components independently (SwaggerHub Portal use case).


kaylachun (approved):
Looks good! Are you planning on releasing in this pr?

mallachari:
Yes, I do. Just added it. 
Thanks!

PR #2441, opened 2023-11-13
chore(deps): upgrade to lerna 7

daniel-white:
upgrades lerna to fix dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "jest-environment-jsdom": "^26.3.1",
     "jest-environment-jsdom-global": "^4.0.0",
     "jest-junit": "12.2.0",
-    "lerna": "^3.22.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -121,5 +121,6 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
-  }
+  },
+  "version": "0.0.0"
```
mallachari, line 125: Does new lerna add this version property in main package? Is it needed?

daniel-white: lerna seemed to add it when i smoke tested using `yarn run version`

mallachari: Strange, it didn't do that before. but that shouldn't break the release anyway so it's fine

mallachari (approved):

PR #2442, opened 2023-11-13
Feature/schema dropdown

KyleJohnst:
Added boolean to control render of sidenav, also exposed api that returns list of endpoints so users can get a list of endpoints to use in own custom nav.


KyleJohnst:
Bad commit, will repush

PR #2450, opened 2023-11-16
feat: provide max ref depth to handle recursive schemas

daniel-white:
exposes the option to handle the max depth of refs when resolving


chohmann (approved):
🚀 looks good to me!

PR #2452, opened 2023-11-17
chore: export stacked layout

mallachari:
Context: https://smartbear.slack.com/archives/C05T7G43M0B/p1700213734193029

Adds export of `APIWithStackedLayout` component to `elements` package.
Adds `showPoweredByLink` property to `APIWithStackedLayout`

### DIFF
--- a/packages/elements/src/components/API/APIWithStackedLayout.tsx
+++ b/packages/elements/src/components/API/APIWithStackedLayout.tsx
@@ -24,6 +24,7 @@ type StackedLayoutProps = {
   exportProps?: ExportButtonProps;
   tryItCredentialsPolicy?: TryItCredentialsPolicy;
   tryItCorsProxy?: string;
+  showPoweredByLink?: boolean;
 };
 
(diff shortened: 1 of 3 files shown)


packages/elements/src/components/API/APIWithStackedLayout.tsx
```diff
@@ -61,7 +63,7 @@ export const APIWithStackedLayout: React.FC<StackedLayoutProps> = ({
             nodeTitle={serviceNode.name}
             nodeType={NodeType.HttpService}
             location={location}
-            layoutOptions={{ showPoweredByLink: true, hideExport }}
+            layoutOptions={{ showPoweredByLink, hideExport }}
```
daniel-white, line 66: is this something we actually want to conditionally control?

mallachari: As discussed since this is internal component limited to stacked layout we might have that conditional control

daniel-white (approved):

PR #2454, opened 2023-11-20
feat(layout): responsive sidebar layout

lanemc:
## Elements responsive layout with sidebar

This pr implements a new `responsive` layout option, leaving the `sidebar` option as-is. I propose we release this new layout and suggest customers change the reference to `responsive`, rather than making the existing `sidebar` layout responsive. This will give us time to test it in the wild and won't break any existing implementations with the `sidebar` layout. If no issues come up, we can replace `sidebar` layout.

- Bumped `mosaic` to use the `Drawer` component
- Added a top nav component to show the project name and a trigger button for the sidebar
- Added sidebar menu drawer to display the project toc + simple css adjustment for drawer width
- Added useResponsiveLayout hook to track and leverage screen size
- Added responsive layout to show the top nav at determined breakpoints
- Added a story for the `responsive` template option

https://github.com/stoplightio/elements/assets/15803865/27ff255f-8130-4392-a1b6-49b430259a40

## Preview

https://deploy-preview-2454--stoplight-elements.netlify.app/iframe.html?args=layout:responsive&id=public-api--stacked-layout&viewMode=story

* Note that I removed the `1rem` of padding from the storybook preview to make the nav sit flush with the top of the screen


packages/elements-core/src/hooks/useResponsiveLayout.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const getElementsBreakpoint = (node: ElementsNode) => {
+    return ElementsBreakpoints[node];
+  };
+
+  React.useLayoutEffect(() => {
+    function handleResize() {
+      setScreenType(getScreenType());
+    }
+    window.addEventListener('resize', handleResize);
```
P0lip, line 38: Even though this hook won't be committed in SSR, I'd still recommend checking for `window`

packages/elements-core/src/styles.css
```diff
@@ -2,3 +2,10 @@
 @import '@stoplight/mosaic/themes/default.css';
 
 @import 'core.css';
+
+/* Drawer */
+
+.drawer-container.in.open .left,
```
P0lip, line 8: That class name is a bit unfortunate as it's quite generic and as such may conflict with user-defined stylesheets. We should prefix it with `sl-` and tweak the usage of the library if they allow custom CSS class names.

packages/elements-dev-portal/src/components/Search/SearchOverlay.tsx
```diff
@@ -0,0 +1,92 @@
...
+  data,
+  toc,
+  nodeSlug,
+  projectSlug,
+  branchSlug,
+  isSearchShowing,
+  onClose,
+  onClick,
+}: SearchOverlayProps) => {
+  const BASE_URL =
```
mallachari, line 39: Looks like this is unused

packages/elements/src/containers/API.tsx
```diff
@@ -190,6 +194,20 @@ export const APIImpl: React.FC<APIProps> = props => {
           tryItCorsProxy={tryItCorsProxy}
         />
       )}
+      {layout === 'responsive' && (
```
mallachari, line 197: We should make one of those layouts default as `layout` property is optional

mallachari: I'm just wondering if we need a separate option for responsive layout. We might treat responsiveness as enhancement of sidebar layout in this case. This way we could also include it in `StoplightProject`

packages/elements-core/src/styles.css
billiegoose: Ah. That will require a fix to `mosaic`: https://github.com/stoplightio/mosaic/blob/58f7db3a8c5a5d599ad3c02fd8a3461657a72998/libs/core/src/components/Drawer/Drawer.tsx#L106-L110

Do you want to do that or shall I?

billiegoose:
Alright, it builds finally. I don't love it but I don't hate it.

examples/react-cra/package.json
```diff
@@ -19,7 +19,6 @@
     "@types/react": "^17.0.11",
     "@types/react-dom": "^16.9.0",
     "@types/react-router-dom": "^5.1.7",
-    "node-sass": "^6.0.1",
```
billiegoose, line 22: We don't need `node-sass` anymore, and it's a pain to install (requires the "right" version of python).

examples/react-gatsby/package.json
```diff
@@ -38,11 +38,9 @@
     "@types/react-helmet": "^6.0.0",
     "gatsby-plugin-sass": "^4.8.0",
     "gh-pages": "^3.2.3",
-    "node-sass": "^6.0.0",
-    "node-sass-package-importer": "^5.3.2",
```
billiegoose, line 42: We don't need `node-sass` anymore, and it's a pain to install (requires the "right" version of python).

paulatulis:
I would love to be able to see storybook locally so I could play around more with the search overlay, but with the responsive story included in the deploy preview, it looks good! 
One thing I noticed off the bat is the inconsistent bg color between top, middle, and bottom of the side nav:
Screenshot 2024-01-18 at 4 38 41 PM

billiegoose:
> One thing I noticed off the bat is the inconsistent bg color between top, middle, and bottom of the side nav

I'll look into that!

packages/elements-core/src/components/Layout/ResponsiveSidebarLayout.tsx
```diff
@@ -0,0 +1,158 @@
...
+      <Flex ml={4} mb={5} alignItems="center">
+        {typeof logo === 'string' ? (
+          <Logo logo={{ url: logo, altText: 'logo' }} />
+        ) : (
+          typeof logo === 'object' && <Logo logo={logo} />
+        )}
+        <Heading size={4}>{name}</Heading>
+      </Flex>
+      <Flex flexGrow flexShrink overflowY="auto" direction="col">
+        <TableOfContents tree={tree} activeId={pathname} Link={Link} onLinkClick={onTocClick} />
```
paulatulis, line 118: Found it
```suggestion
        <TableOfContents tree={tree} activeId={pathname} Link={Link} onLinkClick={onTocClick} isInResponsiveMode />
```

paulatulis (commented):
I think I found the missing prop that will make the sidebar one consistent color. Up to you if you want to implement!

billiegoose:
Aight. To fix the color thing, as well as the overly generic classnames, I'm gonna make a PR to Mosaic.

packages/elements-core/src/components/Layout/ResponsiveSidebarLayout.tsx
billiegoose: Oh! That's lovely! I'll do that.

paulatulis (approved):
Looks great, Billie!

billiegoose (approved):

PR #2456, opened 2023-11-20
feat: stacked layout location property

mallachari:
Motivtion: 
Using `APIWithStackedLayout` component externaly without react router usage that causes compatibility issues

Changes:
- Moves `useLocation` call from `APIWithStackedLayout` to `Api`
- adds `location` property to `APIWithStackedLayout` component
- fixes comparing location hash with item name


kaylachun (approved):
Looks good to me!

brendarearden (approved):

PR #2460, opened 2023-11-29
Feat/2449 replace depfu with dependabot

chohmann:
Replaces depfu with dependabot.


chohmann:
closed https://github.com/stoplightio/elements/pull/2433

PR #2487, opened 2023-12-20
fix: callback operation header

mallachari:
Addresses: https://github.com/stoplightio/platform-internal/issues/18498

Within callbacks we don't want to show base url from parent operation. Since callback path usually comes with full url, either as runtime expression or static url, we want to show only path.

This PR adds an optional property `hideServerUrl` to `OperationHeader` and `MethodPath` and sets it to true for `Callback` component.

Path as runtime expression:
Screenshot 2023-12-20 at 17 13 00

Static path:
Screenshot 2023-12-20 at 17 13 50

### DIFF
--- a/packages/elements-core/src/components/Docs/HttpOperation/Callbacks.tsx
+++ b/packages/elements-core/src/components/Docs/HttpOperation/Callbacks.tsx
@@ -1,8 +1,8 @@
-import { useOptionsCtx } from '@stoplight/elements-core/context/Options';
 import { Box, Flex, NodeAnnotation, Select, VStack } from '@stoplight/mosaic';
 import { IHttpCallbackOperation } from '@stoplight/types';
 import * as React from 'react';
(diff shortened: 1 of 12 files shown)


matthewmurphy (approved):

packages/elements-core/src/components/Docs/HttpOperation/Callbacks.tsx
```diff
@@ -65,9 +65,10 @@ export const Callback = ({ data, isCompact }: CallbackProps) => {
         <OperationHeader
           id={data.id}
           method={data.method}
-          path={`/${data.path}`}
+          path={`${data.path}`}
```
daniel-white, line 68:
```suggestion
          path={data.path}
```

daniel-white (approved):

brendarearden (approved):

PR #2489, opened 2024-01-01
feat(elements): support schema grouping

dddswit:
Unlike Elements Dev Portal, Elements does not group schemas into tags. This PR fixes this issue.

### DIFF
--- a/packages/elements/src/components/API/APIWithStackedLayout.tsx
+++ b/packages/elements/src/components/API/APIWithStackedLayout.tsx
@@ -11,7 +11,7 @@ import { NodeType } from '@stoplight/types';
 import cn from 'classnames';
 import * as React from 'react';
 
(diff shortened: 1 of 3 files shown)


mallachari:
@dddswit thank you for bringing up missing schemas and creating this PR. 
The design for elements is to keep endpoints, schemas and now webhooks under separate categories. That also means grouping them under these categories independently unlike we do in stoplight platform and elements-dev-portal (since we operate on single OAS document instead of a project). 
Having that in mind (and after recently introducing webhooks as separate category with independent grouping) we decided to follow that pattern with schemas. We've just merged this [PR](https://github.com/stoplightio/elements/pull/2520) (released in version 8.0.4) that groups models under Schemas category. Feel free to try it out and suggest any changes that might be worth to have.

PR #2497, opened 2024-01-04
fix(dev-portal): fix broken hash router links

daniel-white:
when using elements-dev-portal and the hash router, a custom toc with slugs would result in a broken page. this fixes the route matching and makes the routes absolute

### DIFF
--- a/packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx
+++ b/packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx
@@ -1,6 +1,7 @@
 import { Box, Flex, Icon } from '@stoplight/mosaic';
 import * as React from 'react';
 
+import { useRouterType } from '../../context/RouterType';
 import { useFirstRender } from '../../hooks/useFirstRender';
 import { VersionBadge } from '../Docs/HttpOperation/Badges';
 import {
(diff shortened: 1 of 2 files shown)


packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx
```diff
@@ -331,53 +343,56 @@ const Node = React.memo<{
...
-        isInResponsiveMode={isInResponsiveMode}
-        onClick={handleClick}
-      />
-    </Box>
-  );
-});
+    return (
+      <Box
+        as={LinkComponent}
+        to={makeSlugAbsoluteRoute && !/^\//.test(item.slug) ? `/${item.slug}` : item.slug}
```
EdVinyard, line 373: It took me a minute to find this change, so I'm just marking it for other reviewers.

Is there an advantage to using a regular expression here instead of `!item.slug.startsWith('/')`?

EdVinyard (approved):
Looks good to me.  Had one tiny question/suggestion.  Feel free to take it or leave it.

packages/elements-core/src/components/MosaicTableOfContents/TableOfContents.tsx
daniel-white: old habits die hard

Issue #2492, opened 2024-01-03
Update to http-spec 7


mallachari:
`Http-spec` v7 brings [breaking changes](https://github.com/stoplightio/http-spec/pull/258/files) in oas operation transformation. It introduces webhook operations and changes signature of `transformOas2Operation` and `transformOas3Operation` methods by replacing `path` with `name` and adding `config` object.
Update elements package to use new `http-spec` version.

daniel-white:
this has been released in 8.0.0

PR #2500, opened 2024-01-09
feat(oas): support webhooks from oas3.1

daniel-white:
BREAKING CHANGE

adds support for oas3.1 webhooks

addresses #2492


packages/elements-core/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -34,6 +34,7 @@ import {
 
 const ActiveIdContext = React.createContext<string | undefined>(undefined);
 const LinkContext = React.createContext<CustomLinkComponent | undefined>(undefined);
+LinkContext.displayName = 'LinkContext';
```
daniel-white, line 37: trying to make this work easier with the dev tools

packages/elements-core/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -261,8 +265,9 @@ const Group = React.memo<{
         depth={depth}
         isActive={showAsActive}
         icon={
-          NODE_TITLE_ICON[item.title] && (
```
daniel-white, line 264: when displaying the zoom example spec, this was causing an endpoint group to have this icon - so added a type prop (not sure if dev portal's toc needs to include this or not)

packages/elements-core/src/components/TableOfContents/types.ts
```diff
@@ -29,14 +29,17 @@ export type TableOfContentsGroupItem =
 export type TableOfContentsGroup = {
   title: string;
   items: TableOfContentsGroupItem[];
+  itemsType?: 'article' | 'http_operation' | 'http_webhook' | 'model';
```
daniel-white, line 32: 🤔 does the ToC api need to set this as well?

mallachari: In a way yes. There a `type` property in `TableOfContentsNode` that we'll need to extend with `http_webhook` type

mallachari (approved):
Code and UI looks great!
Screenshot 2024-01-10 at 01 37 10

Seems like the breaking change applies only to elements-core package.
It might be also worth to add webhook examples to elements-core storybook HttpOperation and TableOfContents stories. And/or adding new story to elements package storybook with simple document with webhooks.

brendarearden (approved):
LGTM!

EdVinyard (approved):

PR #2516, opened 2024-02-01
fix(deps): upgrade http-spec and lodash

daniel-white:
upgrade lodash and http-spec


mallachari (approved):

matthewmurphy (approved):
Yay for fixing bad tests

PR #2522, opened 2024-02-12
fix: node process

mallachari:
Bumps mosaic that removes rollup plugin and fixes failing build for react cra and gatsby example apps.
Updates react documentation for `API` and `StoplightProject` by adding polyfill section.


docs/getting-started/dev-portal/react.md
```diff
@@ -75,4 +76,74 @@ Now start the development server.
 yarn start
 ```
 
-And you should see the API reference documentation for the Zoom API.
\ No newline at end of file
+And you should see the API reference documentation for the Zoom API.
+
+## Polyfills
```
kaylachun, line 81: I stumbled upon [this](https://github.com/stoplightio/elements/blob/0c32e6f454b11a2aa81c4ff07fc1e452ad71cda0/docs/getting-started/elements/react.md?plain=1#L9) as well; are they related?  If so, should we also update that readme?

mallachari: Thanks, I didn't notice that note. Yes, we may update it and point to that new paragraph

mallachari: Updated that note in elements docs and added to dev-portal.
I'll also create PRs to elements and dev-portal template repos.

kaylachun (approved):

PR #2529, opened 2024-02-29
Fix/stop 230 remove search from stoplight project component

chohmann:
[STOP-230](https://smartbear.atlassian.net/browse/STOP-230?atlOrigin=eyJpIjoiNDljODUwMTZhOGFhNDM0Y2E0NTI4NThkNjE1MmEzMzciLCJwIjoiaiJ9)

# Description
Removes the seemingly partially implement search overlay functionality from StoplightProject component.

The API Container continues to have the responsive TOC functionality introduced at the same time of this partially implemented search functionality in [this PR](https://github.com/stoplightio/elements/pull/2454).

[STOP-230]: https://smartbear.atlassian.net/browse/STOP-230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "8.1.1",
+  "version": "8.1.0",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**",
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "8.1.0",
+  "version": "8.1.1",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**",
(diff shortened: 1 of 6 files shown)


packages/elements-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "8.1.0",
+  "version": "8.1.1",
```
mallachari, line 3: The change happens only in `dev-portal` so we don't need to bump `elements-core`

packages/elements/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements",
-  "version": "8.1.0",
+  "version": "8.1.1",
```
mallachari, line 3: Same here, no need for that bump as nothing in that package has changed

mallachari (approved):

PR #2550, opened 2024-03-18
Fix(elements-dev-portal): hook cache invalidation on jwt refresh

kaylachun:

### DIFF
--- a/packages/elements-dev-portal/src/hooks/useGetBranches.ts
+++ b/packages/elements-dev-portal/src/hooks/useGetBranches.ts
@@ -8,7 +8,7 @@ import { getBranches } from '../handlers/getBranches';
 export function useGetBranches({ projectId }: { projectId: string }) {
   const { platformUrl, platformAuthToken } = React.useContext(PlatformContext);
   return useQuery(
-    [...devPortalCacheKeys.branchesList(projectId), platformUrl, platformAuthToken],
+    [...devPortalCacheKeys.branchesList(projectId), platformUrl],
     () => getBranches({ projectId, platformUrl, platformAuthToken }),
     {
       enabled: projectId ? true : false,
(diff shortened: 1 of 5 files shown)


packages/elements-dev-portal/src/hooks/useGetBranches.ts
```diff
@@ -8,7 +8,7 @@ import { getBranches } from '../handlers/getBranches';
 export function useGetBranches({ projectId }: { projectId: string }) {
   const { platformUrl, platformAuthToken } = React.useContext(PlatformContext);
   return useQuery(
-    [...devPortalCacheKeys.branchesList(projectId), platformUrl, platformAuthToken],
```
daniel-white, line 11: is there a user id or some other context we can use?

daniel-white: or like if the user is actually logged out or not?

kaylachun: @daniel-white We do have `state.auth.isLoggedIn` that we can pass through to the provider.  Would that work?

kaylachun: Ok added the `isLoggedIn` context here https://github.com/stoplightio/elements/pull/2550/commits/9db422a2dcac03c3392f2f775c899d5ada11cecf

brendarearden: Is that going to be an issue for anonymous users who never log in?

kaylachun: @brendarearden I just tested it and it works as an anonymous user.  My understanding is the array being passed in to `useQuery` is used as a key and invalidated if any of those values change

brendarearden (approved):
lgtm!

PR #2573, opened 2024-04-25
fix: remove the postfix License after the license name in AdditionalInfo

### LINKED ISSUE #2523: $.info.license.name Always Add Word "License"
Elements adds the word "License" to the end of the string contained in $.info.license.name. This make the assumption that every license is an Apache or MIT or something that would always be followed by the word "License", but it may be that the API uses another convention. 

## Context
Would like full control over the License string.

## Current Behavior
Appends the word "License"

## Possible Solution(s)
Simply remove the word "License". If someone wants to append the word "License", they can do so themselves.

weyert:

- [X ] Read [`CONTRIBUTING.md`](../CONTRIBUTING.md)

This pull request changes the way the license name gets rendered in `HttpService` and should fix #2523


kentbulza:
I don't know the team philosophy when the data doesn't match the OpenAPI spec. You could not display at all if the strict terms "mutually exclusive" aren't followed ( https://spec.openapis.org/oas/v3.1.0#license-object ) ; or you could error. Probably should follow whatever's your convention.

There's not a test case for "name" alone. Only name is required and if name alone is there, then it seems to me it it should be just the name text. You don't want to get "undefined" as the link.

weyert:
@kentbulza Thanks, I have updated it

brendarearden:
We do like this approach for fixing the postfix, but our team needs to do additional testing prior to merging this in.

mnaumanali94 (approved):
LGTM!

ben-smartbear (approved):

weyert:
Can we merge it? ❤️

mnaumanali94:
@weyert Can we fix the tests please and then we should be good to merge!

weyert:
@mnaumanali94 I have updated the tests

weyert:
The end to end tests are funning for me locally. Not sure what's up with the CI 

```
       Spec                                              Tests  Passing  Failing  Pending  Skipped
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  api.cy.ts                                00:04       10       10        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  stoplight-project.cy.ts                  00:09       12       12        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        00:14       22       22        -        -        -
```

PR #2607, opened 2024-07-01
fix(elements-core): Do not omit request body for certain HTTP methods

provokateurin:

The HTTP spec does not forbid using request bodies for GET, DELETE and other methods (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET for example). OpenAPI doesn't forbid this either, so it is possible to have a valid specification that is not correctly executed by this library.

I'm not sure about adding tests, as I couldn't find any tests that cover this area of the code so far, or at least not directly unit testing this method. Let me know if you want to have tests for this change and how I can implement them best.

### DIFF
--- a/packages/elements-dev-portal/package.json
+++ b/packages/elements-dev-portal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-dev-portal",
-  "version": "2.4.4",
+  "version": "2.4.5",
   "description": "UI components for composing beautiful developer documentation.",
   "keywords": [],
   "sideEffects": [
@@ -66,7 +66,7 @@
   "dependencies": {
     "@stoplight/markdown-viewer": "^5.7.1",
     "@stoplight/mosaic": "^1.53.3",
-    "@stoplight/elements-core": "~8.4.4",
+    "@stoplight/elements-core": "~8.4.5",
     "@stoplight/path": "^1.3.2",
     "@stoplight/types": "^14.0.0",
     "classnames": "^2.2.6",
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "8.4.4",
+  "version": "8.4.5",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**",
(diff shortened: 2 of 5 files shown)


mnaumanali94:
@provokateurin Tests seem to be failing particularly the `lint and check` step. Others probably are unrelated. Can you take a look and fix them please. Then you can ask for a re review.

provokateurin:
Done!

provokateurin:
Hey @mnaumanali94 do you mind taking another look? Would be nice to have this fix available soon :)

provokateurin:
Is the e2e failure related? I can't tell from the logs it's giving

provokateurin:
Ok I can bump the version. Should I wait for you to check the tests or do it right away?

darekplawecki:
@provokateurin no need to wait, you can do it right away :)

packages/elements-dev-portal/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-dev-portal",
-  "version": "2.4.4",
+  "version": "2.4.5",
```
darekplawecki, line 3: You also need to upgrade `elements-core` version in the dependencies here below and in the `packages/elements/package.json`.

darekplawecki:
@provokateurin I checked E2E Tests. They are failing while trying to send a request:

Screenshot 2024-07-29 at 14 58 03

This is the error that is thrown: 
```Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.```

So it seems like the `fetch` method from window API in browser forbids request body for GET requests and it breaks the "Try It" functionality.

provokateurin:
Even though I think it is wrong, Chrome seems to prevent this. I already saw this in a real codebase where a GET request had a body and failed, so I don't think there is a way around this :/
Maybe parts of this PR can still be merged, since e.g. curl probably allows this behavior and there is no reason to prevent it. For the TryIt a warning could be displayed that explains why the body isn't visible/in the request.

darekplawecki:
@provokateurin sounds good to me :)

mnaumanali94:
@provokateurin  If you want to take out the stuff around try it, we can review this again.

provokateurin:
Sorry, I'm dealing with other problems at the moment. I'm not sure I will revisit this soon, so feel free to close this PR or pick up the working parts of it.

PR #2777, opened 2025-03-06
Allow TryIt to send request body for GET and DELETE requests, fixes #2230

### LINKED ISSUE #2230: TryIt doesn't send request body for DELETE requests
## Context

When making an api request with method DELETE which has some body, the body is not sent. This seems to be intentional as per these lines:
https://github.com/stoplightio/elements/blob/b20ba4422985ff35617171ec72103719fe3aa2b6/packages/elements-core/src/components/TryIt/build-request.ts#L66

## Current Behavior

As explained above

## Expected Behavior

Request body should be sent for DELETE requests. It's worth noting that DELETE requests with request body is NOT against the specs. There might be many legacy apis that developers are documenting, which may have a DELEET endpoint with request body. Open Api Spec also supports DELETE requests with body.  https://spec.openapis.org/oas/v3.1.0#version-3-1-0

Screenshot 2022-09-05 at 15 22 20

## Steps to Reproduce

1. Use try-it widget to make a DELETE request with request body. Note how in the screenshot below, it is documented that the endpoint accepts some body, but "Request Sample" does not reflect the same

Screenshot 2022-09-05 at 15 23 51

## Environment

- Version used:
- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your environment/workspace/project:

dbergey:

- [X] Read [`CONTRIBUTING.md`](../CONTRIBUTING.md)


packages/elements-core/src/components/TryIt/build-request.ts
```diff
@@ -147,7 +147,7 @@ export async function buildFetchRequest({
   const serverUrl = getServerUrl({ httpOperation, mockData, chosenServer, corsProxy, serverVariableValues });
 
   const shouldIncludeBody =
-    ['PUT', 'POST', 'PATCH'].includes(httpOperation.method.toUpperCase()) && bodyInput !== undefined;
+    ['GET', 'PUT', 'POST', 'PATCH', 'DELETE'].includes(httpOperation.method.toUpperCase()) && bodyInput !== undefined;
```
mkarnicki, line 150: A `GET` does not have a body. It uses the query string parameters to pass data.

mkarnicki: I stand corrected!

```
   A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.
```

marcelstoer:
I guess this should be merged together with #2785.

ThisIsAreku:
Hi ! Can this be merged ?

PR #2826, opened 2025-08-06
Fix lockfile-maintenance

marcelstoer:
The `lockfile-maintenance` task has been broken for some time e.g. see https://app.circleci.com/pipelines/github/stoplightio/elements/10065/workflows/297cd28c-c660-4d1f-87be-5837984512fc/jobs/46146

Either the command needs to be updated or the script renamed.


darekplawecki (approved):
Thanks for fixing that!

PR #2832, opened 2025-09-24
Fix: Stop-4059 due to invalid/undefined code in response for charAt function -- Added validation check before passing code to charAt function

SB-ChetanKorde:
# Defect  Details

## Jira Ticket - https://smartbear.atlassian.net/browse/STOP-4059

The issue was observed as a JavaScript error in the console:
GenericReactErrorBoundaryError: Cannot read properties of undefined (reading 'charAt').
Upon code analysis, it was found that the Docs UI in the platform relies on the response.tsx file from the elements-core package. The root cause was that the codeToIntentVal function was receiving undefined instead of a valid string, leading to the charAt error. _(screenshot is added below for error reproduced)_

**Solution Explanation:**
A type and value check was added to the codeToIntentVal function to ensure it only operates on valid, non-empty strings. If the input code is invalid or empty, the function now returns 'default', preventing the charAt error and ensuring robust handling of unexpected values.

### Note - 
We attempted to reproduce the reported defect locally using both the platform and elements repositories, but initially did not encounter the issue. Upon further analysis of the console logs, we manually tested the function by passing undefined as the code parameter and confirmed that this led to the observed error. This suggests that undefined may be passed as a parameter on the customer site. Based on these findings, we implemented validation checks to prevent such errors. The solution is now in place and should enhance overall robustness.

- Error before code change
- 
image-20250923-071647 (1)

- After fix & Unit testing 
- 
 
Screenshot 2025-09-24 at 12 56 31 PM


examples/angular/package.json
prafullaAtSB, line 1: this is required change, or else angular build will fail

examples/angular/package.json
```diff
@@ -39,8 +39,5 @@
     "@angular/language-service": "^12.0.5",
     "angular-http-server": "^1.10.0",
     "typescript": "4.2.4"
-  },
-  "resolutions": {
```
prafullaAtSB, line 43: this is required change, or else angular build will fai

packages/elements-core/src/utils/ref-resolving/resolvedObject.ts
```diff
@@ -78,8 +78,6 @@ export const isResolvedObjectProxy = (someObject: object): boolean => {
 };
 
 export const getOriginalObject = (resolvedObject: object): object => {
-  if (!resolvedObject) return resolvedObject;
```
prafullaAtSB, line 81: is this IATA change? make sure

examples/angular/package.json
SB-ChetanKorde: We have created new IATA-removal-main branch and from that we have removed IATA changes. It is not yet approved and merge in main branch that's why these changes are visible in this PR.

packages/elements-core/src/utils/ref-resolving/resolvedObject.ts
SB-ChetanKorde: We have created new IATA-removal-main branch and from that we have removed IATA changes. It is not yet approved and merge in main branch that's why these changes are visible in this PR.

these are the IATA changes

PR #2835, opened 2025-10-14
fix: added proper encoding the query parameters and kept decoded in p…

SB-pradeep-bande:
[STOP-1518](https://smartbear.atlassian.net/browse/STOP-1518)

I have updated the decoding logic to ensure that only the actual placeholder parameters are selectively decoded, rather than decoding the entire URL. This preserves the encoding of curly brackets where necessary, resolving the issue observed in the "Request Sample" section.

[STOP-1518]: https://smartbear.atlassian.net/browse/STOP-1518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/elements-core/src/components/RequestSamples/convertRequestToSample.ts
```diff
@@ -12,17 +12,39 @@ export const convertRequestToSample = async (
 
   try {
     const snippet = new HTTPSnippet(request);
-
+    console.log('Request -------', request);
```
bhaskarsontakke: Remove console

PR #2853, opened 2025-11-26
Handled Active Element of Table of Content

SB-venkatyadavilli:
[STOP-4166](https://smartbear.atlassian.net/browse/STOP-4166)

## Description

Previously, if the same item appeared multiple times in the Table of Contents, clicking on any instance would highlight all occurrences as active. With this fix, only the clicked element is shown as active, while all other instances remain inactive.

Handled Active Element of Table Of Content.

## How Has This Been Tested?

Tested by running elements-dev-portal storybook and Yalc testing.

### Before:
image

### After:
image

https://github.com/user-attachments/assets/cc534767-b55e-494d-9a2a-d6a83d2fbc40

[STOP-4166]: https://smartbear.atlassian.net/browse/STOP-4166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/elements-core/src/components/TableOfContents/TableOfContents.tsx
+++ b/packages/elements-core/src/components/TableOfContents/TableOfContents.tsx
@@ -1,7 +1,7 @@
 import { Box, Flex, Icon, ITextColorProps } from '@stoplight/mosaic';
 import { HttpMethod, NodeType } from '@stoplight/types';
 import * as React from 'react';
-import { createContext, useState } from 'react';
+import { useState } from 'react';
 
 import { useFirstRender } from '../../hooks/useFirstRender';
 import { resolveRelativeLink } from '../../utils/string';
@@ -15,8 +15,8 @@ import {
   NODE_TYPE_TITLE_ICON,
 } from './constants';
 import {
+  ActiveItemContextType,
   CustomLinkComponent,
-  GroupContextType,
   TableOfContentsDivider,
   TableOfContentsGroup,
   TableOfContentsGroupItem,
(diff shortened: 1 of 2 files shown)


packages/elements-core/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -37,6 +38,25 @@ const ActiveIdContext = React.createContext<string | undefined>(undefined);
...
 
+// Create the context with a default undefined value
+export const GroupContext = createContext<GroupContextType>({
+  lastActiveIndex: '',
+  setLastActiveIndex: () => {},
+});
+
+// Provider component
+const GroupProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+  const [lastActiveIndex, setLastActiveIndex] = useState<string>(''); // default value 0
```
SB-ChetanKorde: Taking default value for state as empty string, instead that we can use zero or null. should match comment.

packages/elements-core/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -37,6 +40,25 @@ const ActiveIdContext = React.createContext<string | undefined>(undefined);
...
 LinkContext.displayName = 'LinkContext';
 
+// Create the context with a default undefined value
+export const GroupContext = createContext<GroupContextType>({
+  lastActiveIndex: '',
+  setLastActiveIndex: () => {},
+});
+
+// Provider component
+const GroupProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
```
darekplawecki: Do we need a separate Context/Provider. The name is a bit misleading, it's not clear what it does. Couldn't we extend the existing `ActiveIdContext` with `lastActiveIndex` and `setLastActiveIndex`? Perhaps we should rename it to `ActiveItemContext` - it should store both `activeId` and `activeIndex`.

PR #2961, opened 2026-08-07
Stop-6047-Fixed Example Data of Type Integer(int64,32) and Number(double/float)

SB-venkatyadavilli:
[STOP-6047](https://smartbear.atlassian.net/browse/STOP-6047)

## Description:
Fixed the issue currently affecting Integer (int32, int64) and Number (float, double) types. Previously, when no default value was provided, the example section automatically inferred and displayed the minimum value based on the type's range. With this fix, the example section now displays 0 for these types when no default value is specified.

## How Has This Been Tested?:
have tested on Elements-dev-portal

### Before:
image

### After:
image

### Before:
image

### After:
image

## Impact Analysis:
[https://smartbear.atlassian.net/wiki/x/DAAykwE](https://smartbear.atlassian.net/wiki/x/DAAykwE)

[STOP-6047]: https://smartbear.atlassian.net/browse/STOP-6047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/elements-core/package.json
+++ b/packages/elements-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "9.0.24",
+  "version": "9.0.25",
   "sideEffects": [
     "web-components.min.js",
     "src/web-components/**",
--- a/packages/elements-core/src/components/TryIt/Response/Response.tsx
+++ b/packages/elements-core/src/components/TryIt/Response/Response.tsx
@@ -13,6 +13,7 @@ export interface ResponseState {
   bodyText?: string;
   contentType: string | null;
   blob?: Blob;
+  skipBodyParsing?: boolean;
 }
 
 export interface ErrorState {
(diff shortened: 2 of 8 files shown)


pareshb3089 (approved):
looks goods

SB-ChetanKorde (commented):
I think we need to upgrade package version for elements if we make any changes and want to publish it. I dont see files related to it.

SB-venkatyadavilli:
> I think we need to upgrade package version for elements if we make any changes and want to publish it. I dont see files related to it.

have added Bump up Elements versions files

SB-ChetanKorde (approved):
Package upgrades and Changes for newly added OAS numeric format and range looks good to me.

packages/elements-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/elements-core",
-  "version": "9.0.24",
+  "version": "9.0.25",
```
ben-smartbear: this should be 9.0.26

ben-smartbear (changes requested):
lgtm, but the versions are wrong

packages/elements-core/package.json
SB-venkatyadavilli: have updated the versions

ben-smartbear (approved):

